PR #13, opened 2015-09-19
Move and exception safety for final_act. resolves #11 resolves #12

### LINKED ISSUE #11: Design of Final_act is dependent on copy elision to avoid multiple 'final acts'
The move constructor does not flag the moved-from object as non-invokable

    auto invocation_count = 0;
    auto f = finally([&invocation_count]{ ++invocation_count;});
    assert(invocation_count == 1);

If the copy-by-rvalue-reference is not elided (which is possible but not necessary) then the invocation count will be 2. The `final_act` object needs an internal bool to track its state or the C++ standard needs to *require* copy elision. 

See: https://github.com/jbcoe/CppSandbox/blob/master/ScopeGuard/ScopeGuard.cpp

for a similar class.

### LINKED ISSUE #12: Final_act can lead to program termination if the final act throws an exception
Final_act does not check that the object to be invoked is no-except invokable or try to catch any generated exceptions. This could be bad if multiple final acts are used to clear up state on scope exit.

Either Final_act should catch exceptions or there should be an requirement that the invokable object cannot throw exceptions when invoked. It may be desirable to have two implementations of Final_act:
Final_act and Final_act_noexcept which try and catch exceptions or do not depending on the noexcept status of the invokable object.

jbcoe:
Copy elision no longer required for `finally`.
Move construction does not result in multiple function object invocation.
Object returned by final_act cannot terminate program if it throws.
Tests added.

### DIFF
--- a/include/gsl.h
+++ b/include/gsl.h
@@ -40,25 +40,97 @@ using std::shared_ptr;
 //
 
 // Final_act allows you to ensure something gets run at the end of a scope
-template <class F>
-class Final_act
+template <typename F_t>
+class Final_act_noexcept
 {
-public:
-    explicit Final_act(F f) : f_(f) {}
-    
-    Final_act(const Final_act&& other) : f_(other.f_) {}
-    Final_act(const Final_act&) = delete;
-    Final_act& operator=(const Final_act&) = delete;
-    
-    ~Final_act() { f_(); }
+  F_t f;
+  bool run;
 
-private:
-    F f_;
+public:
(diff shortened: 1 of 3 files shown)


neilmacintosh:
I can't accept this PR I'm afraid. We don't want to allow Final_act to swallow exceptions. Final_act should be noexcept (for reasons I outline in the original issue) and so if something it calls happens to throw, then program termination will result. I'd be happy to look at the PR again if the catch(...) rework was removed.

PR #15, opened 2015-09-21
Added missing owner<T>, corrected array_view::sub().

neilmacintosh:
This PR adds the missing owner<T> and corrects the implementation of array_view::sub() to be as per earlier prototypes (in other words, to be correct).


gdr-at-ms:
Close #15

Issue #19, opened 2015-09-21
Broken links in README.md


igetgames:
The links provided in README.md to https://github.com/Microsoft/CppCoreGuidelines and [Standard C++ Foundation](https://github.com/Microsoft/GSL/blob/master/isocpp.org) (https://github.com/Microsoft/GSL/blob/master/isocpp.org) does not resolve. I couldn't find a repository named CppCoreGuidelines under the Microsoft organization (maybe it's private).

gdr-at-ms:
Fixed by #20

mvpete:
I think this link is still broken or private. https://github.com/Microsoft/CppCoreGuidelines

neilmacintosh:
I think it might be broken because we haven't forked off the Microsoft guidelines yet. Andrew?

PR #20, opened 2015-09-21
Fix the broken isocpp.org link in README.md.

igetgames:
This PR fixes the broken [isocpp.org](https://isocpp.org) link mentioned in #19.


gdr-at-ms:
Applied.

PR #21, opened 2015-09-21
Add add_gsl_test() function in tests/CMakeLists.txt

dacap:
With this function we can avoid duplicated code to add a new test.


neilmacintosh:
That's a very helpful PR David, thank you!

Issue #23, opened 2015-09-21
Missing header <stdexcept> in fail_fast.h 


galik:
Build failed under my own Makefile exposing missing header <stdexcept> (patch supplied)

neilmacintosh:
Thanks for the PR. This is now in.

PR #24, opened 2015-09-21
Issue #23   missing header <stdexcept> 

galik:
Missing header <stdexcept> in fail_fast.h #23


neilmacintosh:
Nice catch, galik. Thanks!

Issue #14, opened 2015-09-20
The `final_act` helper function name `finally` is misleading.


jbcoe:
The `final_act` helper function function, `finally`, is misleadingly named as it returns an object which, if not captured, will result in instant invocation of the given function object. 

The following code *looks* correct but database lookups and insertions will fail as the 'final-act' was invoked at [1].

    auto db_handle = my_db_api_open_db("My database name");
    finally([db_handle]{ db_api_close_db(db_handle); }); // [1]
    // do some lookups and insertions
  
I propose that `finally` is renamed to `make_final_act` or is replaced by a macro that captures the return value for the user into a uniquely named temporary variable (I prefer `make_final_act`).

trebconnell:
FYI I opened a [matching issue](https://github.com/isocpp/CppCoreGuidelines/issues/141) on isocpp gsl and it was closed.

neilmacintosh:
As per the issue on the Core Guidelines (https://github.com/isocpp/CppCoreGuidelines/issues/141), the name is considered fine as is. Thanks for the suggestion.

PR #32, opened 2015-09-24
Change finally() to make_final_act()

kernhanda:
Fix issue #14

### DIFF
--- a/include/gsl.h
+++ b/include/gsl.h
@@ -59,9 +59,9 @@ private:
     F f_;
 };
 
-// finally() - convenience function to generate a Final_act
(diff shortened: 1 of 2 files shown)


trebconnell:
This is an issue with [GSL.Util](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#gslutil-utilities).  I think this is a very good change, but perhaps it'd be better to fix it there first so that we remain conformant.

neilmacintosh:
As per the resolution of (https://github.com/isocpp/CppCoreGuidelines/issues/141), we are happy with the name "finally" for now. Thanks.

Issue #40, opened 2015-09-24
Add generated files to .gitignore


trebconnell:
I ran cmake to generate a VS solution and built, and got a ton of generated files not ignored by .gitignore.  Would be good to add these files.

trebconnell:
As per pull request feedback would be best to make directory for generated files and ignore that.

trebconnell:
Ah, I'm unfamiliar with CMake, fixed the bulk of this by issuing CMake to generate files out of source.  Might be nice to issue a warning here like llvm, but that's very polish.

PR #41, opened 2015-09-24
Fix issue #40: Add generated CMake and Visual Studio files to .gitignore

trebconnell:
Add cmake and build generated files to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,18 @@
 tests/unittest-cpp
+tests/*.dir/
+
+# CMake generated files
+CMakeFiles/
+*.cmake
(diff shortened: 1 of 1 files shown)


tamaskenez:
We should not ignore `*.cmake` files. In the future there may be handwritten ones.

Ignoring `cmake_install.cmake` would be a better idea. But on the other hand, instead of collecting all generated files of all compilers in the world, it would be better to agree on the name of the build directory and ignore just that one (for example `/build/`).

trebconnell:
Agreed on all counts. Especially dedicated directory for generated files. I'll open an issue for that.

Issue #39, opened 2015-09-24
Dont use #pragma once as it is not standard


viboes:

trebconnell:
I think it'd be best to have both #pragma once and header guards, as afaik many compilers optimize for #pragma once.

viboes:
No problem as soon as the file has the guards. Any conforming compiler should ignore the not supported pragmas.

michaelbprice:
I absolutely agree with fixing this issue.  Please fix this one.  In fact, the CppCoreGuidelines that are the primary reference to the GSL states that header files should have include guards.

https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rs-guards

This library should at least try to be consistent with the guidelines provided there.

neilmacintosh:
We’d be happy to look at a PR if anyone would like to put one together.

neilmacintosh:
this has now been fixed, thanks Treb Connell!

Zardoz89:
> No problem as soon as the file has the guards. Any conforming compiler should ignore the not supported pragmas.

What compiler don't support #pragma once ? 

I ask because I have curiosity, nothing more. As far I know, G++ , CLang and VStudio support it.

PR #44, opened 2015-09-24
Fix issue #39: Add header guards

trebconnell:


gdr-at-ms:
As suggested elsewhere, we want to prefix GSL-specific macros with `GSL_`.  So, can you prefix the include guards with `GSL_`?

trebconnell:
Yes, there's definitely possibility for name-conflict here.

trebconnell:
Should have had this in it's own branch.  Going to close this PR and reopen once I've fixed my branch.  Still learning the ins and outs here...

Issue #45, opened 2015-09-24
comparing two maybe_null_dbg's can cause fail_fast


trebconnell:
Because one of them is implicitly converted to T() causing a call to get()

trebconnell:
Fix is merged in

PR #46, opened 2015-09-24
Fix issue #45: comparing two maybe_null_dbg's can cause fail_fast

trebconnell:
Add tests for maybe_null_dbg and fix bug where comparing two of them results in an implicit conversion to T, and a call to get().


trebconnell:
Squeezing another small change in here.

neilmacintosh:
Looks good, thanks trebconnell!

Issue #49, opened 2015-09-25
Cannot construct from untested maybe_null_dbg


trebconnell:
Cannot construct a maybe_null_dbg from an untested maybe_null_dbg, because get() is used in the copy constructor.

neilmacintosh:
Fixed by commit 444bf96.

Issue #50, opened 2015-09-25
maybe_null_dbg::operator== can be ambiguous in TestMaybeNullCasting


trebconnell:
This is a regression caused by my pull request #46.  Can be reproduce by changing 
`template<class T> using maybe_null = maybe_null_ret<T>;`

to `maybe_null_dbg<T>;` and compiling.  TestMaybeNullCasting will fail to compile.

trebconnell:
fixed

neilmacintosh:
Thanks for fixing this.

PR #51, opened 2015-09-25
Fix issues #49 #50

trebconnell:
Grouped these all into one PR, as they're small and related.  In addition to not allowing not dereferencable types `ptee_size_` was adding unnecessary size overhead to maybe_null.


neilmacintosh:
The PR appears to fail travis CI and have conflicts. Do you want to update it and I will take another look.

trebconnell:
Closing while I fix build errors on non-msvc platforms

PR #56, opened 2015-09-25
Unsignedcheck

trebconnell:
…o add additional assert at source of signed type.

I found it easier to find the source of error with the additional assert.


neilmacintosh:
In this case, the ineffective static_assert in the code was misleading (and is a leftover from a misunderstanding of the design by an early implementer). I believe there should be no requirement on size_type to be unsigned.

Of course, as I say that, I suspect some of our existing static and fail_fast asserts probably assume that size_type is unsigned...but that's a different matter.

PR #57, opened 2015-09-25
We should be using standard algorithms where possible.

kernhanda:
Use of algorithms in the STL should be promoted where possible.
Also fixed up some whitespace issues.

### DIFF
--- a/include/array_view.h
+++ b/include/array_view.h
@@ -28,6 +28,7 @@
 #include <utility>
 #include <array>
 #include <iterator>
+#include <algorithm>
 #include "fail_fast.h"
 
 #ifndef _MSC_VER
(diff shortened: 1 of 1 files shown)


neilmacintosh:
In principle, I like the idea of using STL algorithms where appropriate. However, the PR appears to fail the travis-ci checks. Additionally, it changes the line endings in use in the existing files. Both of these issues would need to be corrected.

kernhanda:
Fixed the whitespace changes and fixed the build break.

neilmacintosh:
Thanks...unfortunately it appears to have conflicts again now. Once you have them resolved I will take another look.

kernhanda:
Fixed conflicts.

neilmacintosh:
Looks like you have failing tests now.

neilmacintosh:
My mistake...the test failure was not yours! It's green again. However, I'm still seeing line ending changes that look to be purely whitespace diffs....can you address that?

kernhanda:
I don't see any whitespace changes in the latest diff. Even checking it out through the command line results in no indication of whitespace changes. I'm taking a look at https://github.com/Microsoft/GSL/pull/57/files.

neilmacintosh:
Agreed, the diff lost the whitespace changes now. So it all looks good to me. Thanks for the change @kernhanda!

PR #59, opened 2015-09-25
Rank and dimensions should be size_t.

kernhanda:


trebconnell:
Looks good to me

gdr-at-ms:
Fixed #59.

PR #60, opened 2015-09-25
Fix static_asserts to better match error strings and updated tests accordingly.

kernhanda:


trebconnell:
I had a PR #56 out with a related fix that got closed, as it's the error string that's wrong not the check.

neilmacintosh:
The static assert is fine, its error message is unfortunately misleading. I've opened https://github.com/Microsoft/GSL/issues/67 to track getting this fixed.

PR #63, opened 2015-09-25
not_null and maybe_null variants should only work on pointer types.

kernhanda:


trebconnell:
[GSL.View](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#gslview-views) specifies that not_null can be a non-pointer type.  My PR #51 fixes some issues around this.

kernhanda:
There are a number of these issues that are cropping up, all of which beg the questions: how on sync do we want the two and which one usually has higher precedence?

trebconnell:
I've largely moved over to raising issues on isocpp since this is theoretically just an implementation of that (though it's very vague about a lot of things).  I feel this implementation is sitting on a foundation that may be shifting a lot, as the doc still has a lot of details that must be fleshed out.

gdr-at-ms:
By design, we expect `not_null` to be applicable to non-raw pointer types.

Issue #61, opened 2015-09-25
Brace style is inconsistent.


kernhanda:
There are places where we do
```C++
struct Foo {
...
};
```
And others where we do 
```C++
struct Bar
{
...
};
```
We should pick one and stick with it.

tamaskenez:
see my comment #55

neilmacintosh:
Agreed that we need to be consistent about formatting/style. We definitely work towards this.

Issue #55, opened 2015-09-25
Some files use tabs others use spaces


trebconnell:
'array_view.h' uses tabs 'gsl.h' uses spaces.  If my input matters I'd go with spaces :)

neilmacintosh:
Yes, and it is annoying (and probably at least partially my fault).
Personally, I would go with spaces too but I think all that really matters here is that we are consistent.

If you want to submit a PR to correct this (I closed the previous one for an unrelated reason), it would be best to do so separately from any functional changes.

tamaskenez:
What about clangformatting'em all? Maybe the admins could agree on a `.clang-format` file and add it to the repo.

trebconnell:
I agree that an auto formatter would be useful.  I don't have experience with clang format, but have used others in the past with a good experience.

neilmacintosh:
We do need consistency and will get there with future checkins.

PR #64, opened 2015-09-25
Add .clang-format file and format source files accordingly.

kernhanda:
To apply .clang-format to the codebase, the following
command can be run:

```git ls-files -- *.cpp *.h | xargs clang-format -i -style=file```

Addresses issues #61, #55


avlasyuk:
@kernhanda 
@neilmacintosh 
Would you mind adding some sensible column limit? It is currently hard to read code on GitHub because of long lines. Have to scroll text horizontally.

kernhanda:
I don't have a problem with that. @neilmacintosh -- I haven't seen any traction on this PR. Is there a chance it'll get merged in?

Also, @avlasyuk @neilmacintosh do you have any suggestions for column limits? At work, we use 140. Personally, however, I code often in the terminal in a split window, where a column limit of ~90 tends to be far better.

avlasyuk:
@kernhanda 100 looks perfect to me.

May I suggest to add those two lines as well:

    Cpp11BracedListStyle: true
    AllowShortCaseLabelsOnASingleLine: SFS_All

neilmacintosh:
The PR is not going to be merged.

For a start, it currently has conflicts. Second, it includes both the .clang-format file and all the reformatting. Plus, the reformatting appears to have changed all the line endings, which is not desirable for this repo.

I have no objection to including a .clang-format file in the repo. But I'd prefer to adopt a "ensure format as-changes-come-in" strategy as per #68. I will admit that I am far less concerned about consistent layout than a lot of people so I would rather have more readable diffs and a gradual move to a super-consistent layout.

What I will do is merge that .clang-format file as a starting point and if there are improvements that could be made to it, then feel free to offer PRs against it. A good example would be a column-width limitation, which seems pretty sensible given the limitations of viewing via GitHub that @avlasyuk pointed out.

Sorry for the delay on this one.

socantre:
@neilmacintosh is there a preferred column width?

neilmacintosh:
I think 100 sounds pretty reasonable and seems to solve the GitHub viewing issue (at least on my browser, on my machine ;-) )

PR #68, opened 2015-09-26
Add .clang-format file

socantre:
To have any hope of achieving consistent formatting there needs to be a standard and a mechanism for enforcing it. This is a first step in that direction.

It should be noted that this pull request _does not_ reformat any existing code. Instead contributors should use clang-format to format portions of code that they modify.

This format specification is intended to mimic the existing style insofar as the current source has any consistent style. For example the determination as to whether ref qualifiers should bind to the
type or variable name was made by trying both and seeing which produced fewer changes. One exception is the use of tabs vs. spaces, which was decided based on a comment by neilmacintosh [here][1].

To use clang-format with Visual Studio [download][2] and install the clang-format plugin. In VS ctrl-r, ctrl-f formats the current line or text selection.

[1]: https://github.com/Microsoft/GSL/issues/55
[2]: http:llvm.org/builds/

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,21 @@
+ColumnLimit: 0
+
+UseTab: Never
+IndentWidth: 4
+AccessModifierOffset: -4
+NamespaceIndentation: Inner
+
+BreakBeforeBraces: Allman
+AlwaysBreakTemplateDeclarations: false
+BreakConstructorInitializersBeforeComma: true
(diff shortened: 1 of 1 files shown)


neilmacintosh:
I appreciate that this is just the .clang-format file and not a reformatting of the entire repo (which I'd rather do piecemeal, as we change things).

This looks like a good starting point, thanks for the PR!

PR #70, opened 2015-09-27
not_null and maybe_null variants should only work on nullptr-assignable types.

kernhanda:
This is in accordance with the GSL.View guidance on not_null and maybe_null
types in the CppCoreGuidelines document.

It is an update of the commit made for #63, except it's been updated so that the check is performed on all types that are actually nullable.


neilmacintosh:
This looks good, thanks!

Issue #75, opened 2015-09-28
GSL.assert macros should be methods


nibroc:
Considering the intent of Expects/Ensures to eventually be part of declarations, I can understand why they were made macros, but until a tool (or perhaps compiler extension, since I'm not sure it's possible to implement other than a namespace colliding macro?) actually supports placing them in a declaration, I think they would be better as functions.

As in the guidelines ([ES.31](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Res-macros2)), macros violate the usual scope rules, they can behave oddly with parameters, and they should be avoided unless necessary. Given this, I believe that as long as Ensures/Expects are being used in their current form, they would be better suited to functions.

nibroc:
Pull request here: https://github.com/Microsoft/GSL/pull/76

neilmacintosh:
Macros were chosen as an implementation mechanism while awaiting contract support in the language. Macros have the single advantage over functions of being able to capture line-number and source file information pertaining to the contract. I admit it is confusing that the current implementation doesn't do this, but this is only because they were added quickly and were expected to be revised.

PR #76, opened 2015-09-28
Changed GSL.assert macros to functions

nibroc:
As per Issue #75, this changes macros to functions.

Sorry if the pull request immediately after the issue is against the spirit of the contribution guidelines. It was an easy change, so I went ahead and did it, but I'll happily withdraw this if there's issues/disagreement.


neilmacintosh:
See discussion in corresponding issue.

Issue #66, opened 2015-09-25
not_null<>: Missing assignment from related types ? 


martinmoene:
For not_null<>, if there's a constructor for related types:

```
template <typename U, typename Dummy = std::enable_if_t<std::is_convertible<U, T>::value>>
not_null(const not_null<U> &other);
```
I would also expect the assigment:
```
template <typename U, typename Dummy = std::enable_if_t<std::is_convertible<U, T>::value>>
not_null<T> & operator=( not_null<U> const & other );
```
However it's not present.

neilmacintosh:
Yes, it should have one. Feel free to submit a PR if you would like to add it yourself.

neilmacintosh:
This should now be corrected.

PR #77, opened 2015-09-28
Add assignment from related types

martinmoene:
Fixes issue #66.

Please consider test style as a [specification](https://github.com/martinmoene/gsl-lite/blob/master/test/not_null.t.cpp) ;) 

That helped me notice that this assignment was missing.


martinmoene:
mmm, looks like this PR better makes  place for PR #78.

neilmacintosh:
Thanks for the PR, but as you observed, there was #78 covering the same issue which went in first.

PR #86, opened 2015-09-28
it's better to pass a small (e.g. 16 bytes) object by value then by c…

kevgs:
…onst reference

Less typing and less alias analysis for compiler. Same as narrow_cast https://github.com/Microsoft/GSL/blob/master/include/gsl.h#L74


neilmacintosh:
Looks good to me! Thanks for the PR!

PR #88, opened 2015-09-29
Fix issue #49

trebconnell:
In my opinion you shouldn't have to test maybe_null_dbg's before using them to assign or construct another maybe_null_dbg.


neilmacintosh:
That seems like a pretty reasonable observation to me. The change looks good. Thanks for the PR!

Issue #42, opened 2015-09-24
Final_act has constructor taking `const Final_act&&`


trebconnell:
Shouldn't it be `Final_act(Final_act&& other)` so that it's a proper move constructor?  Also shouldn't we std::move from other.f_ into f_?

neilmacintosh:
Yes, it should.

john-lynch:
Related, but I'm pretty sure that the current implementation of `finally` and `Final_act` rely on RVO for correctness, which seems like a bad idea to me. Even if we use `std::move` in a proper copy move constructor as suggested here, if a temporary `Final_act` was created (because RVO didn't happen even though, yes, on all major compilers it would), wouldn't the destruction of that temporary `Final_act` cause `f` to be invoked immediately (and then invoked again later when the local, non-temporary `Final_act` goes out of scope)?

Anyway, I would think Final_act would have to look like this so that we don't rely on an optional compiler optimization for correctness:
```c++
// Final_act allows you to ensure something gets run at the end of a scope
template <class F>
class Final_act
{
public:
    explicit Final_act(F f) : f_(std::move(f)), invoke_(true) {}
    
    Final_act(Final_act&& other) : f_(std::move(other.f_)), invoke_(true) { other._invoke = false; }
    Final_act(const Final_act&) = delete;
    Final_act& operator=(const Final_act&) = delete;
    
    ~Final_act() { if (invoke_) f_(); }

private:
    F f_;
    bool invoke_;
};
```
My apologies if I have misunderstood how moving a lambda works, but I'm pretty sure I've got this right. The following code results in in the lambda being invoked twice (tested on multiple compilers):
```c++
#include <iostream>

using namespace std;

int main()
{
	auto l = [](){ cout << "Hello, lambda!" << endl; };
	auto ll = std::move(l);
	l();
	ll();
}
```

trebconnell:
RVO isn't an optional optimization. It's actually required by the standard, and has been for a while.

neilmacintosh:
Final_act does need some improvement, but not for RVO reasons (as Treb noted).
#11 calls out one issue in its implementation and there was a related PR but it contained other changes that didn't reflect the design, so had to be rejected.

If you'd like to offer a PR that addresses #11 feel free to do so. i won't be able to take a look at fixing it for a day or so.

john-lynch:
Thank you for such a prompt response!

[...]

PR #90, opened 2015-09-29
Fixing move constructor/semantics of Final_act

john-lynch:
The move constructor of Final_act is updated to accept a non-const r-value reference. It also properly moves from the r-value function object member to the function object member of the object being constructed during the move operation. It now has a boolean to indicate whether or not the function object should be invoked in the destructor so that temporary objects that have been moved from do not invoke the function object improperly, removing dependence on copy elision for correct behavior. Finally, a test is added to verify that Final_act objects can be moved, if necessary. This resolves the issues discussed in issue #42.

### DIFF
--- a/include/gsl.h
+++ b/include/gsl.h
@@ -50,16 +50,17 @@ template <class F>
 class Final_act
 {
 public:
-    explicit Final_act(F f) : f_(std::move(f)) {}
-    
-    Final_act(const Final_act&& other) : f_(other.f_) {}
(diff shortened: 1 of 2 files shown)


john-lynch:
This also resolves issue #11. It handles the move safety aspect of PR #13, which was rejected because the exception safety aspect was not desired.

neilmacintosh:
This looks good to me. Thanks for the PR John!

PR #93, opened 2015-09-29
writting <> with every string_view type is non productive and also di…

kevgs:
…ffers from what C++ Core Guidelines proposes

### DIFF
--- a/include/string_view.h
+++ b/include/string_view.h
@@ -57,17 +57,10 @@ using wzstring = wchar_t*;
 template <class CharT, size_t Extent = dynamic_range>
 using basic_string_view = array_view<CharT, Extent>;
 
-template<size_t Extent = dynamic_range>
-using string_view = basic_string_view<char, Extent>;
-
-template<size_t Extent = dynamic_range>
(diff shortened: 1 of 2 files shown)


kevgs:
Looks like clang++-3.6 isn't working on travis server for some reason.

tamaskenez:
@kevgs https://github.com/travis-ci/apt-source-whitelist/issues/156

neilmacintosh:
For now, the <> implements the design correctly and is just a requirement of the language. It may be annoying but it does allow us to write string_view<10> (for example) when we want to specify a fixed-size string_view,

tamaskenez:
What about separate string_view and string_view_n? I think dynamic-range views are far more common.

    template<size_t Extent>
    using string_view_n = basic_string_view<char, Extent>;

and

    using string_view = basic_string_view<char>;

neilmacintosh:
Sure, they are. We've certainly considered that option (with a couple of different name variants). I'll see what people think on the committee as we propose this type for standardization and if we end up with that type of naming scheme, then it will be reflected here.

PR #95, opened 2015-09-29
Making finally work for non-const lvalues and independent on rvo/nrvo

AntonBikineev:
Fixing 94


neilmacintosh:
This has already been resolved by #90. Thanks for the PR anyway!

PR #99, opened 2015-09-29
Making at work for container with not random access iterators

AntonBikineev:


gdr-at-ms:
We should stay closer to the spirit of the `STL` and not provide this operation when it is not constant time.

Issue #103, opened 2015-09-29
Conditional inclusion of <exception> and <stdexcept> in fast_fail.h


davidhunter22:
fast_fail.h includes both <exception> and <stdexcept> unconditionally. It should probably be

#if defined(SAFER_CPP_TESTING)
#include <stdexcept>
#else
#include <exception>
#endif

to avoid including unnecessary header files. ( modules any day now!!! )

neilmacintosh:
Yes, it probably should. This would be a simple PR...;-)

neilmacintosh:
Note that <exception> is always required as when not throwing, we call std::terminate(), which is declared in <exception>.

neilmacintosh:
Thanks to Rico's PR this is now fixed.

PR #106, opened 2015-09-30
Commit to address issue #103...

RicoAntonioFelix:


neilmacintosh:
I believe the header <exception> is always required....as it contains the declaration of std::terminate(). You will need to update the PR before I can accept it.

RicoAntonioFelix:
I tried to do the update on the commit, but it seems I'm doing something wrong. I'll do the simpler thing and create a new PR...

Issue #89, opened 2015-09-29
not_null<unique_ptr<>> doesn't work


trebconnell:
Since guidelines specify: "not_null<T> ... T can be any type for which ==nullptr is meaningful." this should work

trebconnell:
One additional sticking point here is what will get() return?  Can't return the unique_ptr by value.  I would say unique_pointer<>::pointer.  How about for shared_ptr.  That doesn't have a ::pointer typedef.  Of course that *can* return by value, but would incur the cost of locking and incrementing.

galik:
According to the guidlines for passing *smart pointers* it seems that the only reason we should be passing a *smart pointer* to a function is when that fuction mosdifies the *ownership* in some way. Otherwise, the recommenation is that we pass a *raw pointer*.

> ### <a name="Rr-smartptrget"></a> R.37: Do not pass a pointer or reference obtained from an aliased smart pointer

> **Reason**: Violating this rule is the number one cause of losing reference counts and finding yourself with a dangling pointer.
> Functions should prefer to pass raw pointers and references down call chains.
> At the top of the call tree where you obtain the raw pointer or reference from a smart pointer that keeps the object alive.
> You need to be sure that smart pointer cannot be inadvertently be reset or reassigned from within the call tree below



Bearing that in mind I am not sure that `not_null` should be passing smart pointers into functions at all.

After all a function receiving a `not_null` generally wants to be able to make calls against the pointer without having to first check of its validity. It is unlikely a function receiving a `not_null` would want to deal with its *ownership* and, even if it did, would we want to encourage that? After all such a function would have the power to *reset* the pointer breaking its *not null* contract.

Maybe there are use-cases I have missed but it seems to me the main use for the `not_null` (with respect to *smart pointers*) is for *down-calling* and therefore for passing only the *raw pointer* that the *smart pointer* is managing.

Therefore we might not want a `not_null` to *contain* a *smart pointer*, only its associated *raw pointer*.

To that effect we could create `not_null` *specializations* for both `unique_ptr` and `shared_ptr` whereby it captures their managed *raw pointer* using `get()` and implements its *not null* invariant on construction.

kkoenig:
The real question that needs to be asked here is *should* `not_null<T>` *behave as an alias of T*.  

If so, `not_null<T>::get` should only exist if `T::get` exists. This would be especially interesting if it was implemented so that 
```
[...]

PR #109, opened 2015-09-30
not_null class makes an unnecessary copy,can't be used effectively wi…

bashrc-real:
…th unique_ptr

Passing T by const ref instead of by value. This is important specially to prevent perf implications when this class is used for heavy to copy objects. A common example would  be shared_ptr. Another way to prevent two versions of the constructor is to have a single constructor and move the value. <code> not_null(T t) : ptr_(move(t)) {...} </code> This gives best of both worlds. For most cases I would have preferred this. But this can result in slightly bigger code since for the cases where this class is invoked from a copy of he object, the caller has to do the job of creating the copy.

### DIFF
--- a/include/gsl.h
+++ b/include/gsl.h
@@ -111,7 +111,8 @@ class not_null
 {
     static_assert(std::is_assignable<T&, std::nullptr_t>::value, "T cannot be assigned nullptr.");
 public:
-    not_null(T t) : ptr_(t) { ensure_invariant(); }
+    not_null(const T &t) : ptr_(t) { ensure_invariant(); }
+    not_null(T &&t) : ptr_(std::forward<T>(t)) { ensure_invariant(); }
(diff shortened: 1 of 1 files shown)


bashrc-real:
This issue looks somewhat related to https://github.com/Microsoft/GSL/issues/89
This change would still be valid if we allowed creating this class with smart pointers but copying the underlying pointer in the constructor call itself instead of keeping the object around.

neilmacintosh:
If someone has declared a not_null<unique_ptr>, then that object should take ownership of the unique_ptr passed to the ctor. So we should certainly do a better job of moving the ctor argument in to achieve that outcome. However, I think taking the argument as a const& is the wrong change.

bashrc-real:
In that case the other constructor (r-value constructor) that I had provided in my change would come handy. When you say taking the argument as const & is wrong, do you also giving a r-value constructor is wrong?

Issue #91, opened 2015-09-29
Rename Final_act to final_act to follow standard library conventions.


neilmacintosh:

neilmacintosh:
Fixed.

PR #110, opened 2015-09-30
Renamed Final_act to final_act as per issue #91.

neilmacintosh:


gdr-at-ms:
It is all yours! :-)

neilmacintosh:
Thanks :-)

Issue #101, opened 2015-09-29
Split gsl.h


davidhunter22:
I noticed that classes like not_null are declared in gsl.h. gsl.h also includes array_view.h and string_view.h. These headers then include a large number of other headers. In the interests of compile times can all the definitions in gsl.h be moved to a separate header file which gsl.h then includes. gslh then becomes a very small file that just includes all the other headers. Users can then choose to use this or the components headers.

neilmacintosh:
It's a reasonable suggestion. The current structure was not particularly carefully designed, and having components that are "more than a using" in their own headers seems sensible. If you have the time and inclination, I would be happy to consider a PR that breaks out the separate headers...

neilmacintosh:
@davidhunter22 PR #215 breaks out more separate headers for a more useful physical design. I'm going to claim that addresses this issue sufficiently for now. Feel free to reopen the issue (or offer a PR) if you think there is further useful decomposition to be done.

PR #112, opened 2015-10-01
PR to address issue #101

RicoAntonioFelix:
On an attempt to strip down the gsl.h, in this PR I have first removed the final_act class from it into its own file and renamed it to correspond to the guidelines...

### DIFF
--- a/include/final_action.h
+++ b/include/final_action.h
@@ -1,0 +1,71 @@
+/////////////////////////////////////////////////////////////////////////////// 
+// 
+// Copyright (c) 2015 Microsoft Corporation. All rights reserved. 
+// 
+// This code is licensed under the MIT License (MIT). 
+// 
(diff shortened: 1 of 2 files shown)


neilmacintosh:
I appreciate the name change for final_action and breaking it into its own file. However, I would prefer it if we could leave the original member names and have fewer empty-comment lines as per the original. Minimizing non-functional changes makes it easier to review a PR :-)

Also, I suspect this file will need compiler-workaround macros (see those in gsl.h, for example) to stop the noexcept from breaking MSVC 2013.

neilmacintosh:
@RicoAntonioFelix did you want to try and address my comments and the conflicts this PR now has, or shall I close this out?

RicoAntonioFelix:
I think its best to close this out since @trebconnell is very active and doing a good job taking care of things... We should just let the types and library mature before doing a major restructuring of the module...

neilmacintosh:
Thanks for getting back to me on that @RicoAntonioFelix. I'll close this out for now.

PR #114, opened 2015-10-01
If T::get exists, define not_null<T>::get and return not_null<result_…

kkoenig:
Provides an example of how we could begin to treat `not_null<T>` as an alias for `T`

See comments on issue https://github.com/Microsoft/GSL/issues/89

### DIFF
--- a/include/gsl.h
+++ b/include/gsl.h
@@ -112,7 +112,7 @@ public:
     not_null(const not_null &other) = default;
 
     template <typename U, typename Dummy = std::enable_if_t<std::is_convertible<U, T>::value>>
-    not_null(const not_null<U> &other) : ptr_(other.get())
+    not_null(const not_null<U> &other) : ptr_(other)
     {
     }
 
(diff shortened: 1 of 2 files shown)


include/gsl.h
```diff
@@ -122,15 +122,15 @@ class not_null
...
-#ifdef _MSC_VER
-        __assume(ptr_ != nullptr);
-#endif
-        return ptr_;
-    } // the assume() should help the optimizer
-
-    operator T() const {  return get(); }
-    T operator->() const { return get(); }
+    // if T has .get() forward its value wrapped in not_null
+    template<typename U = T, typename = decltype(std::declval<U>().get())>
```
trebconnell: nit: most places in this code doing enable_if type things use the name Dummy for the template parameter

gdr-at-ms: I actually find NOT naming a parameter as a better indication that it is dummy.

trebconnell: I do too, just basing this off the rest of the code

neilmacintosh: In some cases, not naming a template parameter causes a problem for MSVC 2013, so the usage probably spread further than it needed to. If you are making changes and don't want the name and it doesn't cause a target compiler a problem....then feel free to do without it.

JordanMaples:
Maintainer's call: sorry about the delay. This has been superseded as the Guidelines have evolved.
Closing.
Thanks

PR #115, opened 2015-10-02
Add column limit to .clang-format

socantre:


neilmacintosh:
Looks good to me. Thanks @socantre.

PR #116, opened 2015-10-02
Optimize final_act for general case where move is not called

trebconnell:
This prevents storing and testing the bool.


trebconnell:
Closing this, the build errors will be fixed by another change I want to make to final_act

PR #117, opened 2015-10-02
Fix moving a final_act twice

trebconnell:


neilmacintosh:
Thanks Treb, nice fix!

Issue #121, opened 2015-10-03
std::vector constructor arguments transposed in TestConstructFromStdVector


amdn:
In string_view_tests.cpp there is a test that has transposed the arguments to the std::vector constructor.  The test compiles and passes, though probably not as intended.
Instead of a vector with 5 char elements with literal value 'h' (decimal 104 in UTF-8) and due to implicit conversion, the vector is constructed with 104 char elements with literal value 5.

http://en.cppreference.com/w/cpp/container/vector/vector

    TEST(TestConstructFromStdVector)
        {
            std::vector<char> vec('h', 5);
            string_view<> v = vec;
            CHECK(v.length() == vec.size());
        }

neilmacintosh:
Ha ha ha ha. Nice catch. I think the test does actually pass as intended (the string_view's length matches that of the vector). I think the initialization of the vector is not as intended. Would you like to offer the fix?

amdn:
Neil - looks like kkoenig has a fix for this one - being new to git I
wasn't sure how to do it - I see I'd need to fork the project first.  I'll
know what to do next time.

neilmacintosh:
No problem @amdn. Thanks for catching it and thanks to @kkoenig for the fix.

PR #122, opened 2015-10-04
First arg of vector ctor should be num elements, second should be val…

kkoenig:
…ue. Fixes issue #121


amdn:
Looks good to me :+1:

neilmacintosh:
Looks great. Thanks @kkoenig (and @amdn for the original report).

PR #127, opened 2015-10-05
Fix #119 express preconditions with Expects

### LINKED ISSUE #119: Use Expects within the GSL for preconditions
fail_fast_assert is used in many places, but I think it'd be preferable to use Expects

This is dependent on #101, as Expects is currently defined in gsl.h, which prevents usage in array_view.h

trebconnell:
I had to move gsl.assert into it's own header so that array_view.h could use it.  I unhappily chose the name assert.h.  I don't like this name, because it conflicts with assert.h in the cstdlib.  I'm open to alternative naming suggestions.

In a few places I removed temporary variables, so that Expects will be able to move to function declarations.  I also had to remove description strings in some places, as Expects doesn't support these.


trebconnell:
Tried to fix Expects string with non-standard variadic macros.  This unsurprisingly didn't work.  I'm going to close this while I iterate for a bit.  Will reopen when I've figured out a (preferable standard) method to make this work.

trebconnell:
I made Expects and Ensures into functions rather than macros so that I could overload.  I assume they were macros to allow grabbing file and line, but since that isn't happening yet and we need overloading I think it's best for them to be functions.

neilmacintosh:
This was considered when they were originally described in the Core Guidelines, but the feeling was that they need to pick up file-and-line automatically, so need to be macros (for now...ideally they end up moving into a language contracts facility). So let's stick with macro-form for now.

trebconnell:
The difficulty is that I don't think I can portably have overloaded macros, so either a macro taking two args or taking one but not both.  I could have two macros with different names, but that's pretty ugly.

I don't really like any option.  I think requiring strings is excessive, as many statements will be obvious from the condition.

neilmacintosh:
For now the macro doesn't have to take the strings. It would be fine if they were just left in source next to the macro invocations as comments.

galik:
If you pass the macro's variable args immediately to a template function could you not fudge it that way?

I'm not sure about portability of this but it seems to work for me, unless I missed something:
```cpp
template<typename Arg1, typename Arg2, typename Arg3 = int>
void assert_processing(bool test, Arg1 arg1, Arg2 arg2, Arg3 arg3 = {})
{
	if(test)
		return;

	std::ostringstream oss;

	if(!arg3) // no message
		oss << "file: " << arg1 << " line: " << arg2;
	else
		oss << arg1 << ": file: " << arg2 << " line: " << arg3;

	gsl::fail_fast_assert(false, oss.str().c_str());
}

#define Expects(...)  assert_processing(__VA_ARGS__, __FILE__, __LINE__)
#define Ensures(...)  assert_processing(__VA_ARGS__, __FILE__, __LINE__)
```

trebconnell:
Yeah something like that should work given that we require one or more
arguments.  It's a bit nasty to use __VA_ARGS__ since it does act
differently on different compilers, but having it as the first parameter
does fix the issues I was having with gcc.

galik:
Its not particularly *fail friendly* because it allocates memory for `std::ostringstream`. If could me made more *fail friendly* with a static char buffer perhaps but it would require more coding. I just wanted to present the principle.

neilmacintosh:
I think this is now obsolete given that PR #215 addressed the same issue.

PR #141, opened 2015-10-13
Dev/annagrin/remove coordinate facade

annagrin:
Removed coordinate_facade class and moved its functionality to its former derived classes, index and strided bounds. Updated tests.


neilmacintosh:
I like the change Anna, but can you fix the build breaks for non-Microsoft platforms please?

PR #146, opened 2015-10-15
Dev/annagrin/remove coordinate facade

annagrin:
Removed class coordinate_façade, updated tests


neilmacintosh:
Looks great @annagrin , thanks for this first step in our cleanup effort!

PR #147, opened 2015-10-17
std::hash support for gsl::not_null.

samuel3:
Just added std::hash support: std::hash&lt;gsl::not_null&lt;T&gt;&gt;


neilmacintosh:
Looks good and seems entirely reasonable to me. Thanks for the PR @samuel3!

PR #149, opened 2015-10-19
Fix #148 ordering of bounds_iterator

### LINKED ISSUE #148: bounds_iterator is not ordered
The `operator<` of `bounds_iterator` is not antisymmetric, as demonstrated by this program:

``` c++
#include "gsl.h"
#include <iostream>
using namespace std;
using namespace gsl;
int main(){
    static_bounds<size_t, 2, 2> bounds;
    auto i1 = bounds.begin() + 1;
    auto i2 = bounds.begin() + 2;
    cout << (i1 < i2) << endl;
    cout << (i2 < i1) << endl;
}
```

Suggested fix: 1475af7a8095ac2e3195a3e9d3fd01281759764d

panacekcz:
Fixes #148.
The test case (taken from the issue description) is rather minimalistic.

### DIFF
--- a/include/array_view.h
+++ b/include/array_view.h
@@ -1075,6 +1075,8 @@ public:
 		{
 			if (curr[i] < rhs.curr[i])
 				return true;
+			if (rhs.curr[i] < curr[i])
+				return false;
 		}
 		return false;
 	}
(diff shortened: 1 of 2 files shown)


neilmacintosh:
I'm afraid I can't accept this PR without a CLA being signed. Plus, the codebase has moved around a bit since it was originally generated. So I'll close it out for now, but would be happy to consider a new one that targeted the current codebase.

PR #151, opened 2015-10-19
Removed arrow_proxy class, fixes bugs in reverse bounds_iterator

annagrin:


neilmacintosh:
Looks good to me, thanks Anna!

PR #152, opened 2015-10-19
Replaced index constructor from initializer list by a constructor fro…

annagrin:
…m static list

Conflicts:
	include/array_view.h


neilmacintosh:
Looks great...more nice cleanup. Thanks @annagrin!

PR #154, opened 2015-10-19
Removed specializations for Rank=1

annagrin:


neilmacintosh:
Love it, thanks @annagrin!

Issue #157, opened 2015-10-21
include/array_view.h: unused parameter 'ptr'


mattipee:
There were two instances of this a couple of days ago, one removed by "remove_rank1_specialization". The remaining instance is shown below:

In file included from GSL/include/gsl.h:22:
GSL/include/array_view.h:128:66: error: unused parameter 'ptr' [-Werror,-Wunused-parameter]
        constexpr index(const index<Rank, OtherValueType>& other, void* ptr = 0) noexcept

$ clang++ -v
clang version 3.8.0 (http://llvm.org/git/clang.git fd71891aeed72817f187647915ce7420ef6b2960) (http://llvm.org/git/llvm.git d63954dac10035f92809d09ad852caf9c128b729)
Target: x86_64-unknown-linux-gnu

neilmacintosh:
This has now been fixed thanks to the PR from @kevgs.

PR #158, opened 2015-10-21
fix clang warning on unused function parameter

kevgs:


kevgs:
Here is a corresponding issue https://github.com/Microsoft/GSL/issues/157

neilmacintosh:
Looks good to me! Thanks for fixing that up @kevgs!

mattipee:
Thank you, both.

PR #161, opened 2015-10-27
Fix gsl::ensure_z() to work for const char* strings

mattnewport:
Currently the `ensure_z()` function only works for `char` arrays, not for `const char*`. It appears the code was intended to work for `const char*` but had not been tested and did not compile. This fix allows the code to compile and adds a test.


neilmacintosh:
Thanks @mattnewport, for fixing this bug!

PR #163, opened 2015-11-02
array_view: explicitly initialize constexpr function variables

archshift:


neilmacintosh:
Thanks @archshift, nice tidy-up!

Issue #159, opened 2015-10-21
VS2013 Build is Broken


wbic16:
It appears that the README is wrong at the very least. While reviewing CI builds, I only see gcc and clang referenced. A list of commits and issues observed is given below.

Commit c5c8a25: last version that worked
Commit 831c692: produces 'noexcept' errors
Commit 2f25ace: lastest, still broken

Steps to reproduce:
1. Run `git reset --hard c5c8a25f34b1bb1cd0a418749cfd7b6095e48f67`
2. Compile; observe no errors or warnings.
3. Run `git reset --hard 831c6926df2dd17a691abf7d84137435ff172025`
4. Compile; observe "error C3646: 'noexcept' : unknown override specifier"
5. Run `git reset --hard 2f25ace31fe7bff78386aa55fd18371cf0d113d8`
6. Compile; observe "warning C4068: unknown pragma" and "error C2057: expected constant expression" errors in array_view.h.

Hope this helps, thanks!

neilmacintosh:
@annagrin can you look into this please?

subhamoyg:
Hi,

I tried VS2013 build on windows_x64 with cmake_3.2.3 but it failed to compile. Was not able to attach the build log, so pasting below. Sorry for the length of the build report.
===============================================================================
Microsoft (R) Build Engine version 12.0.21005.1
[Microsoft .NET Framework, version 4.0.30319.42000]
Copyright (C) Microsoft Corporation. All rights reserved.

Build started 10/24/2015 5:06:49 PM.
Project "<local_machine>\GIT_Repository\GSL\build-x64\ALL_BUILD.vcxproj" on node 1 (default targets).
Project "<local_machine>\GIT_Repository\GSL\build-x64\ALL_BUILD.vcxproj" (1) is building "<local_machine>\GIT_Repository\GSL\build-x64\tests\unittest-cpp\TestUnitTest++.vcxproj" (2) on node 1 (default targets).
Project "<local_machine>\GIT_Repository\GSL\build-x64\tests\unittest-cpp\TestUnitTest++.vcxproj" (2) is building "<local_machine>\GIT_Repository\GSL\build-x64\tests\unittest-cpp\UnitTest++.vcxproj" (3) on node 1 (default targets).
[...]

PR #169, opened 2015-11-05
Added workaround for MSVC 2013 compiler bug.

neilmacintosh:
This PR should fix the remaining compiler crash reported in #159.

### DIFF
--- a/include/string_span.h
+++ b/include/string_span.h
@@ -22,6 +22,16 @@
 #include "span.h"
 #include <cstring>
 
+// VS 2013 workarounds
+#ifdef _MSC_VER
+#if _MSC_VER <= 1800
+
+#pragma push_macro("GSL_MSVC_HAS_TYPE_DEDUCTION_BUG") 
+#define GSL_MSVC_HAS_TYPE_DEDUCTION_BUG 
+
+#endif // _MSC_VER <= 1800
+#endif // _MSC_VER
+
 namespace gsl
 {
 //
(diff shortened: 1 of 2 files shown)


include/string_span.h
```diff
@@ -22,6 +22,16 @@
 #include "span.h"
 #include <cstring>
 
+// VS 2013 workarounds
+#ifdef _MSC_VER
+#if _MSC_VER <= 1800
+
+#pragma push_macro("GSL_MSVC_HAS_TYPE_DEDUCTION_BUG")
```
gdr-at-ms: Do you expect many definitions of `GSL_MSVC_HAS_TYPE_DEDUCTION_BUG`?
If not, then the push here will hide conflicts that should be loudly reported to consumers if they ever step on it.

include/string_span.h
```diff
@@ -178,4 +213,15 @@ template <size_t Max = dynamic_range>
 using wzstring_builder = basic_zstring_builder<wchar_t, Max>;
 }
 
+// VS 2013 workarounds
+#ifdef _MSC_VER
+#if _MSC_VER <= 1800
+
+#pragma pop_macro("GSL_MSVC_HAS_TYPE_DEDUCTION_BUG") 
+#undef GSL_MSVC_HAS_TYPE_DEDUCTION_BUG 
+
```
gdr-at-ms: The `#undef` after the pop will remove any previous binding that the push was attempting to preserve.
I think it is better to just remove the push and the pops.

rkhlebnikov:
Please add `inline` to the definitions - otherwise linker errors are imminent :) Thanks for the workaround!

rkhlebnikov:
Actually, added a pull request for it.

include/string_span.h
neilmacintosh: Good point. I'm going to clean up all the config macros in a separate pass shortly. Given no one else should be using it anyway it's harmless for now.

PR #179, opened 2015-11-12
Added copy assignment operators

vladon:
`BoundsRanges` and `strided_bounds` have copy constructor, but have not copy assignment operators


neilmacintosh:
Nice cleanup @vladon, thanks!

PR #187, opened 2015-11-13
Add workarounds for g++ 4.9

nguarracino:
This change uses some of the MSVC workarounds for older g++ compilers (gcc < 5.1) which have similar issues with constexpr.

Tests pass on:
- gcc-4.9.2/Linux
- gcc-5.2.1/Linux
- clang-3.6.1/Linux
- MSVC 2015/Windows

### DIFF
--- a/include/gsl.h
+++ b/include/gsl.h
@@ -45,7 +45,7 @@
 
 #endif // _MSC_VER
 
-#if (__GNUC__ < 5) && !defined(__clang__)
+#if defined(__GNUC__) && (__GNUC__ < 5) && !defined(__clang__)
 #define constexpr /* nothing */
 #endif
 
--- a/include/gsl.h
+++ b/include/gsl.h
@@ -45,6 +45,10 @@
 
 #endif // _MSC_VER
 
+#if (__GNUC__ < 5) && !defined(__clang__)
+#define constexpr /* nothing */
+#endif
+
 // In order to test the library, we need it to throw exceptions that we can catch
 #ifdef GSL_THROWS_FOR_TESTING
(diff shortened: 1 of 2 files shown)


include/gsl.h
```diff
@@ -45,6 +45,10 @@
 
 #endif // _MSC_VER
 
+#if (__GNUC__ < 5) && !defined(__clang__)
```
gdr-at-ms: You want to test first that `__GNUC__` is defined, via `defined(__GNUC__)`.

include/span.h
```diff
@@ -60,6 +60,10 @@
 
 #endif // _MSC_VER
 
+#if (__GNUC__ < 5) && !defined(__clang__)
+#define constexpr /* nothing */
```
gdr-at-ms: Same comment here and elsewhere.

gdr-at-ms:
We do really want to set the base line at C++14.

nguarracino:
@gdr-at-ms I understand... problem is, I'm targeting an embedded device with a slightly older compiler. No worries if you want to keep the workarounds to a minimum, I can just keep this as a local patch and wait for my vendor to update to gcc 5.

neilmacintosh:
@nguarracino I appreciate the effort (and interest in using the GSL!) but I'd prefer if you kept the PR in a fork (especially if you think it's likely you'll move to GCC 5 anyway). 

As per the comment in the Readme.md, our aim here is to rely on C++14-compliant compilers. We make an exception for MSVC for the reason that we have a lot of MSVC users internally . But even there we are applying strong pressure on its C++14 compliance so we can drop the workarounds as soon as possible.

PR #190, opened 2015-11-13
Dev/annagrin/remove basic span

annagrin:
Removed basic_span class.


neilmacintosh:
Looks good to me. Thanks @annagrin!

PR #191, opened 2015-11-14
Suggested static_pointer_cast

samuelwtheck:
The current implementation of `gsl::not_null` provides a constructor to convert from `not_null<U>` to `not_null<T>` when there exists an implicit conversion from `U` to `T`.

However it does not provide an easy way to convert from `not_null<U>` to `not_null<T>` when the conversion from `U` to `T` requires the use of `static_cast`.

Manually, one would have to do:

`auto converted = gsl::not_null<T>(static_cast<T>(original.get()))`

In accordance with the STL naming conventions, I propose the function `gsl::static_pointer_cast(gsl::not_null<U> const &)` to do the job. Until the GSL gets merged with the STL, it would lie on the `::gsl` namespace, instead of being on `::std` along with `static_pointer_cast(std::shared_ptr<U> const &)`.

Note 1: `noexcept` could be added if `not_null` constructor were `noexcept`.
Note 2: `dynamic_pointer_cast` doesn't have to be implemented since `dynamic_cast` can return `nullptr`.


samuelwtheck:
Re-thinking this for the case when we have smart pointers instead of raw pointers, such as `gsl::not_null<std::observer_ptr<T>>`, `gsl::not_null<std::unique_ptr<T>>`, `gsl::not_null<std::shared_ptr<T>>`.

PR #195, opened 2015-11-15
fixes pragma undef warnings in MSVC2013

lukhase:
use "#undef" instead of "#pragma undef"


neilmacintosh:
Thanks @lukhase, this looks good to me!

PR #199, opened 2015-11-18
Added explicit cast to size_type in span constructor

matus-chochlik:
This silences implicit sign conversion warnings when constructing span
from containers which return size_t from size().


neilmacintosh:
Thanks @matus-chochlik! Looks good to me.

PR #201, opened 2015-11-19
Made conversion of span<> to bool explicit

matus-chochlik:
This should prevent some accidental implicit conversions from span to integers.

### DIFF
--- a/include/span.h
+++ b/include/span.h
@@ -1434,7 +1434,7 @@ public:
         return m_pdata;
     }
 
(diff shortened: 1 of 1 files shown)


neilmacintosh:
Thanks @matus-chochlik, that conversion should never have been implicit to begin with. Just a warning (in case you have started to rely on the conversion) that I will be removing it altogether soon. Nonetheless, it's good to have the fix now.

PR #204, opened 2015-11-20
Fixed conversion problem when creating strided_span from span and bounds

annagrin:


neilmacintosh:
Thanks @annagrin, this looks good to me.

Issue #192, opened 2015-11-14
as_span(str) returns cstring_span regardless of constness of str


elronayellin:
because it is implemented with .data() which has no nonconst overload in std::string, unlike std::vector

neilmacintosh:
Well spotted @elronayellin. I agree that this should be corrected. Did you have a PR you'd like to offer?

mkurdej:
This was closed by #205, wasn't it?

JordanMaples:
Closing `as_span` no longer exists.

PR #205, opened 2015-11-20
as_span(str) returns cstring_span regardless of constness of str #192

elronayellin:


neilmacintosh:
Thanks @elronayellin! This looks good to me.

PR #207, opened 2015-11-21
Add CMake install target for header files

jasonhorsburgh:
Added install command for gsl header files.

I've set the install directory to include/gsl but happy to change it to something else.


neilmacintosh:
This looks good to me.

Issue #208, opened 2015-11-22
Clang "-Wshadow" warnings


rmpowell77:
When building with "-Wshadow" on clang-702.2.17.1:

```
GSL/span.h:510:43: error: declaration shadows a field of 'TypeListIndexer<TypeChain>' [-Werror,-Wshadow]
        TypeListIndexer(const TypeChain & obj) :obj(obj){}
                                          ^
GSL/span.h:509:27: note: previous declaration is here
        const TypeChain & obj;
                          ^
GSL/span.h:817:60: error: declaration shadows a field of 'bounds_iterator<IndexType>' [-Werror,-Wshadow]
    explicit bounds_iterator(const Bounds& bnd, value_type curr) noexcept
                                                           ^
GSL/span.h:1008:37: note: previous declaration is here
    std::remove_const_t<value_type> curr;
                                    ^
```

gdr-at-ms:
This idiom is so common in C++ that a compiler that warns on it, makes the warning less useful by hiding it a sea of useless warnings.

bdeeming:
The GSL is intended for extremely widespread use, and as such should support as many environments as possible. For example, we have this very warning enabled in our builds, and having a library that generated these warnings would be _highly_ annoying (and add just another hurdle to jump before we could adopt it).
This warning is very easy to fix, and has the potential to find real bugs, so I see no reason not to fix it and enable the warning flag in the main build.

gdr-at-ms:
@bdeeming 
You may have misunderstood what I meant.
"Shadowing warning" is a useful one to have one's build set.
A bad implementation of "shadowing warning" is that useful.  For example, not all compilers implementing "shadowing warning" implement this bad version of it -- which actually has a special rule, specifically designed in the language to accommodate/support the practice.  Yes, I agree with you this particular warning is annoying, but only because the compiler decides to have an annoying/buggy implementation of something that is otherwise useful.

Just like we want GSL to be widely adopted, we are also careful not trying too hard to workaround every single bug in compilers.

bdeeming:
@gdr-at-ms I do apologize, but I still do not understand what you mean. The example warning output shown above is exactly the output I would expect from the warning - it doesn't strike me as being the result of a bug in the compiler (here both the member variable `obj` and the function parameter `obj` are both available in the same scope, and therefore the function parameter over-shadows the member variable). We use GCC and it behaves in the same way as Clang does above. Are you able to point me to an example of a 'good' version of the warning?

[...]

PR #209, opened 2015-11-22
Fix Issue up shadowed variable.

rmpowell77:
Fixing up shadowed variable warning (issue #208) by changing constructor parameter name (adding trailing '_')

### DIFF
--- a/include/span.h
+++ b/include/span.h
@@ -507,7 +507,7 @@ namespace details
     struct TypeListIndexer
     {
         const TypeChain & obj;
-        TypeListIndexer(const TypeChain & obj) :obj(obj){}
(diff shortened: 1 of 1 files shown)


neilmacintosh:
I appreciate the change, but don't like the suffix of the parameters with `_`. That is more commonly seen as a convention for member variables. In fact, I would prefer to see our member variables follow that convention (I don't like the whole `m_` thing and it will eventually get replaced).

rmpowell77:
Changed based on @neilmacintosh comments.  Avoiding `_` and using human readable names. (obj -> object, curr -> current)

neilmacintosh:
This should be superseded by #216.

Issue #210, opened 2015-11-22
Clang "-Wsign-conversion" warnings


rmpowell77:
When building with clang-702.2.17.1 and "-Wsign-conversion" 

```
GSL/string_span.h:112:28: error: implicit conversion changes signedness: 'std::ptrdiff_t' (aka 'long') to 'size_t' (aka 'unsigned long') [-Werror,-Wsign-conversion]
    auto len = strnlen(sz, max);
               ~~~~~~~     ^~~
GSL/string_span.h:119:28: error: implicit conversion changes signedness: 'std::ptrdiff_t' (aka 'long') to 'size_t' (aka 'unsigned long') [-Werror,-Wsign-conversion]
    auto len = strnlen(sz, max);
               ~~~~~~~     ^~~
GSL/string_span.h:125:28: error: implicit conversion changes signedness: 'std::ptrdiff_t' (aka 'long') to 'size_t' (aka 'unsigned long') [-Werror,-Wsign-conversion]
    auto len = wcsnlen(sz, max);
               ~~~~~~~     ^~~
GSL/string_span.h:131:28: error: implicit conversion changes signedness: 'std::ptrdiff_t' (aka 'long') to 'size_t' (aka 'unsigned long') [-Werror,-Wsign-conversion]
    auto len = wcsnlen(sz, max);
               ~~~~~~~     ^~~
```

rmpowell77:
As of de62e7b0, there are still warnings:

```
GSL/tests/unittest-cpp/UnitTest++/Posix/TimeHelpers.cpp:30:15: warning: implicit conversion changes
      signedness: 'int' to 'useconds_t' (aka 'unsigned int') [-Wsign-conversion]
    usleep(ms * 1000);
    ~~~~~~ ~~~^~~~~~
1 warning generated.
```

rmpowell77:
Oh, those are in the unit tests.  Ah, I think then this issue is closed.

grahamreeds:
That is from UnitTest++ not GSL.

Could you log a bug over at https://github.com/unittest-cpp/unittest-cpp so
we can have a look at it, or submit a PR if you have one.

Ta,

GR

PR #211, opened 2015-11-22
Fixing up "-Wsign-conversion"  warnings.

rmpowell77:
Fix for issue #210.  Adding explicit casts to size_t.

### DIFF
--- a/include/string_span.h
+++ b/include/string_span.h
@@ -109,26 +109,26 @@ inline basic_string_span<T, dynamic_range> ensure_z(T* const & sz, std::ptrdiff_
 // TODO (neilmac) there is probably a better template-magic way to get the const and non-const overloads to share an implementation
 inline basic_string_span<char, dynamic_range> ensure_z(char* const& sz, std::ptrdiff_t max)
 {
-    auto len = strnlen(sz, max);
+    auto len = strnlen(sz, static_cast<size_t>(max));
     Ensures(sz[len] == 0);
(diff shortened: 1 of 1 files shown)


include/string_span.h
```diff
@@ -109,26 +109,26 @@ inline basic_string_span<T, dynamic_range> ensure_z(T* const & sz, std::ptrdiff_
 // TODO (neilmac) there is probably a better template-magic way to get the const and non-const overloads to share an implementation
 inline basic_string_span<char, dynamic_range> ensure_z(char* const& sz, std::ptrdiff_t max)
 {
-    auto len = strnlen(sz, max);
+    auto len = strnlen(sz, static_cast<size_t>(max));
     Ensures(sz[len] == 0);
```
gdr-at-ms: Is this a use case for `narrow_cast<size_t>`?
It is simpler to grep for and relate to `narrow<>`.

rmpowell77: That's a good point.  However, I'm not sure how I can easily use `narrow_cast`; gsl.h defines `narrow_cast`, but I think there's a circular header dependency so if I change to `narrow_cast`, it's trying to use it before it's declared.

Perhaps if `narrow_cast` lived in it's own header...?

neilmacintosh: That's a change that's on its way (to resolve that dependency issue and use narrow_cast where appropriate).
________________________________

neilmacintosh:
@rmpowell77 This should have now been fixed, so I am closing this PR. Thanks!

PR #218, opened 2015-12-02
Dev/annagrin/string span fixes

annagrin:
Rewrote string_span as a separate class, changed handling of 0-teminated strings to ensure we do not include a 0 in the span, unless it was included in the length of the original collection the span is created from.  Added tests.


neilmacintosh:
This looks good to me, thanks @annagrin!

PR #221, opened 2015-12-06
Correct missing rename of bytes() to size_bytes()

mcheese:
Corrected a single missing rename of bytes() to size_bytes().


neilmacintosh:
Thanks @mcheese, that looks great.

PR #223, opened 2015-12-08
Dev/annagrin/string span operator fixes

annagrin:
Fixed bugs:
- crash on creating string_span from an empty string
- creation of string_span from pointer and length keeps the given length
- made comparison operators work if one of the types is convertible to string_span


neilmacintosh:
Looks good to me, thanks @annagrin!

PR #225, opened 2015-12-11
VS2013: macro max, move ctors, pragma warning

menuet:
Hi,

I had some small issues when using GSL with VS2013. Maybe, you would like to integrate some of my fixes back in your code.
- First, the macro max defined by Windows.h conflicts with std::max, so I temporarily undefined max in span.h, so that it works for windows users even if NOMINMAX is not defined.
- Second, with VS2013, the move constructors cannot be defaulted so I had to implement them for string_span.
- Third, with VS2013, there is a pragma warning pop without any corresponding pragma warning push in string_span.h

Finally, I have some trouble compiling the unit tests for string_span with VS2013. There is an ICE because of the constructor of string_span taking a span. I am trying to work around it but I don't have a good workaroung yet.

Regards

### DIFF
--- a/include/span.h
+++ b/include/span.h
@@ -45,6 +45,12 @@
 #pragma push_macro("constexpr")
 #define constexpr
 
+// On Windows, if NOMINMAX is not defined, then windows.h defines the macro max
+#ifdef _WIN32
+#pragma push_macro("max")
+#define max max
+#endif
+
 // VS 2013 workarounds
 #if _MSC_VER <= 1800
 
@@ -2195,6 +2201,11 @@ general_span_iterator<Span> operator+(typename general_span_iterator<Span>::diff
 
 #undef constexpr
 #pragma pop_macro("constexpr")
+
+#ifdef _WIN32
+#undef max
+#pragma pop_macro("max")
+#endif
 
 #if _MSC_VER <= 1800
 #pragma warning(pop)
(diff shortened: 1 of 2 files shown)


neilmacintosh:
Hi @menuet! Are you sure you are using VS 2013 Update 5 to build your code that is using the GSL? 

For the second two issues: we have test suites for VS 2013 that build and pass.
For the first issue: I would prefer to rely on users using NOMINMAX and solving the max() macro problem for themselves externally.

menuet:
Sorry, I had VS2013 Update 4. So, after your reply, I just upgraded to Update 5 in order to double-check.
Then I re-built your master branch's unit-tests and I got the same warnings and errors.

1  warning:
  c:\dev\experiments\msft-gsl\include\string_span.h(859): warning C4193: #pragma warning(pop) : no matching '#pragma warning(push)'
2 errors:
  c:\dev\experiments\msft-gsl\include\string_span.h(238): error C2610: 'gsl::basic_string_span<CharT,Extent>::basic_string_span(gsl::basic_string_spa
n<CharT,Extent> &&)' : is not a special member function which can be defaulted
  c:\dev\experiments\msft-gsl\include\string_span.h(244): error C2610: 'gsl::basic_string_span<CharT,Extent> &gsl::basic_string_span<CharT,Extent>::o
perator =(gsl::basic_string_span<CharT,Extent> &&)' : is not a special member function which can be defaulted

About the macro max, I agree, it was just for convenience and it is unnecessary.

For the other issues, they may be real issues, don't you think so? (VS2013 does not generated automatic default move constructors, even in Update 5, right?)

PR #230, opened 2015-12-15
Fixed compilation issues in MSVC 2013

annagrin:
VS2013 compilation break fixes. Made sure that we don't compile the code that is supposed not to compile.


neilmacintosh:
Nice tidy-up work @annagrin, thanks!
Could you follow up by changing the static_cast's in there to narrow_cast when you get a chance (because they are narrowing from signed to unsigned values).

PR #233, opened 2015-12-22
Fix missing #include <string> in string_span.h

tcbrindle:
`string_span.h` uses `std::string`, but `<string>` is not included.


neilmacintosh:
Thanks @tcbrindle! This looks good to me.

PR #246, opened 2016-01-15
Enable indexing by enum when index rank is 1

mkurdej:


neilmacintosh:
Hello @mkurdej! Thanks for the PR, but it would represent a design change that we are not comfortable with right now. As far as I can tell, this does not enable any functionality, all it does is save typing `static_cast` in the case you would like to index by an enum value. I'm not convinced that hiding that `static_cast` in the source code is necessarily a good thing. Note that neither of the standard library containers `std::basic_string` or `std::vector` provide similar functionality although both support indexing operations.

If you feel strongly about a particular scenario you are trying to support, I would encourage you to open an issue that provides some context for what you are looking to achieve. For now, I'll close out this PR.

PR #250, opened 2016-01-19
Fix #245 division by zero in extent

### LINKED ISSUE #245: Division by zero for empty multidimensional dynamic spans
The calculation of extents (and possibly other things) of multidimensional spans with dynamic bounds results in division by zero [in this line](https://github.com/Microsoft/GSL/blob/62b063a27258f8d6344a4ddd49ddd8739f71eb49/include/span.h#L397).

Take for example a 2D view onto a vector:

```
std::vector<int> vec = {};
size_t w = 3; // Arbitrary
size_t h = vec.size() / w;
// Create a (w x h) 2D view
auto view = as_span( as_span( vec ), dim<>( w ), dim<>( h ) );

view.extent<0>(); // Expected to return 0 but crashes.
```

Is this an oversight or not supported?

mkurdej:


JordanMaples:
Maintainer's call: Sorry about the lack of reply here, however it looks like this function no longer exists. Closing.

PR #259, opened 2016-02-04
Added zstring_span and removed zstring_builder to support legacy strings

annagrin:
Removed inconvenient zstring_builder class and added zstring_span wrapper type to convers zero terminated spans into legacy zero terminated strings. Added tests (with one usage case)

### DIFF
--- a/include/string_span.h
+++ b/include/string_span.h
@@ -527,7 +527,7 @@ inline std::wstring to_string(wstring_span<> view)
 
 // zero-terminated string span, used to convert
 // zero-terminated spans to legacy strings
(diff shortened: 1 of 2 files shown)


neilmacintosh:
@annagrin, can you please:

1. take a look at the Travis CI build failures?
2. add =default for copy/assignment/destructor so it's clear we intend them to be so.
3. change the `size_t` for the Extent template parameter to `std::ptrdiff_t` to match the other `span<>` family types.

neilmacintosh:
Terrific, thanks @annagrin nice work!

Issue #222, opened 2015-12-07
narrow should throw when casting negative to unsigned


brunom:
int i = -42;
    unsigned u = gsl::narrow<unsigned>(i);

runs when it should fail with an exception because -42 is not representable as an unsigned.

AndrewPardoe:
Editors discussed 14 Dec 2015

neilmacintosh:
Agreed @brunom! The initial implementation of `narrow()` (and `narrow_cast()`) was pretty bare-bones. Feel free to offer a PR if you are so inclined. :-)

PR #262, opened 2016-02-08
narrow: check for changed sign

chfast:
Fixes https://github.com/Microsoft/GSL/issues/222.

### DIFF
--- a/include/gsl_util.h
+++ b/include/gsl_util.h
@@ -22,6 +22,7 @@
 #include "gsl_assert.h"  // Ensures/Expects
 #include <array>
 #include <utility>
+#include <type_traits>
 #include <exception>
 
 #ifdef _MSC_VER
@@ -90,10 +91,24 @@ inline constexpr T narrow_cast(U u) noexcept
 
 struct narrowing_error : public std::exception {};
 
+template<class T, class U>
+struct is_same_signess
+{
+    static const bool value =
+        std::is_signed<T>::value == std::is_signed<U>::value;
+};
+
 // narrow() : a checked version of narrow_cast() that throws if the cast changed the value
 template<class T, class U>
 inline T narrow(U u)
-{ T t = narrow_cast<T>(u); if (static_cast<U>(t) != u || (t < T{}) != (u < U{})) throw narrowing_error(); return t; }
+{
+    T t = narrow_cast<T>(u);
+    if (static_cast<U>(t) != u)
+        throw narrowing_error();
+    if (!is_same_signess<T, U>::value && ((t < T{}) != (u < U{})))
+        throw narrowing_error();
+    return t;
+}
 
 //
 // at() - Bounds-checked way of accessing static arrays, std::array, std::vector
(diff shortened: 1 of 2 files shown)


neilmacintosh:
Thanks for the PR @chfast! There is one refinement I wanted to add before we implemented a change like this. We should only do the newly added test if there is the possibility of changing sign in the conversion. This helps us remain efficient in cases where you are narrowing between two unsigned integer values of different sizes.

As an illustration of the idea...something like this:
````
template 
constexpr bool IsSignedToUnsigned() { /* suitable template magic */ }
template< class T, class U >
T narrow( U u ) 
{ 
T t = narrow_cast( u ); 
if (static_cast<U>(t)!=u) throw narrowing_error(); 
if (IsSignedToUnsigned<U, T>() && (t > 0)!=(u > 0))) throw narrowing_error();
return t; 
}
````

Would you be interested in revising the PR to do something along these lines?

chfast:
Ok, I get your point. However, I believe a compiler is able to optimize that out without our help in case of two unsigned types as expression `x < 0` is always false in such case. Maybe the hint will be needed in case of two signed types. Anyway, I will check that.

I was also wandering if you and other contributor has thoughts about what semantics of `narrow_cast` and derivatives should be. Questions like:
1. Should it accept only primitive types? 
2. Should it work for `bool`?
3. Should it work for floating point types?
3. What about user types? We can consider user's implicit/explicit conversions here...

My change implies that the types have operator `<` defined.

chfast:
I've updated the PR according to suggestions. However, MSVC produces "conditional expression is constant" warning in some cases.

chfast:
@neilmacintosh?

neilmacintosh:
Thanks @chfast!

A few small things:

1. `is_same_signess` might be better spelled `are_both_signed` or `is_same_signedness`.
2. a helper function like `is_same_signedness` would usually go in a `details` namespace within gsl, to keep it out-of-the-way. You can see other examples of this in `span.h`.
3. The `value` static variable in `is_same_signess` should be `constexpr` I believe. You could just derive from `std::integral_constant` in this case, which declares everything for you and provides exactly the behavior you want.
4. If MSVC issues the warning, then it can be suppressed with an accompanying comment explaining it is by-design in this case. Again, `span.h` has some existing examples of doing this.

chfast:
Thanks for the comments @neilmacintosh.

Unfortunately, it looks like the warning cannot be disabled easily as they are raised in the place of template instantiation and pragmas around the template definition make no difference.

This can be solved using template specialization but it will be uglier in my opinion.

neilmacintosh:
@chfast, looks great, thanks very much for the PR!

No need to worry about the warnings that will be issued at the point of template instantiation.

hsutter:
Editors call: We agree and are updating the specification of `narrow` to cover signedness promotions.

hsutter:
Editors call: Also, note that the Microsoft GSL uses `!=` and the GSL spec uses "not `==`" and we think that's fine, operators should have their natural meanings.

PR #265, opened 2016-02-12
Follow standard-rules for macros

Florianjw:
According to the standard, it is undefined behavior to re-`#define` keywords. The first part of this pull request fixes that.

[According to the Core-Guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Res-ALL_CAPS), macros should always be in `ALL_CAPS`. Since `Expects` and `Ensures` are macros, this commit renames them accordingly:

* By default, their new names are `GSL_EXPECTS` and `GSL_ENSURES`, thereby also not polluting the global macro-namespace.
* if `GSL_USE_UNPREFIXED_CONTRACTS` is defined, `EXPECTS` and `ENSURES` will defined as aliases
* if `GSL_USE_UNPREFIXED_AND_LOWERCASE_CONTRACTS` is defined, `Expects` and `Ensures` will be defined as aliases
* The two-alias-options are independent of each other.

I've compiled and tested the code under x86_64 Debian GNU/Linux with GCC 5.3 and Clang 3.6 and could not observe problems. Since most of the changes of the first part will however only affect VS, it would be very wise to test that as well.

### DIFF
--- a/include/gsl.h
+++ b/include/gsl.h
@@ -19,7 +19,7 @@
 #ifndef GSL_GSL_H
 #define GSL_GSL_H
 
-#include "gsl_assert.h"  // Ensures/Expects
+#include "gsl_assert.h"  // GSL_ENSURES/GSL_EXPECTS
 #include "gsl_util.h"    // finally()/narrow()/narrow_cast()...
 #include "span.h"           // span, strided_span...
 #include "string_span.h"    // zstring, string_span, zstring_builder...
@@ -120,7 +120,7 @@ private:
 
     // we assume that the compiler can hoist/prove away most of the checks inlined from this function
     // if not, we could make them optional via conditional compilation
-    void ensure_invariant() const { Expects(ptr_ != nullptr); }
+    void ensure_invariant() const { GSL_EXPECTS(ptr_ != nullptr); }
 
     // unwanted operators...pointers only point to single objects!
     // TODO ensure all arithmetic ops on this type are unavailable
--- a/include/gsl.h
+++ b/include/gsl.h
@@ -25,17 +25,20 @@
 #include "string_span.h"    // zstring, string_span, zstring_builder...
 #include <memory>
 
+#define GSL_IMPL_CONSTEXPR constexpr
+#define GSL_IMPL_NOEXCEPT noexcept
+
 #ifdef _MSC_VER
 
 // No MSVC does constexpr fully yet
-#pragma push_macro("constexpr")
-#define constexpr 
+#undef GSL_IMPL_CONSTEXPR
+#define GSL_IMPL_CONSTEXPR
 
 // MSVC 2013 workarounds
 #if _MSC_VER <= 1800
-// noexcept is not understood 
-#pragma push_macro("noexcept")
-#define noexcept  
+// noexcept is not understood
+#undef GSL_IMPL_NOEXCEPT
+#define GSL_IMPL_NOEXCEPT
 
 // turn off some misguided warnings
 #pragma warning(push)
@@ -149,20 +152,12 @@ namespace std
 
 } // namespace std
 
-#ifdef _MSC_VER
 
-#undef constexpr
-#pragma pop_macro("constexpr")
+#undef GSL_IMPL_CONSTEXPR
+#undef GSL_IMPL_NOEXCEPT
 
-#if _MSC_VER <= 1800
-
-#undef noexcept
-#pragma pop_macro("noexcept")
- 
+#if defined(_MSC_VER) and (_MSC_VER <= 1800)
 #pragma warning(pop)
-
 #endif // _MSC_VER <= 1800
-
-#endif // _MSC_VER
 
 #endif // GSL_GSL_H
--- a/include/gsl_assert.h
+++ b/include/gsl_assert.h
@@ -54,25 +54,38 @@ struct fail_fast : public std::runtime_error
 };
 }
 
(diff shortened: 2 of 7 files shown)


Florianjw:
Considering the errors, the Travis-build-failure look like the fault of Travis to me. Could someone please look into it and confirm that assumption?

gdr-at-ms:
@Florianjw 
Hmm, which part of the standard supports that?

Keywords like 'constexpr' are just plain identifiers from translation phases 1 through 4.  Preprocessor tokens are converted to tokens (hence keywords) only at translation phase 7.  See paragapgh 2.2/1.

Florianjw:
17.6.4.3.1, paragraph 2 [macro.names]:

> A translation unit shall not #define or #undef names lexically identical to keywords, to the identifiers listed in Table 3, or to the attribute-tokens described in 7.6.


Granted, it is only in the library requirements and you are currently restoring the previous state before leaving the header. I may be missing something, but even so I believe that my version is better style.

chfast:
I think `Expects` and others are exceptions to the rule, explicitly pointed by Guidelines: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#gslassert-assertions.

Florianjw:
> These assertions is currently macros (yuck!) **pending standard commission decisions on** contracts and assertion **syntax**.

I'd prefer to be carefull here. Especially for the case where those identifiers become keywords.

Furthermore this library may be considered as *the* example for how to write C++, so I would prefer, if there are no parts in it, that some people might use as an argument for not having to follow the rules “because the GSL does it also”.

gdr-at-ms:
@Florianjw 
The *Library* paragraph that you are quoting is for a different reason than the one your originally pointed out.

The language rules are clear: keywords don't exist during preprocessing (translation phases 1-4).

The library clause is there to allow C++ implementations to provide _standard headers_ as header files without having to say anything convoluted about what happens when you define a keyword right before including a standard header.

The practice in the GSL is extremely careful: in the `constexpr` case that you're citing, it targets a very specific implementation which has a very specific push/pop mechanism in place to avoid running afool of the issues that the library clause you ae quoting is trying to prevent.

I believe it is OK is it currently is.

Florianjw:
> The practice in the GSL is extremely careful: in the constexpr case that you're citing, it targets a very specific implementation which has a very specific push/pop mechanism in place to avoid running afool of the issues that the library clause you ae quoting is trying to prevent.

Let's ignore ghost drivers like Clang that believe they are entitled to define the identification macros of other compilers. I believe that defending that horrible practice is not something that should be protected.

Nonetheless I believe that the new code is cleaner (and it is definitely shorter) and by that makes reading it and convincing yourself about it's correctness simpler. Which is IMHO sufficient reason to change it.

And then there is the “bad role model” for this thing as well: Even though it is probably correct here, someone will at some point get it wrong and if you try to talk people out of it,  I wouldn't want them to be able to point at the GSL and say “but they do it also”. Sadly “legally correct” != good

gdr-at-ms:
@Florianjw 
I think you are overdramatizing to make a point, that is unfortunately getting in the way of the communication: I am struggling to take away the real essence.

Florianjw:
I assure you, that this is not my intention. 

The essence in one short sentence is that I believe that the new code is cleaner.

neilmacintosh:
@Florianjw thanks for the PR. However, as @gdr-at-ms points out, there is no standards-violation here, so I'll leave the code as-is for now.

FrankHB:
The conclusion is confusing. Such use of preprocessing directives is clearly undefined in every C++ translation unit including standard headers as per the standard, literally.

Note the lexical forms of keywords (i.e. their spelling) can exist whether the TUs are translated to phase 7 or not. Otherwise, [lex.pptoken]/1 cannot work at all.

Undefined behavior in particular source files can be tolerated once you have supplied the additional guarantees to make the whole program's behavior predictable again, e.g. in an implementation of standard library (which **has to** be implementation-specific). But this is questionable to this project. In practice, this may require some additional documentation, as well as something like `#pragma GCC system_header` in specific header files.

It might be OK since you can guarantee such part of code really "targets a very specific implementation", but it is still preferable to mark it explicitly to indicate it is relying on some assumptions out of the standard besides `#ifdef`, and to warn users not to treat such code guidelines-conforming.

MikeGitb:
Imho the definition of constexpr and noexcept is a bit annoying but ok, because it leads to cleaner code and is a temporary measure restricted to the headers of the GSL and a specific compiler for which I assume the implementers have checked that it is ok to do this. 
The problem I have with it is that there is still no separate documentation and if a user just looks at the source code of a particular function he might think that this function is constexpr or noexcept when it actually isn't and get unexpected compiler errors.

What I find really problematic though is the naming of the `Ensures` and `Expects` macros. Bad enough that they are macros in the first place, but I really don't know why someone thought it were a good Idea to violate existing best practices as well as the coreguidelines when naming them without any real need. And no it is not an implementation detail, because contrary to the `constepxr/noexcept`-*fiddling* those macros do leak - by default - into anyone's code that uses the gsl - and into anyone's code that uses a library that uses the gsl and anyone's code that ... (you get the picture). Haven't we learned anything since the day's when someone thought it was a good idea to define things like `min`, `max` and `ERROR` in `windows.h`? 

Sorry for the little rant. I hope you do know that I value and appreciate the work work you are doing here. And I do believe that a usable 90% solution is better than an unavailable 100% solution.

PR #268, opened 2016-02-16
Fix: Issue: #267 gsl::fail_fast only defined for one configuration

### LINKED ISSUE #267: gsl_assert.h only defines gsl::fail_fast exception with GSL_THROW_ON_CONTRACT_VIOLATION
Header file `gsl_assert.h` only defines `gsl::fail_fast` exception with `GSL_THROW_ON_CONTRACT_VIOLATION` defined.

This is a problem because client code needs to be able to compile different library configurations without having to add #ifdef's in its catch blocks.

The macros `Expects` and `Ensures` are defined for all three configurations (even when they do nothing) so it makes sense that `gsl::fail_fast` should be defined for all three as well.

galik:
PR to fix #267

### DIFF
--- a/include/gsl_assert.h
+++ b/include/gsl_assert.h
@@ -20,6 +20,7 @@
 #define GSL_CONTRACTS_H
 
 #include <exception>
+#include <stdexcept>
 
 //
 // There are three configuration options for this GSL implementation's behavior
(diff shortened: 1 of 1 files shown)


neilmacintosh:
LGTM. Thanks for the PR @galik!

Issue #47, opened 2015-09-24
Use a specific version of unittest-cpp


kylog:
Currently the README says to clone https://github.com/Microsoft/unittest-cpp, which means a given SHA of GSL may produce different test results on different days if/as unittest-cpp gets new commits.

Two related questions then:
* can we use a git submodule to pin to a particular SHA of unittest-cpp? Or some other mechanism?
* is Microsoft going to maintain its fork of unittest-cpp or is it upstreaming its changes?

neilmacintosh:
We are upstreaming the changes currently.

Sent from my Windows Phone
________________________________

PR #261, opened 2016-02-08
unittest-cpp as a submodule

chfast:

PR #276, opened 2016-03-24
Making unittest-cpp a submodule.

sergiyor:
This should address the following issue: https://github.com/Microsoft/GSL/issues/47
Note that I invoke git within cmake to automatically initialize the submodule. This assumes that git is in the path.


neilmacintosh:
Thanks @sergiyor! A nice improvement.

chfast:
I feel sad. https://github.com/Microsoft/GSL/pull/261.

PR #279, opened 2016-03-29
[span] Fix UB hack in static_bounds::operator=(const static_bounds&)

CaseyCarter:
* Make `BoundsRanges<dynamic_range, ...>::m_bound` non-`const` and private
* Remove various defaulted copy constructor/assignment declarations from `BoundsRanges` specializations whose only effect is to needlessly suppress the generation of moves
* Remove the hackish `static_bounds::operator=(const static_bounds&)`. The implicitly generated default is now sufficient.

Technically, the `new` expression itself:
```c++
new (&m_ranges) MyRanges(otherBounds.m_ranges); 
```
has defined behavior. The problem is that any access to the new object created atop `m_ranges` through the name of or pointers to the object that previously existed there does result in UB per [basic.life]/7.3.

### DIFF
--- a/gsl/multi_span
+++ b/gsl/multi_span
@@ -308,8 +308,6 @@ namespace details
         {
         }
 
-        BoundsRanges(const BoundsRanges&) = default;
-        BoundsRanges& operator=(const BoundsRanges&) = default;
         BoundsRanges(const std::ptrdiff_t* const) {}
         BoundsRanges() = default;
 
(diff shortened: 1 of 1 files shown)


neilmacintosh:
@CaseyCarter I'm afraid the source has moved around a little underneath you, and this PR applies to `multi_span`, now that `span` has moved off the original multidimensional implementation. Do you want to move it to apply to `multi_span`? If not, then I can pick it up and apply it manually (just wanted you to get the credit if you would like it for the fix).

CaseyCarter:
Sorry for the slow turnaround, there 'tis.

neilmacintosh:
Thanks @CaseyCarter!

PR #280, opened 2016-03-29
[span] Conform to the Ranges TS View concept.

CaseyCarter:
* Allow default construction of spans with fixed-non-zero extent.
* Allow default construction of `contiguous_span_iterator` and `general_span_iterator`.
* `contiguous_span_iterator` members `Expect` a non-default-initialized object as appropriate.

This change makes `span` conform to the `View` concept of the Ranges TS, and its iterators conform to `RandomAccessIterator` as specified in both the Ranges TS and C++14. (Note that the absence of default construction makes the current iterators not conform to C++14's `ForwardIterator` per [forward.iterators]/1.2.)

Notably, the resulting `span` and its iterators integrate well with [range-v3](http://github.com/ericniebler/range-v3)

### DIFF
--- a/include/span.h
+++ b/include/span.h
@@ -1218,10 +1218,6 @@ public:
     // default constructor - same as constructing from nullptr_t
     constexpr span() noexcept : span(nullptr, bounds_type{})
     {
-        static_assert(bounds_type::dynamic_rank != 0 ||
-                          (bounds_type::dynamic_rank == 0 && bounds_type::static_size == 0),
-                      "Default construction of span<T> only possible "
-                      "for dynamic or fixed, zero-length spans.");
     }
 
     // construct from nullptr - get an empty span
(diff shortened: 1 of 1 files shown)


neilmacintosh:
@CaseyCarter thanks for the PR! However, fixed-size, non-zero-extent `span` objects are not default-constructible by-design, so I can't take this change. Your other changes are still useful, but now apply to `multi_span`, as `span` has been simplified to be single-dimension only and uses simpler iterators. If you can be bothered, I'd be happy to accept an updated PR to apply those changes to the `multi_span` iterators.

neilmacintosh:
@CaseyCarter I'm going to close this PR out now as I believe that the Ranges TS may be changing the `DefaultConstructible` requirement on `View` anyway, so `span` will be a conforming `View` at that point. If that turns out wrong, feel free to reopen ;-)

ericniebler:
> as I believe that the Ranges TS may be changing the `DefaultConstructible` requirement on `View` anyway

I wonder what made you think that.

CaseyCarter:
> I wonder what made you think that.

I gave him that notion when I filed ericniebler/stl2#179 after a discussion with Neil, GDR, Bjarne, Herb, and Andrew Pardoe about removing the `DefaultConstructible` requirement for `View` and `Iterator` (but probably not `ForwardIterator`). There's strong resistance to requiring default construction where it is not possible without relaxing type invariants.

PR #281, opened 2016-03-31
Suppressed CppCoreCheck warnings.

neilmacintosh:
This will prevent warnings about the GSL implementation from needlessly annoying users of CppCoreCheck. A more targeted set of suppressions will go into a future revision of the GSL.


LMLB:
You forgot to add `#pragma warning(pop)` at the end of `span.h`.

Issue #284, opened 2016-04-04
using finally without assigning to variable will not work as intended


juandent:
Hi,

If you use finally like so:

void usingFinally() {
     finally(&w);   // where w is a function
     // ... (the final_act object is already gone from here)

   /// ....
}

will construct final_act object but will immediately destroy it (will not wait until end of scope).

This FORCES the assignment to a variable like so:

void usingFinally() {
     auto x = finally(&w);   // where w is a function
     // ...

   /// ....
}

Kind of ugly and need a name for the variable that does not conflict with existing...

Is there a way around this?

Juan

capsocrates:
An ugly way around it that I have considered is a macro that assigns the
result of finally to a uniquely named variable.

#define SCOPE_GUARD(f) auto x = finally(f);

I don't know what kind of macro magic you might use to make 'x' always a
unique name.

chfast:
You can use `__LINE__`, eg. `scope_guard_ ## __LINE__`.

chfast:
The C++17 will have `[[nodiscard]]` attribute. The warning will be raised if an result of an function is imminently discarded by a caller.

Clang and GCC support some version of the attribute already. MSVC has `_Check_return_` annotation.

neilmacintosh:
Thanks for the suggestion @chfast. However, in terms of a solution, I'd prefer to use the C++17 attribute once it becomes available.

PR #285, opened 2016-04-08
Mark finally with nodiscard attribute.

chfast:
The compiler should warn about unsued result of finally().
See https://github.com/Microsoft/GSL/issues/284.

### DIFF
--- a/include/gsl_util.h
+++ b/include/gsl_util.h
@@ -78,13 +78,20 @@ private:
     bool invoke_;
 };
 
+// C++17 [[nodiscard]] attribute for legacy compilers.
+#if _MSC_VER
+#define GSL_NODISCARD _Check_return_
+#else
+#define GSL_NODISCARD __attribute__((warn_unused_result))
+#endif
+
 // finally() - convenience function to generate a final_act
 template <class F>
-inline final_act<F> finally(const F &f)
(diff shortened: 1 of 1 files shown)


chfast:
Ping!

neilmacintosh:
I'd rather avoid adding another macro into the GSL codebase (especially one that resolves to two nonstandard extensions - one of which is also a macro). It would be preferable to use the real C++17 attribute and wait for compiler support to arrive.

jpcima:
To be honest the issue which is described here troubles me as well.
I love the idea of scope guards in C++ and I use them extensively, and it happened that a tiny mistake like omitting to assign the object has caused me some headaches.

Usually I would save myself the risk of error by wrapping the assignment behind a macro expansion, but I understand GSL would not want that.

Maybe this proposal below could be accepted as compliant with the future C++17 standard? until the compiler support arrives.
```
#if defined(__has_cpp_attribute) && __has_cpp_attribute(nodiscard)
[[nodiscard]]
#endif
```

The attribute works under clang 3.9.0, and is planned for version 7 of GCC according to their status page.

gdr-at-ms:
Not arguing for or against the PR.  Just noting it is amusing that it takes far more CPP line noise to write the tokens that have any real observational impact, i.e. `[[nodiscard]]`.

PR #287, opened 2016-04-22
Add an overload for as_span to construct from two iterators.

LYP951018:


LYP951018:
Orz... Is this a bug of GCC?
>/home/travis/build/Microsoft/GSL/tests/../include/span.h: In instantiation of ‘class gsl::contiguous_span_iterator<int>’:
/home/travis/build/Microsoft/GSL/tests/span_tests.cpp:698:36:   recursively required by substitution of ‘template<class Span> constexpr Span gsl::as_span(gsl::contiguous_span_iterator<Span>, gsl::contiguous_span_iterator<Span>) [with Span = int]’

neilmacintosh:
Hello @LYP951018 and thanks for the PR. The layout of the source code has changed slightly since you put together this PR...as `span` lost its multidimensional support and the 'as_span' function now only applies to the multidimensional `multi_span' type. I'm going to close this out for now, but if you have an important use-case for `as_span`, please let me know and we can consider re-adding it to the library.

PR #289, opened 2016-05-09
gsl::at overload for initializer_list

ericLemanissier:
initializer_list do not have subscript operator, so the generic container overload of gsl::at fails to compile.
This commits adds an overload of gsl::at for initializer_lists, using *(initializer_list::begin()+index) instead of subscript operator, and associated test


neilmacintosh:
Thanks for the PR @ericLemanissier. This nicely fills a gap in the existing implementation and the code looks good to me.

PR #297, opened 2016-06-08
Fixed inconsistent indentation by replacing tabs with spaces.

robindegen:
Indentation was inconsistent, which was very apparent when viewing from within github.


robindegen:
Build seems to have failed due to ```gpg: no valid OpenPGP data found.```. Infra problem?

neilmacintosh:
I've been running clang-format over the headers recently before commit, and I took a quick look and this appears to be fixed now. If it doesn't look fixed, this PR has conflicts now anyway....so you could either update it and reopen or creat a new one.

Issue #298, opened 2016-06-11
Differentiating exception type for Expects and Ensures for option [GSL_THROW_ON_CONTRACT_VIOLATION] 


vjacquet:
Both Expects and Ensures throws runtime_error based exception. Shouldn't Expects throw logic_error based exception, defined by the standard to report violations of logical preconditions or class invariants?

neilmacintosh:
Thanks for the feedback @vjacquet! I agree, it should. Feel free to offer a PR that makes this fix.

PR #302, opened 2016-06-26
Differentiating exception type for Expects and Ensures for option [GSL_THROW_ON_CONTRACT_VIOLATION]

vjacquet:
Expects now throws std::logic_error when Ensures throws std::runtime_error (issue #298).
Uses multiple inheritance so that the exception thrown is both the standard and the gsl::fail_fast to limit the risks of breaking existing code.

### DIFF
--- a/include/gsl_assert.h
+++ b/include/gsl_assert.h
@@ -22,6 +22,18 @@
 #include <exception>
 #include <stdexcept>
 
+#ifdef _MSC_VER
+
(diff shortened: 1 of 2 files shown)


neilmacintosh:
Thanks @vjacquet but I think that both contract tests should throw the same exception, and it should derive from `std::logic_error`. This PR has now been obsoleted by #480.

PR #303, opened 2016-07-02
constexpr operators + inconsistent indentation style fix

ravstrix:


ravstrix:
Hm, interesting problem. AppVeyour build failed, but on my compiler (VS 2015) this code works fine.

PR #310, opened 2016-08-02
Fixed #308 - a mismatched pragma.

### LINKED ISSUE #308: Misplaced #pragma warning(pop) in span.h
If `GSL_THROW_ON_CONTRACT_VIOLATION` is not defined then the `#pragma warning(pop)` is skipped (it looks like it was accidentally put inside the wrong `#ifdef _MSC_VER` block), causing most of the CppCoreCheck warnings to stay disabled.

neilmacintosh:


neilmacintosh:
I hear no objections and it's a simple fix, so it's going in.

Issue #260, opened 2016-02-07
I really dislike the angle-brackets in `dim<>(some_number)`


Som1Lse:
Would it make sense to make `dim` a function like the following? (where `Dim` is the exact same type as the current `dim`, just renamed)
```c++
template <std::ptrdiff_t N>
constexpr Dim<N> dim() noexcept {
    static_assert(N >= 0,"something something less than zero");
    return Dim<N>();
}

template <std::ptrdiff_t N = dynamic_range>
constexpr Dim<N> dim(std::ptrdiff_t n) noexcept {
    static_assert(N == dynamic_range,"something something dynamic_range");
    return Dim<>(n);
}

void f(){
    dim<dynamic_range>(4);//still compiles
    dim<>(4);//still compiles
    dim(4);//now compiles
    dim<4>();//now compiles

    dim<4>(4);//does not compile
    dim();//does not compile
    dim<dynamic_range>();//does not compile
}
```
Of course `Dim` does not have to be the name of the type, it is backwards compatible with most user-code. The exception to that is if someone wrote their own `as_span`-like function which takes `dim`s or for some reason decides to create variables of the type (without `auto`), which I would expect to be reasonably rare, and easy to mitigate since these uses will almost certainly cause a compiler error, and the solution is a matter of replacing a few symbols with slightly different ones.

neilmacintosh:
Hello @Som1Lse, this suggestion looks pretty reasonable to me. Maybe dim_t rather than Dim as the name of the type? I find differing types vs. functions just on case hard to follow (and even harder to discuss verbally ;-)). Feel free to put together a PR with the change you're suggesting...

neilmacintosh:
I believe this issue has been resolved with PR #311.

PR #311, opened 2016-08-03
Implemented https://github.com/Microsoft/GSL/issues/260

Som1Lse:
All tests pass with GCC 5.3 on Windows. I have not tested other compilers.
I modified all tests to use the new syntax. I have not added any new tests to ensure that the old style still compiles but I did check that it does.


neilmacintosh:
Thanks @Som1Lse, this looks good to me!

Issue #277, opened 2016-03-25
System `#include` namespace for GSL?


galik:
I am going to suggest renaming the headers folder from `include` to, well just about anything else but something like `ccgsl` and here is why.

I think the library would benefit from choosing a, *include* namespace so developers can start including `GSL` modules in a manner that is less likely to conflict with other projects. Simply putting the headers in `include` increases the likelihood of future clashes when compiling against a *system installed* version of `GSL` eg:

```cpp
#include <span.h> // could clash with a million other libraries
```

Often a library will put their headers in a project sub-folder of the system `include`, for example:

```cpp
#include <SFML/Audio.hpp>
```

But even when the `GSL` is not installed system-wide and in its developmental form it would be good to still include its headers using such a standard include format by simply setting the *include path* for the compiler. So if I store the development version f `GSL` here:

```cpp
/home/galik/dev/MS-GSL
```

Well currently my choices are:

**A)** Set compiler to `-I$(HOME)/dev/MS-GSL/include` and do `#include <span.h>` ...

**B)** Set compiler to `-I$(HOME)/dev/MS-GSL` and do `#include <include/span.h>` ...

Neither of those are really satisfactory for me. So I do some sym-link shenanigans and fudge everything to look like a local copy where I just do:

```cpp
#include "gsl/span.h"
```

This is why I would suggest to change the name of the folder that the `GSL` lives in from `include` to `gsl`.

**EXCEPT!!!**

As was pointed out a while ago, there is a name-clash with another major project that goes by the name "GSL" [GNU Scientific Library](https://www.gnu.org/software/gsl/).

Those of us who have that installed as a system library are already doing this:

```cpp
#include <gsl/...>
```

So putting *this* `GSL` the same place as the *other* `GSL` is a bit problematic.

So I would suggest something a little more unique like `ccg-gsl` or `ccgsl` for "C++ Core Guideline-Guideline Support Library" and "C++ Core Guideline Support Library" respectively (or something else IDK).

Then I can do:

```cpp
#include <ccgsl/span.h> // and be a happy bunny
```

Does this make sense?

neilmacintosh:
Hi @galik! I like this idea and it makes a lot of sense. As a first step in this direction I've merged #207 which means there is now an "install" target in the CMake files which will install to a gsl subdirectory.

I'd be happy to look at a pull request if you've got the time to put one together.

[...]

PR #317, opened 2016-08-09
Renamed include/ folder to gsl/ to address issue #277

galik:
Renamed include/ folder to gsl/ to make including the library consistent whether using it from the development folder, from the installation folder or from being copied into a project. 

    #include <gsl/gsl.h>

Updated headers/tests/instructions/cmake build accordingly

This PR should address https://github.com/Microsoft/GSL/issues/277 (less the renaming of gsl itself)


neilmacintosh:
As a follow-up item, it would be great to drop the .h from the files. Let me know if you'd like to offer a PR to do that as well. Thanks for taking care of this part, though.

neilmacintosh:
Looks good to me.

galik:
@neilmacintosh I can remove the `.h` if you like. It's a little unusual to have no extension for non-standard headers though. I'm not sure how it might effect everyone's favorite `IDE`. When I tried it in my `eclipse` configuration the header disappeared from the project view. Though I can probably fix that with a little tweaking (not sure what happens by default). But if you're definitely sure I'll be happy to make a PR.

gdr-at-ms:
I would be in favor of getting rid of the '.h'.

neilmacintosh:
@galik If you could put together that PR, it would be great, thanks!

MikeGitb:
It's probably too late to revert this decision now, but could you at least shed some light on the rational behind dropping the .h extensions?

Personally, I see no benefit, but two drawbacks:
- First, the only extensionless header names one usually comes across are part of the standard library, so this breaks consistency and might confuse some people (admittedly rather unlikely) 
- Second, many editors (including visual studio) won't automatically recognize them as c++ header files which disables things like syntax highlighting and code completion.

neilmacintosh:
@MikeGitb strictly speaking, the `.h` extension is not required, it is just a combination of tradition and convention. The standard library eschews it, and the GSL aims to be a commonly used library just like the standard library, so there is a consistency argument there. Moreover, a number of GSL types (and their headers in the case of `span`) are currently proposed for standardization and - once accepted and published in a standard - would be without the `.h`. So, again, there is a consistency argument.

I understand some editors (including the one published by my company) will no longer syntax highlight and code complete the files. This is a small hurdle to overcome for people who are editing the files and should not affect the vast majority of users who should simply be `#include`ing the headers and using their contents, rather than editing them ;-)

MikeGitb:
@neilmacintosh: Thanks for the explanation, but it doesn't really convince me (not that you are in any obligation to do so of course). I'm aware, that the drawbacks I mentioned are minor points, but the question remains: What advantages does it give the user or the developer, that would outweight even those minor annoyances.

Also, is the gsl actually in a state, where most people just use it without having a look at the code? Especially, as there seems to be no official documentation or specification for the library as a whole, yet, I'd imagine that quite a lot of users will now and then have to have a look at the source files, even if they don't intent to modify them.

gdr-at-ms:
The safer buffer types in the GSL are proposed for standardization -- hence documented in the proposals. Are you asking for the Core Guidelines (not this repo) to expand on the spec?

MikeGitb:
@gdr-at-ms: I think a general point could be made (and I think there is an according issue in the core guidelines repo) that the gsl should actually be specified in the core guidelines in order to encourage multiple independent implementations of the same (instead of them tracking the microsoft gsl). It is however my impression that working out the details of the gsl (as well as the coreguidelines themselves) is still an ongoing effort, so I'd say no or a very rough specification is better than an unstable and/or outdated one. 

Anyway, I was not so much asking for a documentation of the (ms-) gsl (although I would certainly welcome one), than making a point that - at the moment - the code very much IS the documentation. So my guess would be that even people that "just want to use gsl types" will want to read those header files now and then and hence would welcome working syntax highlighting.

As far as the proposals are concerned (I assume you mean P0122R3 and P0123R2) : 

 1. As you said yourself, they only cover part of the gsl (e.g not multi_span)
 2. As far as I can tell, they are referenced nowhere in this repo (please correct me if I'm wrong)

gdr-at-ms:
You're correct that the Microsoft/GSL repo is meant to be **one** implementation of the spec in the Core Guidelines, not an implementation that people would have to copy.  So, people don't have to argue or make that point. And clearly, both the Guidelines and the GSL (tracking it) are evolving.

PR #319, opened 2016-08-10
Removed .h extension from header files.

galik:
As discussed in [PR #317](https://github.com/Microsoft/GSL/pull/317)


neilmacintosh:
Thanks @galik, this looks good to me.

PR #324, opened 2016-08-13
Fix #323 - Badly positioned noexcept

### LINKED ISSUE #323: noexcept in span causing error in gcc 5.4.0
Simple test program 

```
#include "gsl"
#include <iostream>

void test(gsl::not_null<int*> value)
{
   *value = 10;
}

int main()
{
   int* value = new int();
   *value = 0;

   test(value);

   std::cout << "Value is " << *value << " \n";
   delete value;
}
```

Error:
In file included from gsl:25:0,
                 from test.cpp:1:
span: In constructor 'constexpr gsl::details::span_iterator<Span, IsConst>::span_iterator()':
span:182:63: error: expected '{' before 'noexcept'
         constexpr span_iterator() : span_iterator(nullptr, 0) noexcept {}

Putting the noexcept before the : does correct the issue.

`constexpr span_iterator() noexcept : span_iterator(nullptr, 0) {}`

vincentisambart:
Fixes issue #323 by moving the `noexcept` in the declaration of `span_iterator()` to be at a valid position.
Note that the test suite doesn't catch this because it defines `GSL_THROW_ON_CONTRACT_VIOLATION` that causes `noexcept` to be defined to `/* nothing */`.

Also add a missing newline at the end of `gsl/gsl_byte`.


neilmacintosh:
@vincentisambart thanks for the PR! That looks great to me.

neilmacintosh:
@nickenchev None of the tests allow `noexcept` to exist as a keyword. They remove the restriction in order to throw+catch (so they can test failure cases).

Normally I test compilation with all the flag-combinations before I check in, so that we catch things like a `noexcept` in the wrong place, but for the iterator changes I skipped it and (predictably) introduced the bug.

I noticed that it broke a different product build I have when it picked up the GSL head, but then a weekend happened. Thanks to @vincentisambart for picking it up and fixing it so quickly!

nickenchev:
Thanks for explaining Neil!

rianquinn:
Does the TravisCI have a matrix entry for compiling non-tested, yet supported configurations? From what I could it doesn't look like it, but I might be missing something. You might not be able to test the "noexcept" case, but at least you might be able to determine if it builds automatically via TravisCI

PR #326, opened 2016-08-14
Fix inconsistent naming

EFanZh:
Because [C++ Core Guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md) is using the name “Guideline support library” instead of “Guidelines support library”.


neilmacintosh:
@EFanZh thanks for the PR, looks good to me!

EFanZh:
@neilmacintosh You may also want to modify the GitHub repository description.

PR #328, opened 2016-08-16
fix use of gsl::at with span (broken by PR #309)

ericLemanissier:
PR #309 removed `using value_type = ValueType;` from span class, which broke the use of gsl::at with a span argument. This PR re-enables it.

### DIFF
--- a/gsl/span
+++ b/gsl/span
@@ -377,6 +377,8 @@ public:
     using index_type = std::ptrdiff_t;
     using pointer = element_type*;
     using reference = element_type&;
+    
+    using value_type = ElementType;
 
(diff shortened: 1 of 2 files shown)


neilmacintosh:
@ericLemanissier thanks for the PR, however, I'd prefer to handle this differently, by specializing `at` explicitly for `span`. This allows you to skip the extra `Expects`, which `span` already provides and doesn't add an additional type alias to `span`.

The `value_type` alias was correctly removed from `span` as it is not a container (`value_type` is one of the requirements in the standard library for a container type) but rather a view type (so we use 'element_type' instead).

Feel free to offer a new PR with the alternate change...

PR #334, opened 2016-08-19
Fix issue: Provide platform support for strnlen

rianquinn:
Not all platforms have support for strnlen (e.g. cygwin) as this is a non-standard function. This patch provides a custom implementation of strnlen for these platforms.


neilmacintosh:
@rianquinn looks good to me! Thanks for putting this together.

rianquinn:
No problem

PR #338, opened 2016-08-23
Specialize gsl::at for span

ericLemanissier:
span being a view and not a container, the generic version of gsl::at is not valid any more for span.
This commits adds a specialization of gsl::at for span


neilmacintosh:
@ericLemanissier Thanks! Looks good to me.

Issue #340, opened 2016-08-24
Define constexpr as /*constexpr*/ where applicable


martinmoene:
Files gsl, gsl_byte, gsl_util, multi_span, span and string_span contain
```
#define constexpr /* nothing */
```
I'd argue it's more useful to have
```
#define constexpr /*constexpr*/
```
might one want to look at preprocessed code.

Idem for `noexcept`.

neilmacintosh:
@martinmoene 
a perfectly reasonable suggestion! Would you like to provide a PR with the change?

martinmoene:
Created PR #342.

PR #342, opened 2016-08-24
Define constexpr as /*constexpr*/ when unavailable (also noexcept)

martinmoene:
This PR resolves issue #340.

To give the commented-out `constexpr` the appearance of a single element, there's no whitespace between `constexpr` and `/*`, `*/` .

### DIFF
--- a/gsl/gsl
+++ b/gsl/gsl
@@ -30,13 +30,13 @@
 
 // No MSVC does constexpr fully yet
 #pragma push_macro("constexpr")
-#define constexpr
+#define constexpr /*constexpr*/
 
 // MSVC 2013 workarounds
 #if _MSC_VER <= 1800
 // noexcept is not understood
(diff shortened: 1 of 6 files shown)


gsl/gsl
```diff
@@ -30,13 +30,13 @@
...
 #pragma push_macro("constexpr")
-#define constexpr
+#define constexpr /*constexpr*/
 
 // MSVC 2013 workarounds
 #if _MSC_VER <= 1800
 // noexcept is not understood
 #pragma push_macro("noexcept")
-#define noexcept
+#define noexcept /*noexcept*/
```
gdr-at-ms: Please don't #define 'noexcept' this way.  It is also an operator.

neilmacintosh:
Thanks @martinmoene, looks good to me!

Issue #248, opened 2016-01-17
std::copy doesn't work on 2 span


vlj:
CppCoreGuidelines rule I.13 mention function 
"void copy(span<const T> r, span<T> r2); // copy r to r2"
but I don't find it it gsl code.
std::copy on the other hand requires 3 arguments and requires iterators.

neilmacintosh:
It's a good point @vlj. It would be nice to add a simpler `copy()` overload to the GSL.

This would presumably become unnecessary once the "ranges" versions of copy makes it into the standard (assuming a simple conversion between `span`s and ranges). 

interested in offering a PR to add `copy()`?

chfast:
Should that be a `std::copy()` overload? Does the standard allow overloading functions from `std` namespace?

neilmacintosh:
@chfast that can't be a `std::copy` overload, as overloading parts of the standard library is not ok. However, there is nothing preventing someone offering a PR with a `gsl::copy` that can be used until ranges makes its way into the standard library.

neilmacintosh:
This has now been addressed in PR #344.

PR #344, opened 2016-08-29
Add a copy function for span as mentioned in issue  #248

MikeGitb:
**Update:** The original eiscription is completely outdated -- see the following posts.

-----------------
**Original:**
This PR implements a copy function for span similar to what was mentioned in issue #248. However, instead of void, it returns a span representing the remaining space in the destination span.

There are probably a lot of style issues to be addressed and I'd be happy to incorporate any feedback, but my main question at the moment would be if the design is OK and if that function should be put into the `span` header as I did or into some separate utility / algorithms header.

### DIFF
--- a/tests/algorithm_tests.cpp
+++ b/tests/algorithm_tests.cpp
@@ -1,0 +1,209 @@
+///////////////////////////////////////////////////////////////////////////////
+//
+// Copyright (c) 2015 Microsoft Corporation. All rights reserved.
+//
+// This code is licensed under the MIT License (MIT).
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+///////////////////////////////////////////////////////////////////////////////
+
+#include <UnitTest++/UnitTest++.h>
+#include <gsl/gsl_algorithm>
+
+#include <array>
+
+using namespace std;
+using namespace gsl;
+
+SUITE(copy_tests)
+{
+
+    TEST(same_type)
+    {
+        // dynamic source and destination span
+        {
+            std::array<int, 5> src{1, 2, 3, 4, 5};
+            std::array<int, 10> dst{};
+
+            span<int> src_span(src);
+            span<int> dst_span(dst);
+
+            copy(src_span, dst_span);
+            copy(src_span, dst_span.subspan(src_span.size()));
+
+            for (std::size_t i = 0; i < src.size(); ++i) {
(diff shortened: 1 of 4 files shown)


MikeGitb:
I put the function and tests into separate files for now to prevent re-basing conflicts in the future. I'd be happy to move the contents into existing files if that is preferred.

MikeGitb:
Upon revisiting this PR I realized that returning the remaining destination space is maybe not the most intuitive semantics. It was born from from my primary use case to successively fill a buffer, but that might not be such a common use case in the wild.

Possible alternatives:
- Return a span over the copied elements (the inverse from what it returns now).
- Return the number of copied elements (which is always src.size()).
- Don't return anything for now.

Any thoughts?

MikeGitb:
After seeing this version again in @BjarneStroustrup's keynote slides, I settled for the simple version that returns nothing. 

The question if and what the function should return in addition can then be discussed separately.

tests/algorithm_tests.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+        copy(src_span, dst_span.subspan(src_span.size()));
+
+        for (std::ptrdiff_t i = 0; i < static_cast<std::ptrdiff_t>(src.size()); ++i) {
+            CHECK(dst_span[i] == src[i]);
+            CHECK(dst_span[i + src.size()] == src[i]);
+        }
+    }
+
+    TEST(copy_span_static_diff_type)
```
neilmacintosh: This is for a type that *is* assignable. Could we have a test that validates that the reverse assignment (for example) does not compile? Similarly, we should probably validate that you can't accidentally slice objects using this function.

MikeGitb:
Thanks for the feedback. The header and footer where dump copy past from another file and didn't clean it up afterwards. I'll remove the things you mentioned and expand the tests.

tests/algorithm_tests.cpp
MikeGitb: Should slicing really be prevented? Although it is rarely necessary, I actually like the fact, that std::copy / std::copy_n allow copying from one type to another. If that should not be the case, then we probably should make input and output type identical (modulo constness and maybe volatile)

MikeGitb:
@neilmacintosh: I think I adressed all your points, except the following:

 - Regarding the return type: Personally I'd also prefer to return the remaining target range, I just followed the specification presented by @BjarneStroustrup. The other thing is that the iterator returned by `std::copy_n` can be viewed both as the end of the copied elements or the start of remaining range. For `gsl::copy`, we have to make a decision for one ot the other. 
 In any case, I think this can be discussed in a separete issue/PR. Adding a return value is not a big problem 

- The question about slicing and narrowing conversion: I'm following the convention by the stl, but we can make it more conservative e.g. by 
 - requiring that source and destination span have the same element types
 - Allowing different types, but but not allowing narrowing conversions

neilmacintosh:
@MikeGitb I think you are correct to follow the `std::copy_n` convention here regarding the slicing and narrowing conversions. It's not perfect, but it will be unsurprising to follow existing standard library practice there.

I had a think about the return value, and although Bjarne`s slides demonstrate a void return type, I think that - again - it would be good to stick with the conventions set by the standard library (except in cases where we think they are bad practice). In this case, it seems relatively easy to return a `span` which runs from one past the last-written element in `dest` to `dest.end()`. If there are no unwritten elements remaining in `dest`, then you would just return an empty span. I think this also supports use cases like your original one - where you want to progressively fill a buffer.

However, I can understand if you'd like to take a second PR to add the return type. This has been sitting around for a while ;-) 

Let me know your preference...

MikeGitb:
@neilmacintosh: As the decision about the return type is made it doesn't really matter for me, but if you are fine with the rest, I'd prefer if you merge the PR as it is. Adding the return type is simple, but in case the code and/or unittest changes introduce any issues, I'd like to discuss/fix them separately, as I only work on this very irregularly.

neilmacintosh:
Makes sense @MikeGitb. I'm happy with everything in this PR, and agree we can always change the return type separately. Thanks for taking on this work!

Issue #247, opened 2016-01-15
to_string does not support non-default allocators.


garyfurnish:
Right now I have as a replacement: 
```
  template <typename gCharT,                                                                                                                                                                                                                                                       
            typename CharT = ::std::remove_const<gCharT>,                                                                                                                                                                                                                          
            typename Traits = typename ::std::char_traits<CharT>,                                                                                                                                                                                                                  
            typename Allocator = ::std::allocator<CharT>,                                                                                                                                                                                                                          
            ptrdiff_t Extent,                                                                                                                                                                                                                                                      
            typename = typename ::std::enable_if<::std::is_same<CharT, typename ::std::remove_const<gCharT>::type>::value>::type>                                                                                                                                                  
  std::basic_string<CharT, Traits, Allocator> to_string(basic_string_span<gCharT, Extent> view)                                                                                                                                                 
  {                                                                                                                                                                                                                                                                                
    return {view.data(), static_cast<size_t>(view.length())};                                                                                                                                                                                                                      
  }                                                                                                                                                                                                                                                                                
}
```
I think this is pretty bad, as you can't both type deduce the type of gCharT and avoid redundancy in the template argument list when specifying a non-default string type.  Making it a separate function would solve this problem by getting rid of the default behavior (maybe to_string_full which takes the same template arguments as a basic string).  Is this something you'd be interested in getting a PR for?

neilmacintosh:
The standard library uses `to_string()` and `to_wstring()` for strings with the "well-known" char types and with default allocators. So we should probably use SFINAE to match those two cases and then I think a "full" function than allows any char-type and specification of a non-default allocator could be called `to_basic_string()'...because that is really what it would be constructing. Sound reasonable?

neilmacintosh:
@garyfurnish are you still interested in putting together a PR for this issue?

garyfurnish:
Yeah I can put one together this week.

PR #349, opened 2016-09-04
Add to_basic_string

garyfurnish:
See #247 
Tested on 2015 RC3 and gcc version 6.2.0 20160830.


neilmacintosh:
Thanks @garyfurnish, looks good to me!

PR #351, opened 2016-09-06
gsl_assert: GSL_MESSAGE_ON_CONTRACT_VIOLATION

wilhelmtell:
acknowledge GSL_MESSAGE_ON_CONTRACT_VIOLATION for outputting a
pre/post-condition message when terminating.

this is only relevant when GSL_TERMINATE_ON_CONTRACT_VIOLATION is
defined.

this to make Expect() and Ensure() feel more intuitive in comparison with the standard assert().

### DIFF
--- a/gsl/gsl_assert
+++ b/gsl/gsl_assert
@@ -22,8 +22,10 @@
 #include <exception>
 #include <stdexcept>
 
+#if defined(GSL_TERMINATE_ON_CONTRACT_VIOLATION)
 #if defined(GSL_MESSAGE_ON_CONTRACT_VIOLATION)
 #include <iostream>
+#endif
 #endif
 
 //
--- a/gsl/gsl_assert
+++ b/gsl/gsl_assert
@@ -22,6 +22,10 @@
 #include <exception>
 #include <stdexcept>
 
+#if defined(GSL_MESSAGE_ON_CONTRACT_VIOLATION)
+#include <iostream>
+#endif
+
 //
 // There are three configuration options for this GSL implementation's behavior
 // when pre/post conditions on the GSL types are violated:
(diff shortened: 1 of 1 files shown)


wilhelmtell:
I thought it would be intuitive if, in the case of `GSL_TERMINATE_ON_CONTRACT_VIOLATION`, there would be a message printout with the filename and line number, just like assert() does.

I realized while writing this that nowhere in the code so far is `<iostream>` or even `<cstdio>` included. Is the PR is okay, with the include guard around iostream?

Also, if the general idea of the PR is acceptable, I don't know which would be better here: iostream or cstdio? I think cstdio is faster to compile.

Also, I don't know if the name I quickly picked for the define is okay. Maybe `GSL_TERMINATE_MESSAGE_ON_CONTRACT_VIOLATION` or something akin to that would be better, to express it only makes sense when `GSL_TERMINATE_ON_CONTRACT_VIOLATION` is defined.

wilhelmtell:
Oops, I just realized that the iostream include guard should also check for `GSL_TERMINATE_ON_CONTRACT_VIOLATION`.

neilmacintosh:
@wilhelmtell I like the general idea: making the emission of a message to `cerr` before terminating a configurable option. I need to think a bit more about what the name should be and whether it's an additional option once GSL_TERMINATE_ON_CONTRACT_VIOLATION is already selected or just an alternative option (that's where I'm leaning).

gsl/gsl_assert
```diff
@@ -62,10 +72,25 @@ struct fail_fast : public std::runtime_error
 
 #elif defined(GSL_TERMINATE_ON_CONTRACT_VIOLATION)
 
+#if defined(GSL_MESSAGE_ON_CONTRACT_VIOLATION)
+#define Expects(cond)                                                                              \
+    if (!(cond)) {                                                                                 \
+        std::cerr << "GSL: Precondition failure at "                                               \
+                     __FILE__ ": " GSL_STRINGIFY(__LINE__) "\n";                                   \
```
KindDragon: Also please remove space after `__FILE__`, CLion can automatically detect format `FILE:LINE` in output

wilhelmtell: ? I'm not sure I see any space after `__FILE__` ..

KindDragon: Sorry, I mean after colon `": "`

Epholys:
Any updates for this merge? I'd really like to see this implemented, that's a nice feature.

JordanMaples:
#761

JordanMaples:
Maintainer's call: Closing this PR. Please refer to Herb's reply from #761 https://github.com/microsoft/GSL/pull/761#issuecomment-540727735.

Issue #329, opened 2016-08-16
There does not seem to be a clean way to initialize gsl::byte


KrisTC:
At my company we recently ran into a severe defect in the field that is caused by use of a char* as a byte. We have fixed the issue, but I would rather switch our definition of byte to use the GSL one. However I it doesn't define any operator or method to initialize from an integer.

I saw the submition for the standard type: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0298r0.pdf and followed it's initalization example but unserprisingly it doesn't work. I am using Visual Studio 2013 compiler with warning level 4.

```
gsl::byte x = {0x10};
```
Gives the following error:
```
1>------ Build started: Project: TestApp, Configuration: Debug Win32 ------
1>  TestApp.cpp
1>f:\projects\research\coreguidlines\testapp\testapp\testapp.cpp(12): error C2440: 'initializing' : cannot convert from 'initializer-list' to 'gsl::byte'
1>          Reason: cannot convert from 'int' to 'gsl::byte'
1>          Conversion to enumeration type requires an explicit cast (static_cast, C-style cast or function-style cast)
========== Build: 0 succeeded, 1 failed, 0 up-to-date, 0 skipped ==========
```
Using a static cast works, but is ugly I was wondering if a to_byte method to match the to_intetger method would make the usage a bit nicer:

```
        auto b1 = gsl::to_byte(0x01);                   // instead of:
        auto b2 = static_cast<gsl::byte>(0x01);
```
Or can people think of something nicer?

RicoAntonioFelix:
You can define a `UDL` for your codebase

```
gsl::byte operator""_gslb(unsigned long long value)
{
	return static_cast<gsl::byte>(value);
}
```
Now you can `gsl::byte bitmap = 0xFD_gslb;`

KrisTC:
Thanks @RicoAntonioFelix 
That is a nice idea, unfortunately in my specific case I am stuck with VS2013.
`UDL` is not supported until VS2015

gdr-at-ms:
The example in that paper uses the new rules for constructing values of enumeration types: http://www.open-std.org/JTC1/SC22/WG21/docs/papers/2016/p0138r2.pdf which were adopted into C++17 draft at the Spring 2016 meeting in Jacksonville, Florida.

[...]

PR #352, opened 2016-09-06
Added to_byte method for issue #329

KrisTC:
I have added the to_byte method and updated the unit tests.

This gives slightly nicer syntax than static_cast, is better than the
c-style cast used in the unit test.

See: https://github.com/Microsoft/GSL/issues/329#issuecomment-240588515

I have tested my change on:
* VS2013
* VS2015
* Apple LLVM version 7.3.0 (clang-703.0.31)

Hopefully everything passes

### DIFF
--- a/gsl/gsl_byte
+++ b/gsl/gsl_byte
@@ -106,14 +106,15 @@ constexpr IntegerType to_integer(byte b) noexcept
     return {b};
 }
 
-template<bool, typename T>
+template<bool E, typename T>
 constexpr byte to_byte_impl(T t) noexcept
 {
     static_assert(
-        false, 
+        E,
         "gsl::to_byte(t) must be provided an unsigned char, otherwise data loss may occur. "
         "If you are calling to_byte with an integer contant use: gsl::to_byte<t>() version."
-    )
+    );
+    return static_cast<byte>(t);
 }
 template<>
 constexpr byte to_byte_impl<true, unsigned char>(unsigned char t) noexcept
--- a/gsl/gsl_byte
+++ b/gsl/gsl_byte
@@ -106,11 +106,33 @@ constexpr IntegerType to_integer(byte b) noexcept
     return {b};
 }
 
-constexpr byte to_byte(unsigned char i) noexcept
+template<bool, typename T>
+constexpr byte to_byte_impl(T t) noexcept
 {
-    return static_cast<byte>(i);
+    static_assert(
+        false, 
+        "gsl::to_byte(t) must be provided an unsigned char, otherwise data loss may occur. "
+        "If you are calling to_byte with an integer contant use: gsl::to_byte<t>() version."
+    )
+}
+template<>
+constexpr byte to_byte_impl<true, unsigned char>(unsigned char t) noexcept
+{
+     return byte(t);
 }
 
+template<typename T>
+constexpr byte to_byte(T t) noexcept
+{
+     return to_byte_impl<std::is_same<T, unsigned char>::value, T>(t);
+}
+
+template <int I>
+constexpr byte to_byte() noexcept
+{
+    static_assert(I >= 0 && I <= 255, "gsl::byte only has 8 bits of storage, values must be in range 0-255");
+    return static_cast<byte>(I);
+}
 
 } // namespace gsl
 
--- a/gsl/gsl_byte
+++ b/gsl/gsl_byte
@@ -106,6 +106,12 @@ constexpr IntegerType to_integer(byte b) noexcept
     return {b};
 }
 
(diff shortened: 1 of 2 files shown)


gsl/gsl_byte
```diff
@@ -106,6 +106,12 @@ constexpr IntegerType to_integer(byte b) noexcept
     return {b};
 }
 
+constexpr byte to_byte(unsigned char i) noexcept
+{
```
gdr-at-ms: This version allows "accidental" narrowing conversion. Could you cook a version that weed out non-'unsigned char' via sfinae trickery?  E.g. something along the line of
```
template<typename T, typename = std::enable_if_t<std::is_same<T, unsigned char>::value, T>>
constexpr byte to_byte(T t) noexcept
{
     return byte(t);
}
```

KrisTC: I have confirmed that in Apple LLVM version 7.3.0 (clang-703.0.31) I definitely get accidental narrowing. I am used to C4244 from MS compiler. I tried simply putting your suggestion in, but it fails the unit test on the constants:
```
byte a = to_byte(0x00); // Error with is_same<T, unsigned char>
```
I have had a look at the available traits: http://en.cppreference.com/w/cpp/header/type_traits and I am not seeing anything else that will help.
I played about with overloads static asserts:
```
template <int I>
constexpr byte to_byte() noexcept
{
    static_assert(I >= 0 && I <= 255, "Value must be in range 0-255");
    return static_cast<byte>(I);
}

template <int I>
constexpr byte to_byte_v = to_byte<I>();
```
Which give the following for constants:
```
byte a = to_byte<0x00>();       // a bit odd
byte b = to_byte_v<0x00>;       // maybe better, still nicer than:
byte c = static_cast<byte>(0x00);
```
But I think they are a bit weird. I will think about it some more and see if there is anything I can come up with.

KrisTC: I also experimented with initializer_list's to get syntax like:
```
byte d = to_byte({0x00});
```
Which I think looks better and works better. 
But allows invalid calls:
```
byte e = to_byte({});
byte f = to_byte({0x00, 0x00});
```
I can't see anyway ensure 1 entry.
@gdr-at-ms: Do options a, b or d look like they might be acceptable? Option b won't work with VS2013 :(

KrisTC:
I ended up going with: `gsl::to_byte<0x00>()` for constants and `gsl::to_byte(some_uchar)` for variables. I didn't keep the `gsl::to_byte_v<0x00>` version as it doesn't work with VS2013 and caused my VS2015 to have an internal compiler error. 
I have a static asserts for cases where the wrong type is passed or the constant is out of range, giving dev instructions on how to use to_byte correctly.

I think this is the best I could come up with, I hope it's ok.

neilmacintosh:
@KrisTC just quickly...if you saw an internal compiler error from VS2015, could you please send us a repro? We take fixing those crashes very seriously and I want to make sure we get that one.

KrisTC:
@neilmacintosh sure, I will do that tomorrow. I still wasn't keen on having the API `#if def'd` out for VS2013. I think it is better to have consistency.

KrisTC:
@neilmacintosh, I went to report the error using VS feedback and it is already logged: "constexpr variable template cause internal compiler error" (last modified 10th Aug) - I up voted it.

neilmacintosh:
@KrisTC perfect, thanks! I agree we don't want APIs that are `#ifdef`d out if at all possible, I just wanted to make sure that Microsoft was tracking the defect in the compiler. Thanks for upvoting that bug.

KrisTC:
@gdr-at-ms is there anything else you would like me to do for this?
Thanks Kris

gdr-at-ms:
The distinction between constants and runtime value is a little bit troubling, but I suspect this is strictly an improvement over the current situation, so it looks good to me.  
@neilmacintosh , would you want me to pull the merge?

KrisTC:
@gdr-at-ms, I agree, I couldn't come up with anything cleaner :(
I assume the proposed changes to the standard will have the same issues as my my methods:

```
std::byte b1 = 0xFFFF; // Compile's ok but narrowing
std::byte b2 = {0xFFFF}; // Compile's fail!
std::byte b3 = 0xFF; // OK
std::byte b4 = {0xFF}; // OK
```

gdr-at-ms:
@KrisTC With the proposed changes to the standard (which are in the CD out for ballot), you need to systematically use the braces, and narrowing will be enforced.  There is no provision to convert from *bare integer* to an enum -- that has always been banned in C++.

neilmacintosh:
LGTM too! Thanks for taking this on @KrisTC and congratulations on your first PR!

PR #356, opened 2016-09-13
Added inline to fix compilation warnings on msvc 2013

s-brain:
Added inlines to avoid these compilation warnings on msvc13 where constexpr is not available

warning LNK4006: "enum gsl::byte & __cdecl gsl::operator|=(enum gsl::byte &,enum gsl::byte)" (??_5gsl@@YAAEAW4byte@0@AEAW410@W410@@Z) already defined in Algorithm.obj; second definition ignored

### DIFF
--- a/gsl/gsl_byte
+++ b/gsl/gsl_byte
@@ -26,7 +26,7 @@
 
 // constexpr is not understood
 #pragma push_macro("constexpr")
-#define constexpr /*constexpr*/
+#define constexpr /*constexpr*/ 
 
 // noexcept is not understood
 #pragma push_macro("noexcept")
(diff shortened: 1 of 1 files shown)


gsl/gsl_byte
```diff
@@ -26,7 +26,7 @@
 
 // constexpr is not understood
 #pragma push_macro("constexpr")
-#define constexpr /*constexpr*/
+#define constexpr /*constexpr*/
```
martinmoene: Introducing unwanted trailing space.

neilmacintosh:
Thanks @s-brain, this looks good to me. I can live with the trailing space for now...the next time we run clang-format over the source code it should be get picked up (I assume).

PR #365, opened 2016-09-17
Add tests for to_integer(), fix to_integer()

martinmoene:


neilmacintosh:
Thanks for the PR @martinmoene. That's very much appreciated!

PR #366, opened 2016-09-17
fix wrong types in multi_span_test (<comparison_operators>)

MikeGitb:
I assume those checks where meant to compare `multi_span` variables that where initialized with `nullptr`, not `nullptr_t` variables.


neilmacintosh:
You are correct @MikeGitb :-) Thanks for the PR, it looks great to me!

PR #368, opened 2016-09-18
Negative indices to multi_span are not flagged with neither exception nor abort.

kyku:
I augmented the tests of multi_span to show that you can pass negative indices and they are not detected properly. The second commit to my branch fixes the `Expects()` checks that cause the problem, so that the new checks pass.

```
    TEST(bounds_checks)
    {
        int arr[10][2];
        auto av = as_multi_span(arr);

        fill(begin(av), end(av), 0);

        // ...

        CHECK_THROW(av[-1][0], fail_fast); // ... does not throw before the fix
        CHECK_THROW((av[{-1, 0}]), fail_fast);  // ... does not throw before the fix
}
```

### DIFF
--- a/gsl/multi_span
+++ b/gsl/multi_span
@@ -443,7 +443,7 @@ namespace details
         template <typename T, size_t Dim = 0>
         size_type linearize(const T& arr) const
         {
-            Expects(arr[Dim] < CurrentRange); // Index is out of range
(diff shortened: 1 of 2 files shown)


neilmacintosh:
Nice fix @kyku, thanks for providing the PR!

Issue #313, opened 2016-08-07
span<T> constructors allow unsafe conversions


bogiord:
`std::is_convertible<typename Container::pointer, pointer>` is the wrong test to constrain the constructor from a container. Example:

```c++
#include <vector>
#include <iostream>
#include "gsl.h"

struct B { int i = 7; };
struct D : B { int j = 0; };

int main()
{
   std::vector<D> v = {{}, {}, {}};
   gsl::span<B> s = v;
   std::cout << s[1].i << '\n'; // undefined behaviour, may print 0
}
```

The same problem is present if we construct the `span` from `std::array<D, 3>`, because the constructor from `std::array<ArrayElementType, N>&` is not constrained and then `storage_type` uses `pointer` to store the pointer to the first element, allowing derived-to-base pointer conversions. (By the way, the default template argument for `ArrayElementType` is pointless, it will always be deduced.)

Now, conversions between `span`s: `is_allowed_element_type_conversion` is way too permissive. 

`is_allowed_integral_conversion` is the wrong test in this context. You can't look at an array of `int`s through a `span<long>`, even if `int` and `long` have the same internal representation; this is breaking aliasing rules and is undefined behaviour according to [basic.lval]/8 in the Standard (N4606, the current draft).

`is_allowed_pointer_conversion` is also wrong. Just because a pointer type is convertible to another it doesn't mean that an array of the former is convertible to an array of the latter.

And on top of all this, the `reinterpret_cast<pointer>(other.data())` is the icing on the explosive cake. One example involving pointers:

```c++
#include <iostream>
#include "gsl.h"

struct B1 { int i = 7; };
struct B2 { int j = 3; };
struct D : B1, B2 { int k = 0; };

int main()
{
   D a[] = {{}, {}};
   D* ap[] = {&a[0], &a[1]};
   gsl::span<D*> sd = ap;
   gsl::span<B2*> sb = sd;
   std::cout << sb[0]->j << '\n'; // undefined behaviour, may print 7
}
```

On the other hand, `std::is_same<From, std::remove_cv_t<To>>` disallows conversions like `volatile T` to `const volatile T`, which are safe (it also disallows identical cv-qualified types, like `const T` to `const T`, but those cases are handled by the copy and move constructors, so don't use this test).

[...]

Issue #34, opened 2015-09-24
Where can I find the documentation of the library?


viboes:

neilmacintosh:
The best place to look for documentation around usage of the library is currently the Core Guidelines over at https://github.com/isocpp/CppCoreGuidelines.

viboes:
I believed the GSL was ready to use. Without  documentation How to know if something is part of the library? 

I would label the issue as a task.

viboes:
Sorry, but what is a deferred closed issue. Cannot we reopen an issue?

neilmacintosh:
Yes, it can always be reopened.

I'm not sure how it got the "deferred"..I would blame it on my fingers being inaccurate when choosing a label from the list.

It's closed because the specification for the library will live over with the Core Guidelines (and is currently under development). This is tracked by https://github.com/isocpp/CppCoreGuidelines/issues/144.

PR #369, opened 2016-09-18
[RFC- Don't merge] Alternate byte implementation for non VC / pre c++17 compiler

MikeGitb:
As discussed in issue #313, the current implementation of byte leads to undefined behavior when data is accessed through a pointer to `gsl::byte` (contrary to what we are used with e.g. `unsigned char*`). Due to strict alias assumption in g++ and clang++ 's optimizer, this can demonstrably lead to errors in certain cases, while VC++ seems not to be affected.

This PR implements two alternative versions of `gsl::byte`
- As a simple typedef for `unsigned char`, which gets rid of the UB, but also looses the type safety properties of the enum version.
- as a struct wrapping a single unsigned char as used in the gsl-lite by @martinmoene. 

Warning: I believe the second version is still undefined behavior and it leads to failures on my system. I can't reproduce those failures with the gsl-lite however which - I believe - has the same implementation. So maybe those failure are only caused by a bug in my implementation.


gsl/gsl_byte
```diff
@@ -36,8 +36,167 @@
...
+    explicit operator unsigned char&() { return v; }
+    explicit operator const unsigned char&() const { return v; }
+    friend bool operator==(byte l, byte r) { return l.v == r.v; }
+    friend bool operator!=(byte l, byte r) { return l.v != r.v; }
+};
+
+template <class IntegerType, class = std::enable_if_t<std::is_integral<IntegerType>::value>>
+inline constexpr byte& operator<<=(byte& b, IntegerType shift) noexcept
+{
+    return b.v <<= shift, b;
```
martinmoene: Funny, initially was going to write it as here, but in the end chose `b.v <<= shift; return b;`

martinmoene:
When compiling `byte_tests` via CMake with g++ (GCC) 5.2.0 on Windows it selects `GSL_BYTE_TYPE_ENUM` and fails the `aliasing` test.

In the same situation, but with `GSL_USE_BYTE_TYPE` defined as `GSL_BYTE_TYPE_STRUCT` the aliasing test passes.

gsl/gsl_byte
MikeGitb: I like your version better, but tried to make it compatible with c++11 constexpr rules (which is somewhat pointless, as the ms-gls doesn't work with c++11 anyway). Do you have any Idea, why the unit test fails here but not in your gsl-lite?

MikeGitb:
@martinmoene: Sorry, my mistake - I tested the different values for ´GSL_USE_BYTE_TYPE` prior to creating the PR and then accidentally committed the default enum version instead of the struct version.
Strangely enough, with gcc 6.1 on fedora it still fails for me.

gsl/gsl_byte
martinmoene: Did you specify `@` or `"[.issue #34]"` (with or  without `--pass`) as below to actually invoke the alias test for gsl-lite?

>prompt> gsl-lite-cpp14.t.exe --pass "[.issue #34]"
path/to/issue.t.cpp:47: passed: byte: aliasing rules lead to undefined behaviour when using enum class[.issue #34](GSL #313): 14 == F::f( i, reinterpret_cast<gsl::byte&>( i ) ) for 14 == 14
All 1 selected test passed.

(Did note the s/test/tests/)

MikeGitb: no, I just called `make test`  and assumed that would call all tests. But yes, even after explicitly selecting the test as shown by you, it passes.

MikeGitb:
Rebased to master and integrated fix to select struct based byte by default on non-VC++ compiler

MikeGitb:
Ok, the new travis tests also show that the struct version doesn't reliably work on gcc 6 but on all other. So I guess the question is if 

1.  I should keep all three versions and use the "best" one (`enum` > `struct` > `unsigned int`) that works on a certain compiler.
2. The same as 1. but only with the enum and unsigned int versions
3. Leave the `gsl/gsl_byte` header as it is and just accept that `gsl::byte` doesn't completely work as intended on all compilers

gdr-at-ms:
I continue to suggest `-fno-strict-aliasing` -- it sounds like the compiler is being too aggressive even with time-honored structures.

MikeGitb:
@gdr-at-ms: IIRC, neilmacintosh wasn't to happy with that solution. Personally, I don't care too much anymore, but if you two agree that this is the best way to address this bug, then that requirement should imho be mentioned somewhere (e.g. the readme).

MikeGitb:
As there isn't any new input and you seem to be happy with the solution in #390 for now, I think this PR can be closed. I won't delete the branch in the near future, so if this comes up again it can be used as a basis for further exploration.

Issue #372, opened 2016-09-20
Header "gsl/byte" should disable warning C26493


MikeGitb:
The current implementation of `gsl::byte` related operators make heavy use of c-style (function-style) casts from `unsigned char` to `gsl::byte`. This triggers warning C26493 in the CppCoreChecker downloaded via Nuget. 
Imho, either those casts should be replaced by static casts, or the the warning should be turned of in the header. I'd somewhat favor the latter option, as I assume that the function style casts (`return byte(...);`) are meant as surrogates for the almost identical brace initialization syntax (`return byte{...};`) that will come with c++17.

KrisTC:
Have they updated the version used in the nuget package yet? 
Last time I looked it was an old version that didn't compile with MSVC 2013. Headers had .h extensions etc. 
There wasn't any traceability back to the git revisions either. I thought they would update the nuget packages before cppcon.

MikeGitb:
I don't think they did, but I was referring to the corechecker package not the gsl (although there seems to be a dependency so that the nuget gsl is automatically installed alongside).

The version provided by vcpkg (https://github.com/Microsoft/vcpkg) is up to date though.

gdr-at-ms:
@MikeGitb I agree that locally turning off the warning in the header is a better workaround.

We have to remember that the GSL, by its very nature of having to hide constructs banned or discouraged by the Core Guidelines, sometimes has to do things that the consuming code isn't allowed to do.  That is why it is the *support* library :-)

MikeGitb:
> We have to remember that the GSL, by its very nature of having to hide constructs banned or discouraged by the Core Guidelines, sometimes has to do things that the consuming code isn't allowed to do. That is why it is the support library :-) 

That was also my first thought, but on the other side - as far as this particular warning is concerned - I don't think there is any case where the gsl actually has to use function-style casts. Maybe this is also true for other warnings - any thoughts on this?

MikeGitb:
Come to think of it @neilmacintosh: 

The c++corechecker currently warns about this:

    int i = 5;
    auto v = std::vector<Foo>(i);

with the warning

> warning C26493: Don't use C-style casts that would perform a static_cast downcast, const_cast, or reinterpret_cast. (type.4: http://go.microsoft.com/fwlink/p/?LinkID=620420)

[...]

PR #373, opened 2016-09-21
Turn off warning about function-styl casts in gsl_byte

MikeGitb:
As discussed in issue #372.


neilmacintosh:
Thanks @MikeGitb, looks good to me!

Issue #371, opened 2016-09-20
as_bytes from a single pod object


MikeGitb:
I hope I haven't overlooked it, but there seems to be no easy way to view a single pod variable as a sequence of bytes. Use case would e.g. be to write a header data structure and the actual payload of a network package into a tx buffer.
- Would you accept an overload of `as_bytes` and `as_writeable_bytes`, that accepts a single pod as a parameter?
- If so, should `as_bytes` accept a temporary?

neilmacintosh:
@MikeGitb this was actually in the original design (you should be able to see it in array_view if you go way back in the history). I agree with the use case, which was why we originally incorporated it.

The design back then was that you should be able to construct a `span` from a single element fairly easily. So calling `as_bytes` or `as_writable_bytes` was automatically then available, which seems the easier/more consistent route.

It should probably accept a temporary for consistency. In saying that, I haven't given much time to thinking through what all the call-resolution conflict possibilities might be though (with or without the temporary option to be honest).

I asked for opinions on re-introducing it at the Oulu WG21 meeting and people were generally favorable. So if you are interested in trying a PR, please feel free.

PR #374, opened 2016-09-22
Add overload for as_bytes taking a single object 

MikeGitb:
as discussed in #371

### DIFF
--- a/gsl/span
+++ b/gsl/span
@@ -308,7 +308,7 @@ namespace details
         constexpr extent_type() noexcept {}
 
         template <index_type Other>
-        constexpr extent_type(extent_type<Other> ext) 
+        constexpr extent_type(extent_type<Other> ext)
         {
             static_assert(Other == Ext || Other == dynamic_extent,
                           "Mismatch between fixed-size extent and size of initializing data.");
@@ -632,6 +632,22 @@ as_writeable_bytes(span<ElementType, Extent> s) noexcept
     return {reinterpret_cast<byte*>(s.data()), s.size_bytes()};
 }
 
+
+template <class T, class = std::enable_if_t<std::is_pod<T>::value>>
+span<const byte, sizeof(T)>
+as_bytes(const T& e) noexcept
+{
+    return {reinterpret_cast<const byte*>(&e), sizeof(e)};
+}
+
+template <class T, class = std::enable_if_t<std::is_pod<T>::value>>
+span<byte, sizeof(T)>
+as_writeable_bytes(T& e) noexcept
+{
+    static_assert(!std::is_const<T>::value,"Can't create a span of writeable bytes over a const object");
+    return {reinterpret_cast<byte*>(&e), sizeof(e)};
+}
+
 //
 // make_span() - Utility functions for creating spans
 //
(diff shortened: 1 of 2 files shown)


MikeGitb:
I think the check for `std::is_trivially_copyable` is not strict enough, as it will e.g. also be true for view classes like `std::string_view` or 3rd party ones. I'll fix this over the weekend.

MikeGitb:
Rebased to master and the function now only works for POD types.

gsl/span
```diff
@@ -670,6 +670,19 @@ span<typename Ptr::element_type>
 make_span(Ptr& cont)
 { return span<typename Ptr::element_type>(cont); }
 
+template <class T, class = std::enable_if_t<std::is_pod<T>::value>>
```
neilmacintosh: Actually I think we don't really need to restrict `as_bytes()` or `as_writeable_bytes()' at all (as per the `span` member functions). Isee these as the "sharp knife" power-user functions. ;-) People who are calling them should be aware of the language rules (and any compiler extensions they might have) around accessing object storage.

gsl/span
```diff
@@ -670,6 +670,19 @@ span<typename Ptr::element_type>
...
+span<const byte, sizeof(T)>
+as_bytes(const T& e) noexcept
+{
+    return{ reinterpret_cast<const byte*>(&e),sizeof(T) };
+}
+
+template <class T, class = std::enable_if_t<std::is_trivially_copyable<T>::value>>
```
neilmacintosh: I'm pretty sure the one restriction you do need here is to ensure that `T` is not `const`.

gsl/span
MikeGitb: I can drop the restrictions, but my concern is that someone who reads 

    std::vector<int> data{...};
    auto raw_bytes = gsl::as_bytes(data);

will assume that `raw_bytes` is a span over the content of the vector and not over the private member variables. Especially as I don't believe there are a lot of cases, where you should / want to access the raw memory occupied by a complex object.

gsl/span
MikeGitb: I can't check right now, but the reinterpret cast should ensure this. But Yes, I can add an explicit check for this in the interface.

LYP951018: Will the PR make code like
```cpp
const auto sp = gsl::as_bytes(PodClass{});
const auto b = sp[0];
```
boom?

MikeGitb:
@neilmacintosh: I made some small modifications, but did not fully address your comments:
- The overloads still work only for pods. I still believe that code like this:
 
        std::vector<int> v{1,2,3,4,5};
        auto bytes =as_bytes(v);

should either not compile or generate a span of bytes over the elements of the vector. However, if I'd drop the check on the new overloads, it would create a span of bytes over the internal pointers. That might sometimes be what you want, but I'm pretty sure it is not what the average programmer would expect when reading this code.

- I'm now checking for non-constness explicit, but I'm doing so via a static_assert, whereas the original overloads use enable_if. They should of course be consistent - the question is which to take. Personally, I try to restrict the use of enable_if to cases where I want to "guide" the compiler to the right overload / template specialization. In cases where there is none however (like calling as_writeable_bytes on a constant POD object), a direct error message is imho more useful to the programmer than a lookup failure.

MikeGitb:
@LYP951018: No, unfortunately code like 

    const auto sp = gsl::as_bytes(PodClass{});
    const auto b = sp[0];

would not lead to a compilation error (I asked neil about accepting temporaries in #371), but my hope would be that e.g. the coreguidelinechecker or other lint-like tools would catch it. The problem is the same as when you construct a "normal" span over a temporary container.

gsl/span
LYP951018: Maybe we can mark `as_bytes(const T&&)` as delete.

MikeGitb: @LYP951018: I had actually a few use cases for it, when the result is directly passed to a function (like `socket.write(as_bytes(1+2));`) but I'd tend to agree with you, because it is just too easy to miss those kinds of errors (especially during refactoring).

LYP951018: @MikeGitb I read the PR carefully and I think the const in as_bytes is useless. After removing the const, no temp objects will bother us.

MikeGitb: @LYP951018: That would also be a possibility. I somewhat like your first suggestion better, because it is more explicit (in that we won't take temporaries as well as that the sparameter will not be modified inside the function). In any case, I'd first wait on what neil has to say about the if`, before we go to the how.

MikeGitb:
Is there still interest in merging this PR or can I close it?

MikeGitb:
I read that as a no.

PR #377, opened 2016-09-28
Add branch prediction to Ensures / Expects

rianquinn:
We should be using branch prediction on asserts as these conditions are known to be unlikely, therefore we should be optimizing the likely case. These macros are similar to what the Linux kernel uses. Not really sure what Visual Studio does so at the moment these are disabled for VS.


neilmacintosh:
Thanks @rianquinn! This looks good to me.

PR #378, opened 2016-09-28
Fix compilation issues with Cygwin

rianquinn:
Cygwin recently did an update that broken string_span again. The original update had to provide a custom strnlen function because Cygwin doesn't implement this (it's a non-std extension). We never added wchar_t support, and the Cygwin update now breaks on exactly this. This patch provides the missing wchar_t.

### DIFF
--- a/gsl/string_span
+++ b/gsl/string_span
@@ -118,6 +118,24 @@ namespace details
         return len;
 #endif
     }
+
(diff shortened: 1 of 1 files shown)


neilmacintosh:
Thanks @rianquinn!

PR #379, opened 2016-09-28
Use Ubuntu 14.04 for Travis CI instead of 12.04

rianquinn:
Updating to Trusty provides CMake 2.8.12 by default, which should clean up an existing PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 # Based on https://github.com/ldionne/hana/blob/master/.travis.yml
 
 language: cpp
+dist: trusty
 sudo: false
 
 matrix:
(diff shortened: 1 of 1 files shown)


rianquinn:
Will need a different way to do this for a couple of reasons:
- Trusty doesn't support the container approach which is being used here
- Because containers are not used, they have not added support for 3.6 in trusty and I'm assuming 3.6 is used as a minimum baseline for support at the moment. 

Maybe once certain compilers are dropped, we can fix this. Another approach would be to update Travis CI to use a newer version of CMake. 

https://github.com/travis-ci/apt-source-whitelist/pull/205

PR #381, opened 2016-09-30
[Travis] enlarge the support matrix

CaseyCarter:
* Build on OSX with Xcode 8
* Build on Linux with Clang 3.6/7/8 and libc++
* Build on Linux with GCC 5/6

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,65 +3,140 @@
 language: cpp
 sudo: false
 
+cache:
+  directories:
+    - ${TRAVIS_BUILD_DIR}/deps/cmake
+    - ${TRAVIS_BUILD_DIR}/deps/llvm-3.5.2/install
+    - ${TRAVIS_BUILD_DIR}/deps/llvm-3.6.2/install
+    - ${TRAVIS_BUILD_DIR}/deps/llvm-3.7.1/install
+    - ${TRAVIS_BUILD_DIR}/deps/llvm-3.8.1/install
+    - ${TRAVIS_BUILD_DIR}/deps/llvm-3.9.0/install
+
 matrix:
   include:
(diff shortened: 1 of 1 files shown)


neilmacintosh:
Hey @CaseyCarter, while we wait for this to pass....do you have enough Travis expertise to be able to get us using a more recent version of CMake with our Travis builds? It's holding back a couple of PRs.

CaseyCarter:
> Hey @CaseyCarter, while we wait for this to pass....do you have enough Travis expertise to be able to get us using a more recent version of CMake with our Travis builds?

Since I like you, I will travel back in time to before I submitted this pull request and have it install a more current cmake as part of the build.

CaseyCarter:
There, done.

CaseyCarter:
> I will travel back in time

Just to be clear, this is my "funny" way of telling you that this change already has the requested behavior.

neilmacintosh:
@CaseyCarter I couldn't stop laughing. No, I could. I deserved it for lazily just reading your description and waiting for the CI checks to do the rest. Thank you very much for this magnificent PR, it makes my world a much better place.

MikeGitb:
@CaseyCarter : Is there a way to rerun the tests for existing PRs with the added compiler versions (without adding a new commit)?

rianquinn:
@MikeGitb Yes and now. You can "re-run a build" in Travis CI but since the configuration file is part of the repo, you would need to rebase (which requires a commit) to get the updated config.

MikeGitb:
@rianquinn: Thanks for the explanation. I don't want to rewrite history at the moment, as I'm not sure how that would effect past PR-comments. So I'll go for a merge for now instead.

Issue #375, opened 2016-09-24
Should gsl::span constructor check for size overflow?


MikeGitb:
As shown in issue #322, when I want to construct a span from a container, I have to ensure that the size doesn't exceed `std::numeric_limits<std::ptrdiff_t>::max()`. On 64bit systems this isn't a practical issue and - assuming that the maximum size of any continuous container is `std::numeric_limits<std::size_t>::max() / sizeof(T)` - there are also relatively few instances on other systems where this might be a problem.

Still, I wonder if the constructor should check for such an overflow (taking from @ewil's example):

     Expects(cont.size() <= static_cast<size_t>(std::numeric_limits<index_type>::max()))

rianquinn:
I want to say this was the exact same issue that I had.

neilmacintosh:
@MikeGitb The answer is "yes" ;-)

neilmacintosh:
PR #382 has addressed this problem for construction from containers. Thanks for raising the issue @MikeGitb  and @rianquinn.

PR #382, opened 2016-10-04
For span, fix size_t to index_type conversion warning 

garyfurnish:
Standard containers return size_t which is signed.  Therefore implicit conversion to index_type throws a warning which causes compiler errors on -Wall.  This static casts to avoid this.  

Passes tests on gcc version 6.2.0 20160914 (Ubuntu 6.2.0-3ubuntu15)

### DIFF
--- a/gsl/span
+++ b/gsl/span
@@ -392,7 +392,7 @@ public:
                   std::is_convertible<typename Container::pointer, pointer>::value &&
                   std::is_convertible<typename Container::pointer,
                                       decltype(std::declval<Container>().data())>::value>>
-    constexpr span(Container& cont) : span(cont.data(), cont.size())
+    constexpr span(Container& cont) : span(cont.data(), narrow<index_type>(cont.size()))
     {
     }
 
@@ -402,7 +402,7 @@ public:
                   std::is_convertible<typename Container::pointer, pointer>::value &&
                   std::is_convertible<typename Container::pointer,
                                       decltype(std::declval<Container>().data())>::value>>
-    constexpr span(const Container& cont) : span(cont.data(), cont.size())
+    constexpr span(const Container& cont) : span(cont.data(), narrow<index_type>(cont.size()))
     {
     }
 


maikel:
why not `gsl::narrow`?

garyfurnish:
Its a pretty big performance hit to have an extra branch there when you are reference converting from vector all over.  How would you overflow from size_type to index_type which is defined to be ::std::ptrdiff_t?  Wouldn't that already require undefined behavior in other places?    I'm not opposed to changing to narrow_cast which is unchecked, but is it even possible to have a container return a size() that overflows on conversion?

gdr-at-ms:
@garyfurnish Yes, it is possible.  The real questions are "how common?" and "how can we ensure that most cases don't take **measurable** hits?"

MikeGitb:
I commented on that somewhat on issue #375: I think you only have to make a check in a 32 bit program when sizeof(T) >1.

gdr-at-ms:
@MikeGitb Agreed.  An interesting question is whether the check is *measurable* for most practical uses and if it can be done more efficiently.

neilmacintosh:
@garyfurnish The other option to consider if you really think that extra branch is a big deal (and not optimizable), is to use `gsl::narrow_cast<>()` to at least document your assertion that the narrowing operation is safe.

garyfurnish:
I don't claim to be a standards expert, but I suppose I can think of a bunch of use cases where its possible away from x64, although I am skeptical that there might not be UB sitting around in implementations that hasn't been found yet.  

I'd very much like to be able to use span in performance sensitive situations as a function parameter.  If every conversion is checked, thats not possible without writing a new vector class that uses signed sizes or static casting everywhere on construction, both of which are error prone.  A check here or there may not be bad, but as an interface that check is going to get expensive quickly, even playing tricks like feeding branch predictor info to the compiler.  (Right now I'm using this writing memory allocators, so that branch is awful because it can't be inlined/optimized away, and branch predictors are finite).  Video games and latency sensitive code would also pass on using it in interfaces, and that seems like something that would be disappointing.  

A related question is can we make reasonable assumptions if the container is something we know about.  If you use a vector or array on a reasonable machine, even 32 bit, you already have potential signed overflow if you are subtracting iterators if size would overflow on conversion.  Maybe there should be a guideline of "don't do that" on twos complement machines.  (Which fits into documenting the assertion).

MikeGitb:
Concerning the performance: Is a single check upon construction really a problem? Especially when compared to the checks that are performed on every element access anyway?

garyfurnish:
It may take time to come up with a reasonable benchmark on this.  My intuition was that the extra checks should get eliminated in loops, and elsewhere it shouldn't matter.  As it happens, both clang and gcc seem to be quite good at optimizing out the check on construction when possible, so this may not even be an issue in many cases.  Unfortunately both clang and gcc seem to skip optimizing out extra checks in various places in loops.  So theoretically my answer would be yes the single check is a problem because of branch predictor pollution (pollution O(n) number of unique call sites as opposed to O(n) number of places array is accessed).  In practice its going to take some time to figure out given that clang trunk produces code with -O3 like:
```c++
extern void c(int);
void test(::gsl::span<int> span)
{
  for(auto&& i : span)
    {
      c(i);
    }
}
```
```
   0x00000000004006c0 <+0>:	push   %r15
   0x00000000004006c2 <+2>:	push   %r14
   0x00000000004006c4 <+4>:	push   %rbx
   0x00000000004006c5 <+5>:	mov    %rsi,%r14
   0x00000000004006c8 <+8>:	mov    %rdi,%r15
   0x00000000004006cb <+11>:	test   %r15,%r15
   0x00000000004006ce <+14>:	js     0x400701 <test(gsl::span<int, -1l>)+65>
   0x00000000004006d0 <+16>:	je     0x4006f6 <test(gsl::span<int, -1l>)+54>
   0x00000000004006d2 <+18>:	xor    %ebx,%ebx
   0x00000000004006d4 <+20>:	data16 data16 nopw %cs:0x0(%rax,%rax,1)
   0x00000000004006e0 <+32>:	cmp    %r15,%rbx
   0x00000000004006e3 <+35>:	jge    0x4006fc <test(gsl::span<int, -1l>)+60>
   0x00000000004006e5 <+37>:	mov    (%r14,%rbx,4),%edi
   0x00000000004006e9 <+41>:	callq  0x400710 <c(int)>
   0x00000000004006ee <+46>:	inc    %rbx
   0x00000000004006f1 <+49>:	cmp    %rbx,%r15
   0x00000000004006f4 <+52>:	jne    0x4006e0 <test(gsl::span<int, -1l>)+32>
   0x00000000004006f6 <+54>:	pop    %rbx
   0x00000000004006f7 <+55>:	pop    %r14
   0x00000000004006f9 <+57>:	pop    %r15
   0x00000000004006fb <+59>:	retq   
   0x00000000004006fc <+60>:	callq  0x400580 <_ZSt9terminatev@plt>
   0x0000000000400701 <+65>:	callq  0x400580 <_ZSt9terminatev@plt>
```
The jne on 52 can jump to 37 instead because the jge on 35 can never be true after increment.  
g++ also seems to have issues with optimizing out checks with iterators.  

So the long answer is: I don't know yet.  Improve compiler optimization issues or work around their failures, yes it will be detectable.  As is, compiler optimization issues are going to dominate in any realistic benchmark.  I also have not tried this on MSVC yet; maybe its optimizer does better.

gdr-at-ms:
This is great!
This sounds to me like a fantastic opportunity to report a missed optimization to all compilers affected.

neilmacintosh:
@garyfurnish I agree with @gdr-at-ms, your micro-benchmark shows some opportunities for optimization improvements in compilers.

Using `gsl::narrow()` here is the correct thing to do. We emphasize correctness first in the GSL. Do you want to update the PR to use `gsl::narrow()`?

garyfurnish:
Not sure if I redid the pull request correctly, but done.

neilmacintosh:
@garyfurnish it looks perfect to me. Thanks very much for the PR and the great discussion (especially the interesting analysis of optimizations).

PR #389, opened 2016-10-07
Fix #388 Deprecated implicit copy assignment operator for span_iterat…

### LINKED ISSUE #388: Deprecated implicit copy assignment operator for span_iterator
In clang trunk, with c++1z, 
```
gsl/span:161:19: error: definition of implicit copy assignment operator for 'span_iterator<gsl::span<void **, -1>, false>' is deprecated because it has a user-declared copy constructor [-Werror,-Wdeprecated]
        constexpr span_iterator(const span_iterator<Span, false>& other) noexcept
```
PR to come.

garyfurnish:
Fix #388 by explicit default.  Unlike the copy constructor this does not perform an extra Expects check.  I can redo the pr to explicitly do that if requested because check for invalid iterator is desired on operator=.

### DIFF
--- a/gsl/span
+++ b/gsl/span
@@ -163,6 +163,8 @@ namespace details
         {
         }
 
+        constexpr span_iterator<Span, IsConst>& operator=(const span_iterator<Span, IsConst>&) noexcept = default;
+
         constexpr reference operator*() const
         {
             Expects(span_);


gsl/span
```diff
@@ -163,6 +163,8 @@ namespace details
         {
         }
 
+        constexpr span_iterator<Span, IsConst>& operator=(const span_iterator<Span, IsConst>&) = default;
```
garyfurnish: I think that is intentional to allow casting iterators pointing to nonconst values to an iterator pointing to const values.  I did fix this just now to be explicitly noexcept.

neilmacintosh: @garyfurnish is correct, it is intentional to allow conversions between the two types of iterator.

MikeGitb: Right of course - it is too late ;)

neilmacintosh:
Thanks @garyfurnish, LGTM!

PR #390, opened 2016-10-08
Address strict aliasing

MikeGitb:
This is an alternative PR, related to the `gsl::byte` aliasing problem (as e.g. discussed in issue #313 and PR #369). Contrary to the ideas tested in #313 it doesn't change the implementation of gsl::byte, but just adds

- a (deactivated) unittest that can reproduce the problem 
- a note to the readme, recommending to use the `-fno-strict-aliasing` flag on gcc and clang.

Please note, that it turns on optimizations for Unittests.


tests/CMakeLists.txt
```diff
@@ -24,9 +24,9 @@ else()
     CHECK_CXX_COMPILER_FLAG("-std=c++14" COMPILER_SUPPORTS_CXX14)
     CHECK_CXX_COMPILER_FLAG("-std=c++11" COMPILER_SUPPORTS_CXX11)
     if(COMPILER_SUPPORTS_CXX14)
-    	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14 -Wall -Wno-missing-braces")
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14 -O3 -Wall -Wno-missing-braces")
```
neilmacintosh: I would like to add `-fno-strict-aliasing` to these flags and then add a comment above these lines to explain why it's important to have both `-O3` and `-f-no-strict-aliasing` (i.e this way the tests can try to ensure there are not aliasing issues with the compiler when that combination of flags is applied).

tests/byte_tests.cpp
```diff
@@ -114,6 +114,24 @@ SUITE(byte_tests)
...
+        i = 10;
+        b = to_byte<5>();
+        return i;
+    }
+
+    TEST(aliasing)
+    {
+        int i{ 0 };
+        int res = modify_both(reinterpret_cast<byte&>(i), i);
+#ifdef GSL_CHECK_BYTE_ALIASING
```
neilmacintosh: I don't think you need to #ifdef out the test. Instead (as per my other comment) add all the flags to avoid the problem with GCC and Clang and then use this test to make sure they work as expected.

MikeGitb:
Done. If you are fine with the result, I can squash and rebase (or you can do it yourself). Btw. Using -O3 activates quite a few warnings in g++ when compiling the unit tests.

neilmacintosh:
Thanks @MikeGitb! I actually like the fact that we'll get those warnings...we need to get clean on them! Now we just need to go work out which ones we should fix and which we should suppress.

LYP951018:
Maybe __attribute__((__may_alias__)) save us all?

LYP951018:
Maybe `__attribute__((__may_alias__))` save us all?

MikeGitb:
@LYP951018: Sounds like a good idea to me - especially, as I got the impression that the std byte proposal will not make it into c++17

PR #393, opened 2016-10-11
Don't install tests

marc-kdab:
'make install' should install the GSL library (ie. the headers),
not the tests.

It still installs the UnitTest++ headers, but that will be more
complex to fix, as GSL just imports UnitTest++ as a git submodule,
and the install command propagates down to UnitTest++'s
CMakeLists.txt.


neilmacintosh:
@marc-kdab thanks for finding and fixing this! The change looks good to me.

Issue #391, opened 2016-10-11
string_span's string_length and strnlen fail for arm-none-eabi


Josaphat:
Including the use of `strnlen` is technically un-portable because it is not specified by the C or C++ standards, so using `-std=c++11` with gcc *should* fail IMO because it's providing non-standard functionality. But really I don't have an issue with the GSL using `strnlen` beyond the fact that it doesn't work on my platform (arm-none-eabi-gcc). The solution that was submitted in #334 is almost there but there are two short-comings:

1. Although my platform is not cygwin, `strnlen` is not declared in `string.h` if `std=c++..` is specified. arm-none-eabi-gcc doesn't allow me to use `-isystem` for c++ files because it implicitly wraps system includes with `extern "C" { }`, so gsl must be in my regular include path. I compile with `-Werror` so the implicit function declaration of `strnlen` causes my builds to fail.
2. `string_length` fails to compile cleanly because of an implicit narrowing conversion from `std::size_t` to `int` when `strnlen` is not used:
```
    span<const char> str_span{str, n}; // n is declared as std::size_t
```

To fix (1), is it reasonable to simply check if the symbol `strnlen` is defined?
```
#ifdef strnlen
#define GSL_PLATFORM_HAS_STRNLEN
#endif // strnlen
```

If this isn't a good solution then I would propose not defining `GSL_PLATFORM_HAS_STRNLEN` directly in the code at all and assuming that the platform does not have strnlen. If users of GSL require `string_length` to use `strnlen`, they could define `GSL_PLATFORM_HAS_STRNLEN` themselves.

For (2) I would simply add a checked narrowing conversion: `gsl::narrow<int>(n)`.

I would happily submit a PR which includes these changes.

MikeGitb:
Radical suggestion: If the GSL has to provide it's own implementation of strlen anyway, why not remove `std::strnlen` completely for now? That would reduces conditional compilation and I don't believe this will cause a significant performance problem.

Will c11's strnlen_s be part of c++17?

rianquinn:
@MikeGitb Correct me if I'm wrong, but I think the `cstring` headers are just namespaced wrappers around existing C functions, and thus `std::strnlen` doesn't exist either. If that's not the case then I would agree that it would be the better approach. 

[...]

PR #394, opened 2016-10-11
[#391] Remove strnlen and use consistent length type

Josaphat:
This implements the changes discussed in #391 . The end goal of the changes is to support arm-none-eabi though it should be more portable overall.


rianquinn:
+1

MikeGitb:
Looks good to me.

neilmacintosh:
@Josaphat thanks for raising the original issue and putting together the PR. This looks like a good change to me.

PR #398, opened 2016-10-18
Update narrow_cast to use std::forward

rianquinn:
Based on [F19](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#f19-for-forward-parameters-pass-by-tp-and-only-stdforward-the-parameter), I believe `gsl::narrow_cast` should be implemented using forward semantics.

### DIFF
--- a/gsl/gsl_util
+++ b/gsl/gsl_util
@@ -93,11 +93,19 @@ inline final_act<F> finally(F&& f) noexcept
 }
 
 // narrow_cast(): a searchable way to do narrowing casts of values
+#if _MSC_VER <= 1800
+template <class T, class U>
+inline constexpr T narrow_cast(U u) noexcept
+{
+    return static_cast<T>(u);
+}
+#else
 template <class T, class U>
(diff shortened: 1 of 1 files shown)


neilmacintosh:
@rianquinn if I scanned the AppVeyor output correctly, it looks like you may have to #ifdef out the change for MSVC 2013 due to a bug in that compiler.

rianquinn:
@neilmacintosh ha... yeah didn't see that coming. So long as your cool with that, I'll update the PR.

neilmacintosh:
@rianquinn yep, that's the simplest and best resolution for now. Thanks!

rianquinn:
No problem. We are using narrow_cast a lot (its really annoying how C++ auto converts to an int when doing arithmetic on an 8 or 16 bit value) and so I did a quick check to make sure it could be optimized out and realized there was a little room for improvement. Hopefully this get absorbed into the standard. 

Thanks a ton

neilmacintosh:
Thanks @rianquinn, this is a useful improvement!

PR #401, opened 2016-10-24
no conversion from T to not_null<T>

akrzemi1:


tests/notnull_tests.cpp
```diff
@@ -80,20 +80,20 @@ SUITE(NotNullTests)
...
-        not_null<Unrelated*> t = reinterpret_cast<Unrelated*>(p.get());
+        not_null<Unrelated*> t(reinterpret_cast<Unrelated*>(p.get()));
         CHECK((void*)p.get() == (void*)t.get());
     }
 
     TEST(TestNotNullAssignment)
     {
         int i = 12;
-        not_null<int*> p = &i; 
+        not_null<int*> p(&i);
```
rianquinn: If we are getting rid of the assignment operator of T, do we need the test at all? I believe there are already tests for construction.

gdr-at-ms: That line isn't exercising assignment though.

viboes: Do we need a make_not_null factory ?
```
void f(not_null<int*>);
f(make_not_null(&i))
```

Wondering if in the particular case of `not_null<T*>` it couldn't be copy-constructed and assigned from a `T&`.

```
f(i); // implicit conversion from T& to `not_null<T*>`
```

huangqinjin: @neilmacintosh `make_not_null()` is not a good idea, but what about `gsl::addressof()` just like the one in `std`?

annagrin:
please see discussion in #699. Closiing this since we decided to remove the explicit constructor and add strict_not_null type for practical purposes.

Issue #402, opened 2016-10-25
question: why no span.at()? Why no unique_ptr -> span?


rianquinn:
I'm working on some paging code, and I have to hold two lists, one is to the physical pages / directories being used by memory, and the other is to a list of my own PTE objects. The list in physical hardware has to be exactly 4K in size, so I cannot use `std::vector`, but at the same time my list of PTE objects can be. So to prevent the use of two `gsl::span` classes I use `std::vector` for my PTE objects, and I use `std::unique_ptr`/ `gsl::span` for the physical pages / directories (since the `[]` that `std::unique_ptr` provides is useless now that we have the Core Guidelines)

So a couple of questions:
- I have to use `.get()` to create the span. It would be nice to just pass a `std::unique_ptr` directly to span. It's a small thing but I feel like it's slowly eating away at my sole each time I see it. 
- Why is there no `.at()` for span? It's actually weird to me that there is a `[]` since we seem to be deprecating it's use in the STL but no `.at()` which is what we have to use in the STL. When looking at my code, half of it uses `.at()` and the other half uses `[]` (depending on whether your using the vector, or the span). The inconsistency is confusing.

neilmacintosh:
Hi @rianquinn! You make an interesting point about `unique_ptr`. I guess it currently "falls through the cracks" because it doesn't meet the standard library's criteria for contiguous container. I think that's definitely worth considering adding - support for simplified `unique_ptr` and `shared_ptr` construction.

Can we spin off a separate issue to track that feature request?

Regarding `at()`: there's already a free-function version of `at()` in the GSL. It should work for any container that is currently not range-checked on `[]` (array, `std::array`, `std::vector`) and I think it should be specialized for `span`. Would that be useful to you?

rianquinn:
I can toss in a PR for the `unique_ptr` and `shared_ptr` so that we can discuss it there. Adding the constructors for those should be pretty simple I think. 

[...]

PR #404, opened 2016-10-26
Add std::unique_ptr support to gsl::span

rianquinn:
This patch adds support for std::unique_ptr to the gsl::span
class instead of having to manually grab the pointer from
a std::unique_ptr.

It was suggestd that we also add support for std::shared_ptr,
but no array syntax exists for this class, and the element
types don't match. Until the following is accepted, I suggest
support for std::shared_ptr is left out:

http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4077.html

For reference, this is part of the following issue:
https://github.com/Microsoft/GSL/issues/402

### DIFF
--- a/gsl/span
+++ b/gsl/span
@@ -29,6 +29,7 @@
 #include <stdexcept>
 #include <type_traits>
 #include <utility>
+#include <memory>
 
 #ifdef _MSC_VER
 
@@ -385,6 +386,12 @@ public:
         : storage_(&arr[0], details::extent_type<N>())
     {
     }
+
+    template<class ArrayElementType = std::add_pointer<element_type>>
+    constexpr span(const std::unique_ptr<ArrayElementType>& ptr, index_type count) : storage_(ptr.get(), count) {}
+
+    constexpr span(const std::unique_ptr<ElementType>& ptr) : storage_(ptr.get(), ptr.get() ? 1 : 0) {}
+    constexpr span(const std::shared_ptr<ElementType>& ptr) : storage_(ptr.get(), ptr.get() ? 1 : 0) {}
 
     // NB: the SFINAE here uses .data() as a incomplete/imperfect proxy for the requirement
     // on Container to be a contiguous sequence container.
--- a/tests/span_tests.cpp
+++ b/tests/span_tests.cpp
@@ -483,6 +483,72 @@ SUITE(span_tests)
 #endif
     }
 
+    TEST(from_unique_pointer_construction)
+    {
+        {
+            auto ptr = std::make_unique<int>(4);
+
+            {
+                span<int> s{ptr};
+                CHECK(s.length() == 1 && s.data() == ptr.get());
+                CHECK(s[0] == 4);
+            }
+        }
+
+        {
+            auto ptr = std::unique_ptr<int>{nullptr};
+
+            {
+                span<int> s{ptr};
+                CHECK(s.length() == 0 && s.data() == nullptr);
+            }
+        }
+
+        {
+            auto arr = std::make_unique<int[]>(4);
+
+            for (auto i = 0; i < 4; i++)
+                arr[i] = i + 1;
+
+            {
+                span<int> s{arr, 4};
+                CHECK(s.length() == 4 && s.data() == arr.get());
+                CHECK(s[0] == 1 && s[1] == 2);
+            }
+        }
+
+        {
+            auto ptr = std::unique_ptr<int[]>{nullptr};
+
+            {
+                span<int> s{ptr, 0};
+                CHECK(s.length() == 0 && s.data() == nullptr);
+            }
+        }
+    }
+
+    TEST(from_shared_pointer_construction)
+    {
+        {
+            auto ptr = std::make_shared<int>(4);
+
+            {
+                span<int> s{ptr};
+                CHECK(s.length() == 1 && s.data() == ptr.get());
+                CHECK(s[0] == 4);
+            }
+        }
+
+        {
+            auto ptr = std::shared_ptr<int>{nullptr};
+
+            {
+                span<int> s{ptr};
+                CHECK(s.length() == 0 && s.data() == nullptr);
+            }
+        }
+    }
+
     TEST(from_container_constructor)
     {
         std::vector<int> v = {1, 2, 3};


neilmacintosh:
@rianquinn I like Jonathan Wakeley's paper, but I don't think there's any need to delay adding support for `shared_ptr`. A single element is still a contiguous sequence of one. So I would want to support creating a `span` easily from both `unique_ptr` and `shared-ptr`, regardless of whether they contain a single element or an array...given both are commonly used standard library "containers" (even though they don't meet the requirements for a contiguous container).

rianquinn:
That's fine. There a couple of issues (some are just things I have to learn about template programming). 
- unique_ptr and shared_ptr don't take the same `class T`. It just so happens that unique_ptr and span are in sync so it work, trying to do the same thing with shared_ptr didn't work as it was complaining about the fact that you would have to define the shared_ptr with T*, but span with T, so I just have to figure out how to strip the reference to make that work (I think... 😄)
- The unit test will have some funk code in it to initialize the shared_ptr with some known values. My original though to do that is to actually create a standard array so that I can initialize it, and then pass the pointer to shared_ptr to be managed from there, which in turn would be used to create the span. Obviously all of that could do away once that spec is implemented.

rianquinn:
@neilmacintosh I added support for std::shared_ptr. I actually had it backwards and needed to add a reference to unique_ptr so all around I think this is better. 

Sadly, I left my editor in a mode were it removed trailing spaces, which I guess is in issue in the code. If you want, I can re-submit this with the spaces added back in. Up to you.

rianquinn:
I fixed the whitespace issue, so we should be good to go.

rianquinn:
Any other comments on this PR?

neilmacintosh:
I'll take a look a little later on today.

tests/span_tests.cpp
```diff
@@ -483,6 +483,66 @@ SUITE(span_tests)
 #endif
     }
 
+    TEST(from_unique_pointer_construction)
+    {
+        {
+            auto arr = new int[4];
+            auto ptr = std::unique_ptr<int>(arr);
```
neilmacintosh: Doesn't this result in the wrong `delete` being called when `ptr` goes out of scope?

tests/span_tests.cpp
```diff
@@ -483,6 +483,66 @@ SUITE(span_tests)
...
+
+            for (auto i = 0; i < 4; i++)
+                arr[i] = i + 1;
+
+            {
+                span<int> s{arr, 4};
```
neilmacintosh: Is `arr` here supposed to be `ptr`? I assumed you were trying to test construction of the `span` from the `unique_ptr`.

tests/span_tests.cpp
```diff
@@ -483,6 +483,66 @@ SUITE(span_tests)
...
+                CHECK(s.length() == 4 && s.data() == ptr.get());
+                CHECK(s[0] == 1 && s[1] == 2);
+            }
+
+            {
+                span<int> s{std::unique_ptr<int>{nullptr}, 0};
```
neilmacintosh: While I get that it's probably ok in this exact test case, it seems like it might encourage a misunderstanding to construct a `span` from a temporary `unique_ptr`. For any initial value other than `nullptr` that would result in disaster ;-)

tests/span_tests.cpp
```diff
@@ -483,6 +483,66 @@ SUITE(span_tests)
...
+                arr[i] = i + 1;
+
+            {
+                span<int> s{arr, 4};
+                CHECK(s.length() == 4 && s.data() == arr.get());
+                CHECK(s[0] == 1 && s[1] == 2);
+            }
+
+            {
+                span<int> s{std::unique_ptr<int>{nullptr}, 0};
```
neilmacintosh: As per previous comment, container lifetime must always exceed view lifetime.

tests/span_tests.cpp
```diff
@@ -483,6 +483,66 @@ SUITE(span_tests)
...
+            arr[i] = i + 1;
+
+        {
+            span<int> s{arr, 4};
+            CHECK(s.length() == 4 && s.data() == ptr.get());
+            CHECK(s[0] == 1 && s[1] == 2);
+        }
+
+        {
+            span<int> s{std::unique_ptr<int>{nullptr}, 0};
```
neilmacintosh: I'm assuming this should be `shared_ptr`, not `unique_ptr`, but then it should also outlive the `span` created from it.

tests/span_tests.cpp
```diff
@@ -483,6 +483,66 @@ SUITE(span_tests)
...
+                span<int> s{std::unique_ptr<int>{nullptr}, 0};
+                CHECK(s.length() == 0 && s.data() == nullptr);
+            }
+        }
+    }
+
+    TEST(from_shared_pointer_construction)
+    {
+        auto arr = new int[4];
+        auto ptr = std::shared_ptr<int>(arr);
```
neilmacintosh: I'm pretty sure this will use the wrong `delete` unless you provide a custom deleter argument.

gsl/span
```diff
@@ -386,6 +387,11 @@ public:
     {
     }
 
+    template<class ArrayElementType = std::add_pointer<element_type>>
+    constexpr span(const std::unique_ptr<ArrayElementType>& ptr, index_type count) : storage_(ptr.get(), count) {}
```
neilmacintosh: My feeling is that we want two constructors here, to distinguish between the array and non-array holding versions of `unique_ptr'. The non-array version doesn't need a count, it can just be a convenient way to initialize a single-element `span` from that particular standard container.

gsl/span
```diff
@@ -386,6 +387,11 @@ public:
...
+
+    constexpr span(const std::shared_ptr<ElementType>& ptr, index_type count) : storage_(ptr.get(), count) {}
```
neilmacintosh: Until `shared_ptr` can officially take multiple elements, I'd be tempted to just have this constructor take a `shared_ptr` and construct a single-element `span` from it. If we hear that a lot of people go to the effort of putting dynamically-allocated arrays into `shared_ptr`, then we can always add the second overload.

rianquinn:
@neilmacintosh Thanks for the awesome feedback. Sorry it took so long to get the fixes in. I think this patch addresses all of your concerns, and I agree that this is a much cleaner approach as the non-array versions make a lot more sense this way.

neilmacintosh:
Thanks @rianquinn, this is a nice addition to the group of standard library containers supported by `span`.

PR #405, opened 2016-10-26
Fix compiler warning

rianquinn:
When building with GCC 5.4, the compiler spits out a warning about
the following:

warning: assuming signed overflow does not occur when assuming
that (X - c) > X is always false [-Wstrict-overflow]

This patch fixes this issue.

### DIFF
--- a/gsl/string_span
+++ b/gsl/string_span
@@ -554,7 +554,8 @@ public:
 
     constexpr string_span_type as_string_span() const noexcept
     {
-        return span_.first(span_.size() - 1);
+        auto sz = span_.size();
+        return span_.first(sz <= 0 ? 0 : sz - 1);
     }
 
     constexpr string_span_type ensure_z() const noexcept { return gsl::ensure_z(span_); }


neilmacintosh:
@rianquinn I am all for removing warnings in the GSL. However, before I accept the PR, I am hoping someone could explain to me what is actually wrong with the original code, because your change seems equivalent, just more characters.

rianquinn:
The compiler is making an assumption that could cause an overflow due to UB with signed integers. The code is actually not the same with respect to UB.

gsl/span
```diff
@@ -472,13 +472,13 @@ public:
 
     constexpr span<element_type, dynamic_extent> first(index_type count) const
     {
-        Expects(count >= 0 && count <= size());
+        Expects(count >= 0 && (count < size() || count == size()));
         return {data(), count};
```
gdr-at-ms: **if** you are going to call `size()` more than once, you should store the result in a local variable and reuse it.
Another occurrence below.

neilmacintosh:
@rianquinn Sorry to be dim, but could you explain the overflow potential in the `<=` comparison between two signed integers of the same size?

I ask, because I am not so sure that the compiler is pointing to a valid potential to cause overflow in that expression.

I have a suspicion that this warning is being issued because GCC is performing its analysis for that warning after optimization/inlining (and I'm no GCC expert) that the compiler is warning us about the interaction between potential overflow and whether or not it can optimize away that check.

gsl/span
rianquinn: fixed

rianquinn:
I am no expert on this so don't quote me, but my understanding is that GCC is trying to perform the optimization (X - c) > X, and when doing that, a potential overflow is possible, which is why it's tossing up the warning. From everything I have read, it's best to re-write the expression so that the compiler can optimize without the possibility of an overflow. Either that or you turn the optimization off (which cannot be done from a library's perspective).

neilmacintosh:
@rianquinn I agree with your reasoning and had the same suspicion...thank you for confirming it!

So my position holds. The expression *as written* in the source code is not at risk of causing overflow. I would rather not rewrite our expressions to become longer and more mental effort to process simply to pander to the intersection of any particular compiler's optimizer + warning analysis.

gdr-at-ms:
> my understanding is that GCC is trying to perform the optimization (X - c) > X, and when doing that, a potential overflow is possible, which is why it's tossing up the warning. From everything I have read, it's best to re-write the expression so that the compiler can optimize without the possibility of an overflow. Either that or you turn the optimization off (which cannot be done from a library's perspective).

Hmm, could you break this step in into smaller steps?  So, far it looks like a compiler bug; I would love to be proven wrong.

rianquinn:
@gdr-at-ms Actually I tried that first with `Expects(count <= size())` and that was enough to cause the warning. 

@neilmacintosh My understanding is that it will cause an overflow if the optimization is applied. We could just suppress the warning, but I think that would be ignoring the issue. If you look here:

https://travis-ci.org/Microsoft/GSL/jobs/170858906

You can see that the current builds on Travis CI are all showing the warning as well, so it's probably been there for a while.

rianquinn:
And I agree that it's a compiler bug with GCC and both VS and Clang do not have the same issue. I think GCC is trying to do something here it should not be, and then it spits out a warning tell you that it's about to make a mistake based on how the code is written.

neilmacintosh:
@rianquinn I am with @gdr-at-ms. If the original expression can cause signed integer overflow, then I have a suspicion it is a bug in the compiler....though I may be missing something. I am fairly sure nothing in the standard makes that expression susceptible to signed integer overflow.

neilmacintosh:
@rianquinn sorry...our replies "passed like ships in the night" ;-) If all three of us are suspicious about the validity of the warning then let's not rewrite the expression, and instead we should probably write a minimal repro case and ask GCC developers for more insight into what's going on.

rianquinn:
@neilmacintosh lol

I concur. I'll update the PR to ignore the warning for now and leave the code as is, and then I will file a bug with GCC to see if they can tell us why this is wrong 😄 .

rianquinn:
@neilmacintosh I ended having to disable strict overflow during compilation as GCC seemed to ignore the disable warning macros. I did some more research and since we aren't actually doing any arithmetic here, I agree that no overflow should be occurring so I'm not sure why it's complaining but the issue exists in GCC 5 and GCC 6 if you look at Travis CI.

rianquinn:
I filed a bug with GCC. If it does turn out to be a bug it's unlikely GCC 5 will be fixed, and thus the patch will likely be needed for a while until the fix is available. 

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78130

rianquinn:
Based on what they found, it is an issue with the code. Somewhere, there is a "-1" that is causing the possibility of an overflow. I'll see if I can track it down, but the optimization did actually find a bug in the code.

octoploid:
Here: https://github.com/Microsoft/GSL/blob/master/gsl/string_span#L557

rianquinn:
Replacing the code identified with the following does in fact fix the issue. Would you like me to commit this change instead?

```cpp
auto sz = span_.size();
return span_.first(sz <= 0 ? 0 : sz - 1);
```

neilmacintosh:
Thanks @rianquinn. So that warning + optimization is a good defect-finding tool...as long as you have GCC developers available to help you find the offending line of code that exists somewhere ;-)

Yes, that fix looks good!

rianquinn:
Yeah the debug output that GCC spits out is nuts so I'm glad they were able to help. I'll update the PR to address this.

neilmacintosh:
@rianquinn thanks very much for taking the time to dig into this issue so thoroughly and fix a bug in the library!

neilmacintosh:
And many thanks also to @octoploid for helping us pinpoint the exact location of the bug!

PR #407, opened 2016-10-27
Cleanup additional warnings

rianquinn:
When turning on the following flags, several additional warnings
were generated, which have been cleaned up in this patch. The
flags included:

-Wextra
-Wpedantic
-Wconversion
-Wsign-conversion
-Wctor-dtor-privacy
-Wshadow
-Wnon-virtual-dtor
-Wold-style-cast
-Wcast-align
-Woverloaded-virtual


gsl/span
```diff
@@ -630,7 +630,7 @@ template <class ElementType, std::ptrdiff_t Extent>
 constexpr ElementType& at(const span<ElementType, Extent>& s, size_t index)
 {
     // No bounds checking here because it is done in span::operator[] called below
-    return s[index];
+    return s[narrow_cast<typename span<ElementType, Extent>::index_type>(index)];
```
rianquinn: This is the only one that I was not sure about what to do as we could use `gsl::narrow` or `gsl::narrow_cast`. I used `gsl::narrow_cast` as I know a lot of people have brought up issues with performance so I figured that would be the likely choice.

tests/at_tests.cpp
```diff
@@ -26,39 +26,39 @@ SUITE(at_tests)
 {
     TEST(static_array)
     {
-        int a[] = { 1, 2, 3, 4 };
+        size_t a[] = { 1, 2, 3, 4 };
 
-        for (int i = 0; i < 4; ++i)
+        for (size_t i = 0; i < 4; ++i)
             CHECK(at(a, i) == i+1);
```
gdr-at-ms: Why `size_t`?

rianquinn: `at` takes a size_t so attempting to put an int in it will result in a signed conversion warning.

gdr-at-ms: Hmm, I missed that.  Why is `at()` taking a different type than `[]`?

rianquinn: You tell me 😄 

That at if the one that's been there for a while (this is testing `gsl::at` and not the new `gsl::span::at`).

gdr-at-ms: Yes; it was question to "us".  It is interesting the compiler isn't warning about `i < 4`, presumably because `4` is small enough?

rianquinn: Good question. I agree that should be `4U` and not `4` but yeah, it's not mad about that. I didn't try with Clang so it might just be GCC being nice.

Do you want me to change that?

rianquinn: I'll test with Clang, but I'm almost positive it will complain about the same thing as I have had to make a lot of changes like this to support Clang Tidy with these warnings enabled

rianquinn: Yes, clang has the same warning (and some others I might add, but we can address those in a different PR)

```
/home/user/gsl/tests/at_tests.cpp:62:25: warning: implicit conversion changes signedness: 'int' to 'size_t' (aka 'unsigned long')
      [-Wsign-conversion]
            CHECK(at(a, i) == i+1);
```

neilmacintosh: I'm with @gdr-at-ms ... why does `at()` take `size_t`? It should not. It should take `ptrdiff_t`. If that's my fault, it's probably because we never updated it after the original prototype back in the mists of the past. But that should change, rather than be worked around.

rianquinn: @neilmacintosh ok I can update the PR to change this.

rianquinn:
I'd like to work on the following PR first, before we continue with this one. 
https://github.com/Microsoft/GSL/pull/408

Once we address the issue with `gsl::at`, I can take a stab at this again.

gsl/span
neilmacintosh: I think you're fine to use `gsl::narrow_cast()` here because you know that `s` will validate the bounds anyway. So I like it as you have it.

rianquinn: Since the other PR changes the index type to std::ptrdiff_t, there is no need for this cast anymore. Once the other PR is merged, I will clean this one up so that only the fixes that are needed there.

rianquinn:
@neilmacintosh I updated this PR now that the `size_t` issue has been addressed with `at()`. I also enabled the rest of the C++ warnings that GCC provides, so I think that is all of them. With this PR, people should be able to enable `-Werror` and any of the GCC warnings.

neilmacintosh:
@rianquinn thanks very much for doing this useful cleanup! It looks great to me.

PR #408, opened 2016-10-27
Fix gsl::at's use of size_t

rianquinn:
gsl::at should use std::ptrdiff_t instead of size_t. This patch
fixes that, and adds missing unit tests

### DIFF
--- a/gsl/gsl_util
+++ b/gsl/gsl_util
@@ -135,30 +135,30 @@ inline T narrow(U u)
 // at() - Bounds-checked way of accessing static arrays, std::array, std::vector
 //
 template <class T, size_t N>
-constexpr T& at(T (&arr)[N], size_t index)
+constexpr T& at(T (&arr)[N], std::ptrdiff_t index)
 {
-    Expects(index < N);
-    return arr[index];
+    Expects(index >= 0 && index < narrow_cast<std::ptrdiff_t>(N));
+    return arr[static_cast<size_t>(index)];
 }
 
 template <class T, size_t N>
-constexpr T& at(std::array<T, N>& arr, size_t index)
+constexpr T& at(std::array<T, N>& arr, std::ptrdiff_t index)
 {
-    Expects(index < N);
-    return arr[index];
+    Expects(index >= 0 && index < narrow_cast<std::ptrdiff_t>(N));
+    return arr[static_cast<size_t>(index)];
 }
 
 template <class Cont>
-constexpr typename Cont::value_type& at(Cont& cont, size_t index)
+constexpr typename Cont::value_type& at(Cont& cont, std::ptrdiff_t index)
 {
-    Expects(index < cont.size());
-    return cont[index];
+    Expects(index >= 0 && index < narrow_cast<std::ptrdiff_t>(cont.size()));
+    return cont[static_cast<typename Cont::size_type>(index)];
 }
 
 template <class T>
-constexpr const T& at(std::initializer_list<T> cont, size_t index)
+constexpr const T& at(std::initializer_list<T> cont, std::ptrdiff_t index)
 {
-    Expects(index < cont.size());
+    Expects(index >= 0 && index < narrow_cast<std::ptrdiff_t>(cont.size()));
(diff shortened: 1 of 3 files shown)


gsl/gsl_util
```diff
@@ -135,30 +135,30 @@ inline T narrow(U u)
 // at() - Bounds-checked way of accessing static arrays, std::array, std::vector
 //
 template <class T, size_t N>
-constexpr T& at(T (&arr)[N], size_t index)
+constexpr T& at(T (&arr)[N], std::ptrdiff_t index)
 {
-    Expects(index < N);
-    return arr[index];
+    Expects(index >= 0 && index < narrow_cast<decltype(index)>(N));
+    return arr[static_cast<size_t>(index)];
```
gdr-at-ms: Why `decltype(index)`?  Why is the cast even needed?

rianquinn: Now that I look at it, I probably should have used decltype(N) for size_t instead of using size_t directly for the same reasons

rianquinn:
I fixed the use of `size_t` and used `decltype(N)` and `typename Cont::size_type` in it's place.

gsl/gsl_util
gdr-at-ms: > index is already defined so if it gets changed again in the future, you don't have to worry about the cast (i.e. you only have to change one). 

I suspected that but wasn't sure.
I actually prefer the simpler `std::ptrdff_t` as it makes obvious whether the cast is meaningful. 
I don't expect anyone to change back to `std::size_t`, but if someone does for some reason, the `decltype` version won't call attention -- because "you have to change only one place" and "it is obviously correct".  This is one of the very few places where I see "redundancy" as feature.

gsl/gsl_util
```diff
@@ -135,30 +135,30 @@ inline T narrow(U u)
...
 
 template <class Cont>
-constexpr typename Cont::value_type& at(Cont& cont, size_t index)
+constexpr typename Cont::value_type& at(Cont& cont, std::ptrdiff_t index)
 {
-    Expects(index < cont.size());
-    return cont[index];
+    Expects(index >= 0 && index < narrow_cast<decltype(index)>(cont.size()));
+    return cont[static_cast<typename Cont::size_type>(index)];
 }
```
gdr-at-ms: +1.

gsl/gsl_util
```diff
@@ -135,30 +135,30 @@ inline T narrow(U u)
...
 
 template <class T, size_t N>
-constexpr T& at(std::array<T, N>& arr, size_t index)
+constexpr T& at(std::array<T, N>& arr, std::ptrdiff_t index)
 {
-    Expects(index < N);
-    return arr[index];
+    Expects(index >= 0 && index < narrow_cast<decltype(index)>(N));
+    return arr[static_cast<decltype(N)>(index)];
 }
```
gdr-at-ms: I still have mixed feelings about these :-)

rianquinn: 😄 

I can certainly change it back, but we use aliases in classes for all of our types for the same reasons (well there are many reasons for it, but this is one of them). Since this is a standalone function you would have to make the aliases global which is not a good idea, so I used decltype() as the next best thing for consistency. 

I have no issue changing it back (it's an easy change), but IMO it seems weird to redefine the type twice. 

I leave it up to you.

neilmacintosh: I think it would just read easier (esp. in line 141) if it was: `static_cast<size_t>(index)` and 'narrow_cast<std::ptrdiff_t>(N)'. And to be honest, I don't think the `decltype`'s will end up saving us a lot in terms of maintenance.

Sorry to be painful, but can we just go with the boring old named types?

neilmacintosh:
@rianquinn Thanks so much for putting together this PR. It tidies up an oversight and does so complete with some additional tests, I love it! I think there's just the style issue that @gdr-at-ms identified and then this looks good to go.

rianquinn:
Fixed

neilmacintosh:
Terrific job @rianquinn, thanks for taking the time to make this improvement!

Issue #411, opened 2016-11-01
make_span(...)?


rianquinn:
Is there a reason we don't have a make_span() helper to remove the need for declaring the template type? It would be similar to gsl::finally creating a final_act. It should not be needed once C++17 comes out, but it's going to be a couple of years before C++17 is out and usable by most of the people using this library so having a helper for now would certainly clean up the code a bit. 

Something like (not tested, or compiled)
```cpp
template <class T, class... Args>
inline auto make_span() {
    return span<T>{forward<Args>(args)...};
}
```

There would probably need to be a couple of these to handle the different types of spans, but it should be pretty simple to setup, and then the unit tests could be converted to use the `make_span` version instead to not only simplify them, but ensure the added API is complete

MikeGitb:
What would be the difference compared to the removed  `as_span` function we discussed in  #380?

rianquinn:
huh... I just assumed that as_span was something else, so I guess not much, other than I think `make_span` is more inline with the STL. 

I just added one to our downstream to see how it would work and it seems to work fine (the implementation is a little different than what I put above). It really helps to remove needless verbosity, as I use a lot of type aliases, so my type names can get pretty long.

MikeGitb:
Well - as I wrote in the other issue I also think it would be handy. If I understood neilmacintosh correctly he would probably accept a PR that provides such a function in a separate header.

rianquinn:
I saw that too (I think he said gsl_util), so I can work on upstreaming my changes once I get a chance. I'll likely update the unit tests to use the helpers instead of directly using span, as it tests both at the same time as a result. Also, not all of them will work, for example, the constructors that create a null span won't work since you don't have a type in the function definition to deduce, but for the ones we can do, I will write them, and update the tests as needed.

neilmacintosh:
@rianquinn The suggestion seems pretty reasonable to me. There is certainly plenty of precedent in the standard library for `make_` names and I do understand the convenience of the syntax. However, as I think I observed in a previous issue, the fancy new template constructor deduction feature (not the technical term) in C++17 may reduce the syntactic need for `make_span`.

Nonetheless, I think the addition would be a good one to help people out while we wait for C++17 to be finalized and then propagate through the world.

MikeGitb:
Stupid question: has anyone investigated, whether that c++17 feature would actually work for gsl::span's constructors, as they are overloaded, templated and use somewhat complicated enable_if expressions?

[...]

Issue #380, opened 2016-09-30
I did not find gsl::as_span.


efinegold:
I did not find funtion gsl::as_span. ( ...but I am s new user). Did you remove this function from the library?

neilmacintosh:
Hi @efinegold! The function still exists for `multi_span`, but as it is not part of the interface of `span` that is currently proposed for standardization, I left it out of `span`. In theory, there should be enough constructors to achieve what `as_span()` used to do.

MikeGitb:
> In theory, there should be enough constructors to achieve what as_span() used to do. 

Well, the main advantage of `as_span()` would be the automatic type deduction, but with c++17 that might no longer be an issue (haven't yet looked up the type deduction rules for templated constructors of templated classes).

rianquinn:
I think that applies to a lot of things in the GSL, `gsl::finally` being a great example of that.

neilmacintosh:
@MikeGitb, @rianquinn you are both correct on both counts. If there are painful spots, then I think it's best to wait for the template constructor type deduction changes in C++17 which should resolve most (all?) of them.

With that said, I could imagine adding `as_span()` back into a separate header (maybe `gsl_util`) to help people without C++17 support, or who face a persistent problem now and don't want to wait.

efinegold:
Thanks. Yes. It worked in some cases:

void reportInt(gsl::span<int> data)
{
	for (auto const item : data) std::cout << item << ' ';
	std::cout << std::endl;
}

int main()
{
int arr[] = { 100, 200, 300, 400, 500 };

MikeGitb:
@neilmacintosh: Well, it would certainly be handy to have them but imho nothing critical.

neilmacintosh:
Ok. Then I'll close this out for now. Feel free to reopen or open a new issue if someone hits something that is particularly annoying.

efinegold:
Sounds good. Thanks.
Ed.

PR #413, opened 2016-11-02
Add make_span()

rianquinn:
This patch provides a make_span function to simplify the
creation of spans until C++17 is available. In addtion
this patch updates the unit tests to includes tests that
verify this new functionality.

### DIFF
--- a/gsl/span
+++ b/gsl/span
@@ -632,6 +632,45 @@ as_writeable_bytes(span<ElementType, Extent> s) noexcept
     return {reinterpret_cast<byte*>(s.data()), s.size_bytes()};
 }
 
+//
+// make_span() - Utility functions for creating spans
+//
+template <class ElementType>
+span<ElementType>
+make_span(ElementType* ptr, typename span<ElementType>::index_type count)
+{ return span<ElementType>(ptr, count); }
+
+template <class ElementType>
+span<ElementType>
+make_span(ElementType* firstElem, ElementType* lastElem)
+{ return span<ElementType>(firstElem, lastElem); }
+
+template <class ElementType, size_t N>
+span<ElementType>
+make_span(ElementType (&arr)[N])
+{ return span<ElementType>(arr); }
+
+template <class Container>
+span<typename Container::value_type>
+make_span(Container &cont)
+{ return span<typename Container::value_type>(cont); }
+
+template <class Container>
+span<const typename Container::value_type>
+make_span(const Container &cont)
+{ return span<const typename Container::value_type>(cont); }
+
+template <class Ptr>
+span<typename Ptr::element_type>
+make_span(Ptr& cont, std::ptrdiff_t count)
+{ return span<typename Ptr::element_type>(cont, count); }
+
+template <class Ptr>
+span<typename Ptr::element_type>
+make_span(Ptr& cont)
+{ return span<typename Ptr::element_type>(cont); }
+
+
 // Specialization of gsl::at for span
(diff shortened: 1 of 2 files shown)


rianquinn:
Addresses the following issue:
https://github.com/Microsoft/GSL/issues/411

As well as:
https://github.com/Microsoft/GSL/issues/380

There is a circular reference issue if this code is added to gsl_util with narrow_cast so for now I left these functions in span's header (similar to std::memory which has the definition / implementation of unique_ptr with it's make functions right underneath).

tests/span_tests.cpp
```diff
@@ -204,6 +204,24 @@ SUITE(span_tests)
...
+
+        {
+            auto s = make_span(&arr[0], 2);
+            CHECK(s.length() == 2 && s.data() == &arr[0]);
+            CHECK(s[0] == 1 && s[1] == 2);
+        }
+
+        {
+            int* p = nullptr;
+            auto s = make_span(p, static_cast<span<int>::index_type>(0));
```
rianquinn: possibly, but I'm doing the same thing the non make_span test is doing to be consistent. To create the tests, I just cut / past all of the tests that apply and changed them to make_span to be sure that we hit everything

gsl/span
```diff
@@ -625,6 +625,34 @@ as_writeable_bytes(span<ElementType, Extent> s) noexcept
     return {reinterpret_cast<byte*>(s.data()), s.size_bytes()};
 }
 
+//
+// make_span() - Utility functions for creating spans
+//
+template <class ElementType>
+span<ElementType>
+make_span(ElementType *ptr, typename span<ElementType>::index_type &&count)
```
neilmacintosh: The `&&` seems unnecessary given count is a ptrdiff_t.

neilmacintosh: Nit: we put `*` next to the type name in GSL code.

gsl/span
```diff
@@ -625,6 +625,34 @@ as_writeable_bytes(span<ElementType, Extent> s) noexcept
...
+{ return span<ElementType>(ptr, std::forward<typename span<ElementType>::index_type>(count)); }
+
+template <class ElementType>
+span<ElementType>
+make_span(ElementType *firstElem, ElementType *lastElem)
+{ return span<ElementType>(firstElem, lastElem); }
+
+template <class ElementType, size_t N>
+span<ElementType>
+make_span(ElementType (&arr)[N])
```
neilmacintosh: no std::array overload?

gsl/span
```diff
@@ -625,6 +625,34 @@ as_writeable_bytes(span<ElementType, Extent> s) noexcept
...
+template <class Container>
+span<typename Container::value_type>
+make_span(Container &cont)
+{ return span<typename Container::value_type>(cont); }
+
+template <class Container>
+span<const typename Container::value_type>
+make_span(const Container &cont)
+{ return span<const typename Container::value_type>(cont); }
+
```
neilmacintosh: Do you think we should add overloads for `unique_ptr` and `shared_ptr` now we've added those ctors?

rianquinn: Yes, I will added these as part of the update.

rianquinn:
@neilmacintosh make_span has been updated to address you concerns. Let me know if there is anything else you would like me to do.

neilmacintosh:
Thanks very much for putting this togother @rianquinn. It looks good to me and I hope people find these convenience functions useful!

PR #414, opened 2016-11-03
remove noexcept on functions that use Expects macro

vladon:
remove noexcept on functions that use Expects macro (because Excepts actually can throw which can lead to std::terminate)

### DIFF
--- a/gsl/span
+++ b/gsl/span
@@ -177,7 +177,7 @@ namespace details
             return &((*span_)[index_]);
         }
 
-        constexpr span_iterator& operator++() noexcept
+        constexpr span_iterator& operator++()
         {
             Expects(span_ && index_ >= 0 && index_ < span_->length());
             ++index_;
@@ -191,7 +191,7 @@ namespace details
             return ret;
         }
 
-        constexpr span_iterator& operator--() noexcept
+        constexpr span_iterator& operator--()
         {
             Expects(span_ && index_ > 0 && index_ <= span_->length());
             --index_;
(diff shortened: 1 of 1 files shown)


viboes:
Why?

vladon:
@viboes Because functions marked `noexcept` cannot throw, but macro `Expects` which used in them, actually throws.

This can lead to `std::terminate` in user program, because you cannot catch that exception in `try..catch` block.

rianquinn:
+1

neilmacintosh:
The reasoning behind this PR is incorrect. However, the changes in it seem valid for different reasons ;-)

First I want to be clear: simply having an `Expects()` is not a statement about whether or not a function should be `noexcept`.

The `Expects()` macro is simply used to enforce a precondition on a function. The fact that it *may* throw an exception in a certain configuration of the library is purely an implementation detail. Whereas `noexcept` is part of the interface of the library. I wouldn't want to change it at compile-time, based on a particular enforcement configuration. (Yes, there is the obscure case of unit-testing...where it can be a convenience to make `noexcept` go away sometimes, but that is a different issue).

For `gsl::span` and friends, we have generally followed the standard library's approach of using `noexcept` where a function cannot fail. Note that I'm not endorsing this as the correct approach for applying `noexcept` in general. Rather, we want to have `span` included in the standard library, so by following their guideline we smooth that process. :-)

I think this PR points to a few places where functions could reasonably be expected to fail due to bad inputs (such as `operator+='). So for that reason, I'm happy to take it.

viboes:
Curiously Reasoning Template Pattern.

neilmacintosh:
I'm not sure I'm smart enough to understand the joke.

viboes:
it was a bad joke.

MikeGitb:
@neilmacintosh: Will the behavior of span's functions be standardized with regard to if and how they fail if their preconditions are violated?

PR #418, opened 2016-11-08
Maybe PR: basic_string_span: test for std::array; ctor for std::array not called

martinmoene:
This PR adds a test for the `std::array` constructor of `basic_string_span`.

The test shows that constructing a `basic_string_span` from a `std::array` (or `std::vector`) that includes a terminating `'\0'` does include that `'\0'` in the `basic_string_span` . Note that for `std::array` this behavior differs from what `basic_string_span` proposal [p0123r2](http://wg21.link/p0123) suggests.

In *gsl lite* I'm [leaning to follow the proposal](https://github.com/martinmoene/gsl-lite/blob/5c64f98e1ef394abadc8b1186ae72ed8f5c42367/include/gsl/gsl-lite.h#L1566) and use `remove_z( arr )` for the `std::array` case.

This PR also adds `std::terminate()` (!) to `basic_string_span`'s constructor for `std::array` to show it is not called. Instead the `Container` version takes precedence as std::array is not excluded for this constructor as specified in the `basic_string_span` proposal [p0123r2](http://wg21.link/p0123) (see *gsl lite* [here](https://github.com/martinmoene/gsl-lite/blob/5c64f98e1ef394abadc8b1186ae72ed8f5c42367/include/gsl/gsl-lite.h#L1583)).

### DIFF
--- a/gsl/string_span
+++ b/gsl/string_span
@@ -326,6 +326,8 @@ public:
     template <size_t N, class ArrayElementType = std::remove_const_t<element_type>>
     constexpr basic_string_span(const std::array<ArrayElementType, N>& arr) noexcept : span_(arr)
     {
+        // never called, Container version below takes precedence
+        std::terminate();
     }
 
     // Container signature should work for basic_string after C++17 version exists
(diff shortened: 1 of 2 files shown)


JordanMaples:
Maintainer's call: `std::terminate` should not be called regardless of whether or not that constructor is used or not, even then it should probably be an expects call. Adding the `std::array` tests to the `string_span` might be a nice addition to the tests if you'd like to add them in another PR. Thanks and sorry about the delay.

Issue #333, opened 2016-08-18
Invalid call to narrow_cast with unpacked pack parameter


viboes:
The current definition

```
    template <typename FirstIndex, typename... OtherIndices>
    constexpr reference operator()(FirstIndex index, OtherIndices... indices)
    {
        index_type idx = {narrow_cast<std::ptrdiff_t>(index),
                          narrow_cast<std::ptrdiff_t>(indices...)}; // (*)
        return this->operator[](idx);
    }
```

should in my opinion be replaced the line (*)                           

```
                          narrow_cast<std::ptrdiff_t>(indices)...};
```

neilmacintosh:
Fixed by PR #419.

PR #419, opened 2016-11-12
Fix issue #333 by moving a parenthesis.

neilmacintosh:


gsl/multi_span
```diff
@@ -1488,7 +1488,7 @@ public:
     constexpr reference operator()(FirstIndex index, OtherIndices... indices)
     {
         index_type idx = {narrow_cast<std::ptrdiff_t>(index),
-                          narrow_cast<std::ptrdiff_t>(indices...)};
+                          narrow_cast<std::ptrdiff_t>(indices)...};
```
viboes: Should some test be added to show that there was an error?

neilmacintosh: @viboes Yes, you're a little ahead of me. I was just lazily using the PR to force the CI systems to go build with all the different compilers for me, in case I'd created some silly syntax error. ;-)

PR #422, opened 2016-11-19
gsl::copy returns remaining range

MikeGitb:
As originally planned for #344, `gsl::copy` now returns the remaining destination range. 

If both source and destination span have a static extent, the result is a new span of static extent. Otherwise a span of dynamic extent is returned.

Additionally, the unit tests are now more fine granular than before. Due to the resulting indentation change I'd recommend viewing the diff of  `tests/algorithm_tests.cpp`  with the `-w` flag.

### DIFF
--- a/include/gsl/gsl_algorithm
+++ b/include/gsl/gsl_algorithm
@@ -37,9 +37,21 @@
 namespace gsl
 {
 
-template <class SrcElementType, std::ptrdiff_t SrcExtent,
+namespace details_copy
+{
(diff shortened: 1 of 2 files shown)


MikeGitb:
As this PR needs some overhaul before it can be merged: Is there any interest in merging this in the near future?

ewoudvc:
I'm a bit late to the this party.
Would be handy that gsl::copy has a return similar to std::copy_n, no?

PR #425, opened 2016-11-27
Add value_type to span

Novakov:
Currently I'm working on project which involves a lot of `span`s and mocking via Google Mock. Unfortunately a lot of standard matchers requires `value_type` type definition inside container which `gsl::span` lacks. 

This pull request add `value_type` type definition inside `gsl::span`

### DIFF
--- a/include/gsl/span
+++ b/include/gsl/span
@@ -133,7 +133,7 @@ namespace details
         using element_type_ = typename Span::element_type;
     public:
         using iterator_category = std::random_access_iterator_tag;
-        using value_type = std::remove_const_t<element_type_>;
+        using value_type = std::remove_cv_t<element_type_>;
         using difference_type = typename Span::index_type;
 
         using reference =
@@ -337,7 +337,7 @@ class span
 public:
     // constants and types
     using element_type = ElementType;
-    using value_type = ElementType;    
+    using value_type = std::remove_cv_t<ElementType>;    
     using index_type = std::ptrdiff_t;
     using pointer = element_type*;
     using reference = element_type&;
--- a/include/gsl/span
+++ b/include/gsl/span
@@ -337,6 +337,7 @@ class span
 public:
     // constants and types
     using element_type = ElementType;
+    using value_type = ElementType;    
     using index_type = std::ptrdiff_t;
     using pointer = element_type*;
     using reference = element_type&;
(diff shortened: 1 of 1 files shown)


gsl/span
```diff
@@ -347,6 +347,7 @@ class span
 public:
     // constants and types
     using element_type = ElementType;
+    using value_type = ElementType;
```
CaseyCarter: This should be `using value_type = std::remove_cv_t<ElementType>;` - in standard library parlance, value types are never cv-qualified.

Novakov: Good point - fixed :)

Novakov: I'm not so sure about this. First, adding `std::remove_cv_t` causes a lot of compilation errors, than I look how `value_type` is implemented in `vector` - it's directly type of the element (cv is not removed).

CaseyCarter: While you're at it, the `remove_const_t` in the definition of `span_iterator::value_type` should be `remove_cv_t` as well. There's no reason to go out of our way to support `volatile` types in the library, but when it's this easy there's no reason not to do so.

> adding `std::remove_cv_t` causes a lot of compilation errors

Such as?

> `value_type` is implemented in `vector` - it's directly type of the element (cv is not removed)

It's not an issue for containers: they do not support cv-qualified element types (http://stackoverflow.com/questions/6954906/does-c11-allow-vectorconst-t). Were they to do so, I assure you that `vector::value_type` would strip cv-qualifiers from the element type. (Well, it would certainly strip `const`. LWG is very fond of pretending that `volatile` doesn't exist ;))

Novakov: Please ignore everything I said ;) I was blind, changed wrong place and didn't notice that. `remove_cv_t` works like charm :)

neilmacintosh:
@Novakov I'm happy to take this PR, as it will help people writing generic code, and make `span` consistent with standard library types. However, I'm curious...before I push the "merge" button. Does the fact that `value_type` is not cv-qualified, but that a `span` can contain `const T` as opposed to `T` impact your usage scenario at all? (Note that `element_type` should still give you the original cv-qualified `T`.)

Novakov:
HI @neilmacintosh, we are using this fork of GSL for some time already (also with cv change) and didn't notice any problems with it :) Even `span<const..> works nicely :)

neilmacintosh:
Terrific, thanks for the information @Novakov. This looks good to me, thanks for the contribution!

PR #426, opened 2016-11-29
Fix a signed/unsigned comparison warning.

daviddoria:
This warning was seen with gcc 6.2.0


neilmacintosh:
Thanks for the PR @daviddoria!

For the future, it would be preferable to use `narrow_cast` to plain `static_cast` here. But given this is test code rather than production code, I'm happy to take the fix as-is to suppress the warning and let it get cleaned up in the future.

PR #427, opened 2016-12-05
[RFC] Add algorithms

MikeGitb:
Add wrapper functions for some STL algorithms. So far I only implemented a few and unittests are still missing, but as there is a lot of copy & paste involved, I want to first get some feedback on the general style, before replicating it.

### DIFF
--- a/gsl/gsl_algorithm
+++ b/gsl/gsl_algorithm
@@ -83,7 +83,35 @@ auto find_if_not(const span<SrcElementType, SrcExtent>& rng, UnaryPredicate p) -
     return details_algo::toIt(rng, ptr);
 }
 
+// lower_bound
+template <class SrcElementType, std::ptrdiff_t SrcExtent, class T>
+auto lower_bound(const span<SrcElementType, SrcExtent>& rng, const T& value) -> decltype(rng.begin())
+{
+    auto* ptr = std::lower_bound(rng.data(), rng.data() + rng.size(), value);
+    return details_algo::toIt(rng, ptr);
+}
 
+template <class SrcElementType, std::ptrdiff_t SrcExtent, class T, class Compare>
+auto lower_bound(const span<SrcElementType, SrcExtent>& rng, const T& value, Compare comp) -> decltype(rng.begin())
+{
+    auto* ptr = std::lower_bound(rng.data(), rng.data() + rng.size(), value, comp);
+    return details_algo::toIt(rng, ptr);
+}
+
+// upper_bound
+template <class SrcElementType, std::ptrdiff_t SrcExtent, class T>
+auto upper_bound(const span<SrcElementType, SrcExtent>& rng, const T& value) -> decltype(rng.begin())
+{
+    auto* ptr = std::upper_bound(rng.data(), rng.data() + rng.size(), value);
+    return details_algo::toIt(rng, ptr);
+}
+
+template <class SrcElementType, std::ptrdiff_t SrcExtent, class T, class Compare>
+auto upper_bound(const span<SrcElementType, SrcExtent>& rng, const T& value, Compare comp) -> decltype(rng.begin())
+{
+    auto* ptr = std::upper_bound(rng.data(), rng.data() + rng.size(), value, comp);
+    return details_algo::toIt(rng, ptr);
+}
 
 } // namespace gsl
 
--- a/gsl/gsl_algorithm
+++ b/gsl/gsl_algorithm
@@ -113,6 +113,19 @@ auto upper_bound(const span<SrcElementType, SrcExtent>& rng, const T& value, Com
     return details_algo::toIt(rng, ptr);
 }
 
+// binary_search
+template <class SrcElementType, std::ptrdiff_t SrcExtent, class T>
+bool binary_search(span<SrcElementType, SrcExtent> rng, const T& value)
+{
+    return std::binary_search(rng.data(), rng.data() + rng.size(), value);
+}
+
+template <class SrcElementType, std::ptrdiff_t SrcExtent, class T, class Compare>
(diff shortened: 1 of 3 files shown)


MikeGitb:
Forgot that VS2013 doesn't support auto return type deduction - will fix soon (If supporting VS2013 is still a concern).

neilmacintosh:
@MikeGitb I'm afraid that supporting VS2013 is still a requirement for the library until VS2017 is released.

MikeGitb:
@neilmacintosh: Ok, I fixed the return type declaration and also an interesting snag: In order to return a valid iterator, the functions have to take the span parameter by reference. Obvious in hindsight, but it took me some time to realize it. 
I'm mentioning this because the cppcorechecker - which I downloaded via nuget (14.0 from June) - didn't warn me about the iterator iinvalidation. Would the version in VS2017 do? Is there something that should/could be changed inside the gsl?

MikeGitb:
I've probably some time over the weekend and would like to implement more wrappers for standard algorithms. Should I go on with the same scheme presented so far or are there any concerns?

MikeGitb:
One question in particular: Due to the fact that returning an `gsl::span::iterator` that was created from a local span is not possible, some algorithms that return an iterator take the span by const ref, while the rest takes it by value. Should I keep it this way or should all algorithms use `const span&`  consistently?

neilmacintosh:
@MikeGitb I would think about the design for these wrappers the following way. You want to follow the Ranges approach wherever possible, as that is a close conceptual cousin to `span`. My understanding of the Ranges approach to algorithms, is that they take ranges and return ranges (very rarely do they return iterators). So I suspect you could take the `span` parameters by-value here and then return a `span` by-value. 

Yes, this means the callers don't look like callers to today's STL algorithms. But neither do callers that use the Ranges library ;-)

MikeGitb:
@neilmacintosh: That is complete news to me. As far as I can tell from the ranges TS proposal [n4610](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/n4610.pdf) every algorithm that returns an iterator in the current standard library would - under the TS - also either return an iterator or a tuple of iterators into different ranges. The latter is e.g. the case with the ranges TS version of `std::copy`, where one iterator points to the end of the input range and one iterator points to the new end of the output range (which unfortunately conflicts with my PR #422 for the return type of gsl::copy).

Two examples from mentioned document:

    template <InputRange Rng, class T, class Proj = identity>
    requires IndirectRelation<equal_to<>, projected<iterator_t<Rng>, Proj>, const T*>()
    safe_iterator_t<Rng>
    find(Rng&& rng, const T& value, Proj proj = Proj{});
  
    [... other versions of find ...]

>1 Returns: The first iterator i in the range [first,last) for which the following corresponding conditions
hold: invoke(proj, *i) == value, invoke(pred, invoke(proj, *i)) != false, invoke(pred,
invoke(proj, *i)) == false. Returns last if no such iterator is found.
2 Complexity: At most last - first applications of the corresponding predicate and projection.
 
    template <InputRange Rng, WeaklyIncrementable O>
    requires IndirectlyCopyable<iterator_t<Rng>, O>()
    tagged_pair<tag::in(safe_iterator_t<Rng>), tag::out(O)>
        copy(Rng&& rng, O result);

> 1 Effects: Copies elements in the range [first,last) into the range [result,result + (last -
first)) starting from first and proceeding to last. For each non-negative integer n < (last -
first), performs *(result + n) = *(first + n).
2 Returns: {last, result + (last - first)}.
3 Requires: result shall not be in the range [first,last).
4 Complexity: Exactly last - first assignments.

------------

Approaching this from a user's perspective: What would you like the gsl_algorithm version of this to look like:

    span<int> s = ...
    auto it = std::rotate(s.begin(), std::find(s.begin(),s.end(), 5), s.end());

my current design would result in something like this:

    span<int> s = ...
    auto it = gsl::rotate(s, gsl::find(s, 5));

But I can certainly adapt it if you envision something different

neilmacintosh:
Sorry @MikeGitb, I was confusing myself there. Ranges design has evolved a lot over time and my brain is not always so good and tracking it all. You are correct that the current TS does return iterators - like the existing STL. 

In order to do that and respect the fact that iterators should not outlive the `span` to which they belong, then you'd have to follow the approach that the Ranges TS takes and take the `span` parameters as rvalue-references (so they bind to pretty much anything). This does create the risk of returning a "dangling" iterator to a temporary `span`, of course. Ranges work around that with lots of template magic. 

The essence of the approach here would be to assume that `span` will act as a Range (more precisely, a View), and so you're just providing some local implementations of the Ranges TS while we wait for it to be standardized and appear "in-box" in supported environments.

An alternative approach would be to not provide the algorithms here and just use the Ranges implementation if it is available to you in your environment. (that is definitely the simpler and probably more robust path).

MikeGitb:
To be honest I personally would prefer a solution within the gsl library: For one it is hard to tell, when the TS-support will be widespread enough such that those functions can be used in cross-platform projects and until span is standardized itself I ´don't expect them to provide the span-specific optimizations.

Now it should be not too difficult to write a local implementation of the ranges TS-algorithms without the concept checks and SFINAE them away for anything except `gsl::spans`. However, I'm personally more inclined to either declare overloads for R-value reference as deleted or just ignore the problem for now. As far as I can tell, those functions won't generate "hidden" temporary spans, so the danger is rather limited:

    std::vector<int> v{1,4,5,6,7};
    auto it = gsl::fiind(v, 5); // compile time error
    auto it = gsl::fiind(gsl::span<int>(v), 5); //compiles, but temporary is obvioous

Also the difference for the user (once he can switch to the TS) should be pretty small.

In the end the question is if you want those functions in the gsl or not. I believe it would help its adoption, but personally I can live with having my private extension for now.

PR #429, opened 2016-12-07
Fix #428

### LINKED ISSUE #428: -Wundef failure from _MSC_VER
```
GSL/gsl/gsl_util:96:5: error: '_MSC_VER' is not defined, evaluates to 0 [-Werror,-Wundef]
#if _MSC_VER <= 1800
```
This needs to check if defined before using <=.

garyfurnish:
This seems pretty safe.

### DIFF
--- a/gsl/gsl_util
+++ b/gsl/gsl_util
@@ -93,7 +93,7 @@ inline final_act<F> finally(F&& f) noexcept
 }
 
 // narrow_cast(): a searchable way to do narrowing casts of values
-#if _MSC_VER <= 1800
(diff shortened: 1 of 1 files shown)


neilmacintosh:
Thanks for catching this and providing the PR @garyfurnish! Yes, that fix looks pretty safe to me too. Normally I try to define an appropriate name up in the top "configuration ugliness" of the file and use that as a feature-test. Something like MSVC_HAS_NO_STD_FORWARD or whatever. But given this is for MSVC 2013 support - which is not going to hang around for that much longer - I'm ok with doing it the simple way.

Issue #434, opened 2016-12-08
Don't const-qualify value_type in span::const_iterator


ericniebler:
I see this in span_iterator:

```c++
using value_type =
    std::conditional_t<IsConst, std::add_const_t<typename Span::element_type>,
                       typename Span::element_type>;
```

That's probably non-standard. See https://stackoverflow.com/questions/41014661/what-does-commonreference-used-by-readable-in-range-v3-do for a discussion.

neilmacintosh:
Thanks for catching that issue @ericniebler. Would you like to offer a PR to fix it?

ericniebler:
I'm not set up to build and test gsl. I can offer an untested local patch if that helps.

neilmacintosh:
Sure! I always like to offer people a chance to fix what they find. How about this: if you want to throw an untested patch at us, then I'll do whatever fixups are necessary to make it pass and get it in.

ericniebler:
Looks like tests are passing now after a few false starts.

PR #435, opened 2016-12-09
Don't const-qualify span::const_iterator's value_type

ericniebler:
Fixes Microsoft/GSL#434


ericniebler:
Third time's the charm. Sorry for the mess. Suggest squash-and-merge to keep history clean.

neilmacintosh:
Thanks very much @ericniebler!

Issue #431, opened 2016-12-07
Mismatch/misplaced pragma warning push/pop in file multi_span


ErnestPenfold:
The last pragma warning pop in line 2226 is misplaced which leads to a warning C5031. Must be moved to line 2218 or at least outside #if defined(GSL_THROW_ON_CONTRACT_VIOLATION) at the end of the mentioned file.

neilmacintosh:
Thanks for noticing this and raising theissue @ErnestPenfold. Would you like to offer a PR to fix the problem?

neilmacintosh:
Thanks @ErnestPenfold for PR #436 that fixed this!

PR #436, opened 2016-12-17
Fix #pragma warning push/pop mismatch in multi_span (Issue #431)

ErnestPenfold:
Rearranged preprocessor code at end of file multi_span


neilmacintosh:
Looks good to me @ErnestPenfold! Thanks very much for the fix!
(and sorry for the delay in picking it up)

Issue #291, opened 2016-05-14
span.h(96): warning C4003: not enough actual parameters for macro 'max'


LeftBased:
Get that warning from span.h that came from GSL.

emlai:
Seems like a duplicate of #253.

The solution is to simply `#define NOMINMAX` before including `windows.h` in your project.

neilmacintosh:
Hi @LeftBased, @emlai  is correct. This is not an issue with the GSL but with Microsoft's use of a macro to define `max`. The referenced issues should give you a method to resolve the problem.

alexey-malov:
Pull request #512 fixes this issue by replacing
```c++
static const SizeType max_value = std::numeric_limits<SizeType>::max();
```
with
```c++
static const SizeType max_value = (std::numeric_limits<SizeType>::max)();
```

neilmacintosh:
We've covered this issue before and our position is that this is a problem for the relevant header files that define `max` as a macro, rather than this library.

PR #439, opened 2017-01-13
Fix MSVC warning

RaptorFactor:
Inhibit macro expansion for call to numeric_limits<T>::max in SizeTypeTraits. Resolves warning C4003 which is generated under MSVC in some circumstances.

### DIFF
--- a/gsl/multi_span
+++ b/gsl/multi_span
@@ -87,7 +87,7 @@ namespace details
     template <typename SizeType>
     struct SizeTypeTraits
     {
-        static const SizeType max_value = std::numeric_limits<SizeType>::max();
+        static const SizeType max_value = (std::numeric_limits<SizeType>::max)();
     };
 
     template <typename... Ts>


neilmacintosh:
Thanks for the PR @RaptorFactor. However, this is not a problem in the GSL. It is a problem in the Windows headers that choose to `#define` the `max` and `min` tokens by default. If you are hitting this warning, then we recommend that you `#define NOMINMAX` for compilation, as that will ask the Windows headers to do the right thing. Please see #291 as an example of where we've discussed this topic before.

RaptorFactor:
I would've thought that having the library work under the default configuration would be desirable given how trivial and non-invasive the workaround is. Especially since I seem to recall that in some cases defining NOMINMAX can break other system headers (ATL comes to mind, but that was a long time ago since I last checked). Is there a reason to not do this other than some sort of argument for 'purity'? Just curious... Thanks.

neilmacintosh:
The principle here isn't purity, but giving users the information they need to make decisions for themselves.

If we suppress or workaround the warning, then the user has no way of knowing that their project settings mean we are not going to call the function that the source code in the library says is being called.

If we leave the code as-is, then users can make the decision that's appropriate for them - either use NOMINMAX to resolve the issue (depending on their version of MSVC and the libs in use, this may be straightforward), or suppress/ignore the warning as something that is acceptable given their other project constraints.

RaptorFactor:
Thanks for taking the time to respond. However I'm still a little unclear, because in this case the workaround DOES ensure that we "call the function that the source code in the library says is being called". This is not just a suppression, because by adding the parentheses you guarantee that you actually call numeric_limits<T>::max (as opposed to some macro nonsense interfering), right?

neilmacintosh:
Sure, but then you hit the simplicity argument. Why should we have to write the function call with unusually ugly syntax to work around a historical mistake that only exists on one platform? And how long would that workaround then persist? It becomes the "thin end of the wedge" ;-)

I would rather write the code naturally and when people hit it (which hopefully they do less and less often), they can make a decision about how to correct their environment (and in many cases, make their environment better by reducing the number of macros flying around).

MikeGitb:
Quite frankly: Don't thoe two braces pale in comparison to the amount of pragmas and macro machinery the code currently has in order to suppress warnings on a single platform / to support it at all? ;)

neilmacintosh:
@MikeGitb hahaha, it is definitely the case at the moment that the ugly pragmas and macros are in the majority. Without wanting to continue what is a relatively pedantic discussion too much further...I do see a difference. A lot of those pragmas and macros are about to disappear: we are just waiting on VS 2017 to release. There would be no aim (or driver or in-source reminder) to remove the unnecessary parentheses. That's the difference as I see it.

I was happy to have the pragmas and macros so that we could initially support VS 2013 - because I knew they would only have to exist as long as our support for VS 2013. In the long term, our aim is to be portable C++11 (14, 17 and so on, as time goes on). Part of that aim include using straightforward C++ without odd hacks to work around shortcomings in specific platforms.

gsl/multi_span
```diff
@@ -87,7 +87,7 @@ namespace details
     template <typename SizeType>
     struct SizeTypeTraits
     {
-        static const SizeType max_value = std::numeric_limits<SizeType>::max();
+        static const SizeType max_value = (std::numeric_limits<SizeType>::max)();
```
gdr-at-ms: This is a problem with windows headers. Best dealt with differently than transmogrifying source code.  E.g. could define NOMINMAX on build command line.

Issue #430, opened 2016-12-07
natvis support for GSL


LYP951018:
For reference:
``` XML
<?xml version="1.0" encoding="utf-8"?>  
<AutoVisualizer xmlns="http://schemas.microsoft.com/vstudio/debugger/natvis/2010">  
 <Type Name="gsl::span&lt;*,-1&gt;">
    <Intrinsic Name="size" Expression="storage_.size_" />
    <DisplayString>{{ size={size()} }}</DisplayString>
    <Expand>
      <ArrayItems>
        <Size>size()</Size>
        <ValuePointer>storage_.data_</ValuePointer>
        </ArrayItems>
    </Expand>
  </Type>

  <Type Name="gsl::span&lt;*,*&gt;">
    <Intrinsic Name="size" Expression="$T2" />
    <DisplayString>{{ size={size()} }}</DisplayString>
    <Expand>
      <ArrayItems>
        <Size>size()</Size>
        <ValuePointer>storage_.data_</ValuePointer>
        </ArrayItems>
    </Expand>
  </Type>
</AutoVisualizer>
```
Hope that helps :)

neilmacintosh:
Thanks @LYP951018, I'm sure that Visual Studio users will find that particularly helpful.

neilmacintosh:
@LYP951018 I'd be happy to add this to the repo as a file, so that others can easily consume it. Would you be able to put together a PR that adds the file? That way I can be sure we've got all the right legal sign-off to accept the contribution.

neilmacintosh:
This issue has now been closed by #440. Thanks for the original suggestion @LYP951018  and thanks for the PR @GuildMasterInfinite.

PR #440, opened 2017-01-14
Add Visual Studio Native Visualiser support

GuildMasterInfinite:
This pull request adds a .natvis file to the root of the repo. This file can be added to Visual Studio projects that use the GSL (using File > Add), and, when debugging, the GSL classes will appear with custom visualisers (e.g., spans will have their extent shown, when expaned they will have elements).

 This commit adds support for all the (publicly exposed) types in the GSL, except for `multi_span`. I've been unable to add support for those, due to the fact they use variadic templates and Natvis has no support for this kind of templates.

 This commit also closes issue #430

### DIFF
--- a/.gitattribute
+++ b/.gitattribute
@@ -1,3 +1,3 @@
 # .natvis files are not properly detected by GitHub as .xml, so we override it here.
 # See https://github.com/github/linguist#overrides
-*.natvis linguist-language=XML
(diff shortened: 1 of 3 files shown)


neilmacintosh:
Thanks @GuildMasterInfinite this looks good to me.

PR #444, opened 2017-01-23
Fix gsl_byte in clang 3.9.1

hunger:
type_traits are needed for gsl_byte to be used in clang 3.9.1


neilmacintosh:
Nice catch @hunger, this looks good to me. Thanks for the PR!

PR #445, opened 2017-01-27
Make is_default_constructible work for spans.

CaseyCarter:
..so that `std::is_default_constructible<gsl::span<int, 42>>::value` doesn't `static_assert`. Makes spans of static extent usable with range-v3.

### DIFF
--- a/gsl/span
+++ b/gsl/span
@@ -308,7 +308,7 @@ namespace details
         constexpr extent_type() noexcept {}
 
         template <index_type Other>
-        constexpr extent_type(extent_type<Other> ext) 
+        constexpr extent_type(extent_type<Other> ext)
         {
             static_assert(Other == Ext || Other == dynamic_extent,
                           "Mismatch between fixed-size extent and size of initializing data.");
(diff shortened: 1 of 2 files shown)


gsl/span
```diff
@@ -359,6 +359,7 @@ public:
     constexpr static const index_type extent = Extent;
 
     // [span.cons], span constructors, copy, assignment, and destructor
+    template <bool Unused = false, class = std::enable_if_t<(Unused || Extent <= 0)>>
```
neilmacintosh: Can you leave a comment that explains why Unused exists? I'll have to admit that I don't know why ;-)

CaseyCarter: Since `Extent <= 0` does not depend on a parameter of this function template, `std::enable_if_t<Extent <= 0>` is ill-formed when `Extent` is greater than `0`. If we tried to constrain this constructor template on simply `std::enable_if_t<Extent <= 0>` the compiler would rightfully complain about `std::enable_if<false>` having no member named `type`. `Unused || Extent <= 0`, on the other hand, *is* an expression that depends on a template parameter. If, when the compiler tries to substitute `false` for `Unused` in `std::enable_if_t<(Unused || Extent <= 0)>` and discovers that `std::enable_if<false>` has no member named `type`, SFINAE kicks in and removes the overload from the set of viable candidates rather than reporting an error.

(`Unused` is actually a poor name here. `Dependent` is a little more suggestive. I'll change it, and leave a more concise comment in the code.)

neilmacintosh:
Thanks @CaseyCarter, that's perfect. I wondered if that boolean template parameter had something to do with SFINAE...thanks very much for educating me!

PR #446, opened 2017-02-05
Reorganize include structure:

### LINKED ISSUE #441: including GSL also includes GSL/tests
There was the issue https://github.com/Microsoft/GSL/issues/277 that was fixed. Now we are supposed to do /I c:\GSL. This also includes everything that is in c:\GSL\tests what you maybe don't want.

I suggest doing deeper nesting: having c:\GSL\include and inside the gsl folder. So you would include c:\GSL\include. See https://github.com/electronicarts/EASTL. They have similar structure which works well.

CaseyCarter:
* Nest "gsl" directory inside a new "include" directory.

* use `constexpr` on VS2017! 

* Cleanup the `_MSC_VER` conditionals a bit.

* Don't `#define noexcept` on non-Microsoft implementations.

* Workaround VS2017 bug in `multi_span` (See https://godbolt.org/g/cJqgbi). Also implement `==` and `!=` for `static_bounds_dynamic_range_t` because I'm an EoP semantic soundness snob.

Fixes #441.


CaseyCarter:
The OSX Travis failures are transient; the same commit passed CI successfully [on my Travis](https://travis-ci.org/CaseyCarter/GSL/builds/198623579).

include/gsl/gsl
```diff
@@ -19,32 +19,28 @@
...
-#define constexpr /*constexpr*/
-
-// MSVC 2013 workarounds
-#if _MSC_VER <= 1800
-// noexcept is not understood
-#pragma push_macro("noexcept")
-#define noexcept /*noexcept*/
+#if defined(_MSC_VER) && _MSC_VER < 1910
+ #pragma push_macro("constexpr")
+ #define constexpr inline
```
gdr-at-ms: I would rather we don't to that define, in case we need a constexpr variable at some point.

include/gsl/gsl
```diff
@@ -146,27 +142,23 @@ private:
 namespace std
 {
 template <class T>
-struct hash<gsl::not_null<T>>
+struct hash<::gsl::not_null<T>>
 {
```
gdr-at-ms: I understand where you're coming from, but please, let's not write it that way.  Several instances below and elsewhere.

include/gsl/gsl_algorithm
```diff
@@ -19,21 +19,19 @@
...
-// more targeted suppressions will be added in a future update to the GSL
-#pragma warning(disable : 26481 26482 26483 26485 26490 26491 26492 26493 26495)
+ // turn off some warnings that are noisy about our Expects statements
+ #pragma warning(disable : 4127) // conditional expression is constant
 
+ // blanket turn off warnings from CppCoreCheck for now
+ // so people aren't annoyed by them when running the tool.
+ // more targeted suppressions will be added in a future update to the GSL
+ #pragma warning(disable : 26481 26482 26483 26485 26490 26491 26492 26493 26495)
 #endif // _MSC_VER
```
gdr-at-ms: Do we already have open issue to track those?

include/gsl/gsl_byte
```diff
@@ -52,69 +48,69 @@ enum class byte : unsigned char
...
 template <class IntegerType, class = std::enable_if_t<std::is_integral<IntegerType>::value>>
-inline constexpr byte& operator<<=(byte& b, IntegerType shift) noexcept
+constexpr byte& operator<<=(byte& b, IntegerType shift) noexcept
 {
     return b = byte(static_cast<unsigned char>(b) << shift);
 }
 
 template <class IntegerType, class = std::enable_if_t<std::is_integral<IntegerType>::value>>
-inline constexpr byte operator<<(byte b, IntegerType shift) noexcept
+constexpr byte operator<<(byte b, IntegerType shift) noexcept
```
gdr-at-ms: Please undo this change.  Several instances here and elsewhere.

include/gsl/gsl
CaseyCarter: There's a trade-off being made here. We can either:

* Define `constexpr` as whitespace/empty, and use the abominable `constexpr inline` on functions and `constexpr const` on data,

* Define `constexpr` as `inline` so it can be used directly on functions. Define a separate macro for data (I used `GSL_CONSTEXPR_DATA` in `<gsl/multi_span>`) as either `constexpr` or `const`.

Both options require some non-idiomatic "garbage" syntax. For the first option, *every* usage of `constexpr` must instead be `constexpr inline` (function) or `constexpr const` (data). For the second, `constexpr` functions are written idiomatically, and only `constexpr` data - which is at least two orders of magnitude less common than `constexpr` functions in the current GSL - needs to instead be tagged with `GSL_CONSTEXPR_DATA`.

I prefer option two since it is less intrusive overall. If you disagree, say so and I'll retag all the `constexpr` entities.

include/gsl/gsl
CaseyCarter: Will revert.

include/gsl/gsl_algorithm
CaseyCarter: No idea; I don't even know what these warnings are. These warning disables are pre-existing, I only changed the indentation.

include/gsl/gsl_byte
CaseyCarter: If you disagree with the argument presented [above](https://github.com/Microsoft/GSL/pull/446#discussion_r99503826), I'll revert these.

include/gsl/gsl
gdr-at-ms: I don't consider "constexpr inline" and "constexpr const" idiomatic garbage.  Even if they were, defining "constexpr" to nothing is strictly more useful.

.gitignore
```diff
@@ -11,3 +11,6 @@ tests/*tests.dir
 Testing/Temporary/*.*
 CMakeCache.txt
 *.suo
+.vs/
+.vscode/
+build*/
```
CaseyCarter: Ignoring `.vs/` and `.vscode` is all fine and dandy, but I did not intend to commit my "ignore all the build folders" line here. Shout if you'd like me to remove this from the PR.

include/gsl/gsl_algorithm
CaseyCarter: @neilmacintosh, is there a bug# for these in VSO that we could put in this comment for cross-referencing?

neilmacintosh: I don't like putting VSTS bugs into open source code, as noone outside of us MSFT people can see them. So they are less than useful. I'll get to these over the next few weeks as I commit some other tidy-ups.

.gitignore
neilmacintosh: I guess we should probably pull it out to be "tidy".

CaseyCarter:
All comments addressed.

gdr-at-ms:
Looks good to me.  Thanks!

neilmacintosh:
Thanks @CaseyCarter, this looks great!

PR #449, opened 2017-02-11
not_null updates:

### LINKED ISSUE #447: gsl::not_null still supports pointer arithmetic
````
int i = 0;
gsl::not_null<int*> p(&i);
p[0];
1 + p;
p - p;
````

You need:

    void operator[](size_t) = delete;

I suggest removing the members `operator+` and `operator-` and adding deleted overloads at namespace scope:

````
template<class T>
  void operator+(not_null<T>, std::ptrdiff_t) = delete;
template<class T>
  void operator+(std::ptrdiff_t, not_null<T>) = delete;
template<class T>
  void operator-(not_null<T>, std::ptrdiff_t) = delete;
template<class T, class U>
  void operator-(not_null<T>, not_null<U>) = delete;
````

Is there a reason the deleted members are private _and_ deleted? Making them private seems redundant (and makes GCC give two errors rather than one).

Finally, strictly speaking that file should be using `std::size_t` not `size_t` because it doesn't include `<stddef.h>`.

### PR DESCRIPTION
**Pull request by CaseyCarter**
* constexpr all the things.

* remove `operator=(const T&)`
  * it leaves `*this` in an invalid state if `ensure_invariant` fails
  * implicitly converting the `T` to `not_null` and then assigning is in every way superior.

* Enforce the invariant after conversion from `not_null<U>`. For a (poorly designed) user type, a non-null `U` could convert to a null `T`. For pointers, the check should optimize away.

* remove the converting assignment operator; again let the conversion constructor and self-assignment operator do the work.

*  Define macro `GSL_ASSUME` in `gsl_util` to semi-portably inform the optimizer that a predicate MUST hold. Use `GSL_ASSUME` in `not_null::get`.

* Cover the remaining pointer arithmetic operations as Wakely suggests in issue #447.

* Implement unary `+`, because it's valid and a handy library writer's trick to convert an lvalue expression to a prvalue.

Together with #448, closes #447.

### DIFF
--- a/include/gsl/gsl
+++ b/include/gsl/gsl
@@ -62,80 +62,62 @@ using owner = T;
 // Has zero size overhead over T.
 //
 // If T is a pointer (i.e. T == U*) then
-// - allow construction from U* or U&
+// - allow construction from U*
 // - disallow construction from nullptr_t
 // - disallow default construction
-// - ensure construction from U* fails with nullptr
+// - ensure construction from null U* fails
 // - allow implicit conversion to U*
 //
 template <class T>
 class not_null
 {
+public:
     static_assert(std::is_assignable<T&, std::nullptr_t>::value, "T cannot be assigned nullptr.");
 
-public:
-    not_null(T t) : ptr_(t) { ensure_invariant(); }
-    not_null& operator=(const T& t)
-    {
-        ptr_ = t;
-        ensure_invariant();
-        return *this;
-    }
+    template <typename U, typename Dummy = std::enable_if_t<std::is_convertible<U, T>::value>>
+    constexpr not_null(U&& u) : ptr_(std::forward<U>(u)) { Expects(ptr_ != nullptr); }
+
+    template <typename U, typename Dummy = std::enable_if_t<std::is_convertible<U, T>::value>>
+    constexpr not_null(const not_null<U>& other) : not_null(other.get()) {}
 
     not_null(const not_null& other) = default;
     not_null& operator=(const not_null& other) = default;
 
-    template <typename U, typename Dummy = std::enable_if_t<std::is_convertible<U, T>::value>>
-    not_null(const not_null<U>& other)
+    constexpr T get() const
     {
-        *this = other;
-    }
-
-    template <typename U, typename Dummy = std::enable_if_t<std::is_convertible<U, T>::value>>
-    not_null& operator=(const not_null<U>& other)
-    {
-        ptr_ = other.get();
-        return *this;
-    }
-
-    // prevents compilation when someone attempts to assign a nullptr
-    not_null(std::nullptr_t) = delete;
-    not_null(int) = delete;
-    not_null<T>& operator=(std::nullptr_t) = delete;
-    not_null<T>& operator=(int) = delete;
-
-    T get() const
-    {
-#ifdef _MSC_VER
-        __assume(ptr_ != nullptr);
-#endif
+        Ensures(ptr_ != nullptr);
         return ptr_;
-    } // the assume() should help the optimizer
+    }
 
-    operator T() const { return get(); }
-    T operator->() const { return get(); }
+    constexpr operator T() const { return get(); }
+    constexpr T operator->() const { return get(); }
 
-    bool operator==(const T& rhs) const { return ptr_ == rhs; }
-    bool operator!=(const T& rhs) const { return !(*this == rhs); }
-private:
-    T ptr_;
-
-    // we assume that the compiler can hoist/prove away most of the checks inlined from this
-    // function
-    // if not, we could make them optional via conditional compilation
-    void ensure_invariant() const { Expects(ptr_ != nullptr); }
+    // prevents compilation when someone attempts to assign a null pointer constant
+    not_null(std::nullptr_t) = delete;
+    not_null& operator=(std::nullptr_t) = delete;
 
     // unwanted operators...pointers only point to single objects!
-    // TODO ensure all arithmetic ops on this type are unavailable
-    not_null<T>& operator++() = delete;
-    not_null<T>& operator--() = delete;
-    not_null<T> operator++(int) = delete;
-    not_null<T> operator--(int) = delete;
-    not_null<T>& operator+(size_t) = delete;
-    not_null<T>& operator+=(size_t) = delete;
-    not_null<T>& operator-(size_t) = delete;
-    not_null<T>& operator-=(size_t) = delete;
+    not_null& operator++() = delete;
+    not_null& operator--() = delete;
+    not_null operator++(int) = delete;
+    not_null operator--(int) = delete;
+    not_null& operator+=(std::ptrdiff_t) = delete;
+    not_null& operator-=(std::ptrdiff_t) = delete;
+    void operator[](std::ptrdiff_t) const = delete;
+
+private:
+    T ptr_;
 };
+
+// more unwanted operators
+template <class T, class U>
+std::ptrdiff_t operator-(const not_null<T>&, const not_null<U>&) = delete;
+template <class T>
+not_null<T> operator-(const not_null<T>&, std::ptrdiff_t)  = delete;
+template <class T>
+not_null<T> operator+(const not_null<T>&, std::ptrdiff_t) = delete;
+template <class T>
+not_null<T> operator+(std::ptrdiff_t, const not_null<T>&) = delete;
 
 } // namespace gsl
 

### CODE REVIEW

File: include/gsl/gsl [hunk 1]
```diff
@@ -74,69 +74,65 @@ class not_null
...
     }
 
-    template <typename U, typename Dummy = std::enable_if_t<std::is_convertible<U, T>::value>>
-    not_null& operator=(const not_null<U>& other)
+    constexpr T get() const
     {
-        ptr_ = other.get();
-        return *this;
+        // help the optimizer
+        GSL_ASSUME(ptr_ != nullptr);
```
neilmacintosh: I appreciate the aim to aid the optimizer. But could we combine this with also testing/indicating the contract on the function? How about replacing the `GSL_ASSUME()` with an `Ensures()` and then updating `Ensures()` (for the relevant platforms) to use `__assume()`?

include/gsl/gsl
```diff
@@ -74,69 +74,65 @@ class not_null
...
+        return ptr_;
     }
 
+    constexpr operator T() const { return get(); }
+    constexpr T operator->() const { return get(); }
+
+    // unary +: because there's no particular reason to forbid it,
+    // and it's sometimes useful to convert an lvalue to a prvalue.
```
neilmacintosh: I have no problem with this. I learned something today: that unary `+` is useful to library implementers ;-)

include/gsl/gsl
CaseyCarter: Aside: This is pre-existing (See line 110 red below). I didn't add the assume, I simply made it work on GCC and Clang as well as VC++.

It would be pointless to incorporate an assumption into `Ensures`: `Ensures` checks the condition, and optimizers already recognize that `if (!foo) { throw XXX; /* or */ call_no_return_function(); }` ensures that `foo` holds in the following.

I agree with the original code's choice to use assumption rather than verification here. In my experience, compilers are fairly stupid about recognizing that invariants established by all the constructors and modifying member functions must hold in the non-modifying members. A library component like this is only useful when the abstraction cost is extremely low, and I think it will fail to meet that bar if we force the compiler to revalidate the class invariant on every access for which it fails to prove that invariant holds.

I'm not a big fan of postcondition validation in general: it's great when you're developing and validating a new component, but destroys zero-cost abstractions when applied to library components. Users don't want to pay debugging costs for our libraries, they want our libraries to work and get out of the way so they can debug their own code.

include/gsl/gsl
```diff
@@ -74,69 +74,63 @@ class not_null
...
+    constexpr operator T() const { return get(); }
+    constexpr T operator->() const { return get(); }
 
-    // prevents compilation when someone attempts to assign a nullptr
-    not_null(std::nullptr_t) = delete;
-    not_null(int) = delete;
-    not_null<T>& operator=(std::nullptr_t) = delete;
-    not_null<T>& operator=(int) = delete;
+    // unary +: because there's no particular reason to forbid it,
+    // and it's sometimes useful to convert an lvalue to a prvalue.
```
gdr-at-ms: I would rather not do that.  We shouldn't add to an interface because "there's no particular reason to forbid it", but rather we have good reasons to add to the interface and good example of uses.

gdr-at-ms:
Regarding `Ensures`: the way it works is that the check is conditional on build.  But the optimizer is always welcome to use it to aid optimization of the body; so Neil is correct.

include/gsl/gsl
CaseyCarter: I didn't realize when I added this, but it's unnecessary to overload unary `+`: the language is happy to convert implicitly to the underlying type and use its `+` operator. Consequently, unary `+` isn't an expansion of the interface but an arithmetic operation that was previously - perhaps unintentionally - valid for `not_null`.

Are we all happy with deleting / forbidding unary `+` despite that it was previously valid?

gdr-at-ms: My recommendation is to remove that from the intentional interface.  Uses of unary `+` in general will be delt with by other Core Guidelines rules -- it is a hack that I don't see part of what we would recommend people to write.

CaseyCarter:
> Regarding `Ensures`: the way it works is that the check is conditional on build.  But the optimizer is always welcome to use it to aid optimization of the body; so Neil is correct.

So the suggestion is that I implement a fourth mode for the assertions that does not check, but does optimize on the assumption that contracts aren't violated? (`GSL_CONTRACTS_ASSUMED`? `GSL_UNDEFINED_ON_CONTRACT_VIOLATION`? `GSL_NASAL_DEMONS_ON_CONTRACT_VIOLATION`?) That does seem like a useful feature for users who want to squeeze out the last 0.5% of performance, but it does fall afoul of the issue I describe above: there's no way to distinguish between an `Ensures` in just-written undebugged user code and an `Ensures` on a ten-year-old library component for which it would be pointless to check.

I suppose we could implement more of the contracts design from P0380, i.e. `(Expects|Ensures)_(default|axiom|audit)`, with new user knobs for "optimize/do not optimize assuming unchecked contracts hold" and "check/don't check `audit` contracts." We could then mark this postcondition in `not_null` as `Ensures_audit` so that it's only checked under "heavy" debug conditions.

include/gsl/gsl
neilmacintosh: I'm happy with removing it from the interface if it's not needed for any purpose (and it seems it no longer is...I was assuming you had a valid "library-writer's weird template need" use-case...but it sounds like that is already taken care of).

neilmacintosh:
Not quite.

Firstly, I think it’s a great idea that the Guidelines eventually align with whatever contracts-in-the-language end up looking like. But we’re not there yet, and even if we were, I wouldn’t mix that work into this PR.

Secondly, I think GDR and I are in agreement here that a postcondition is the right thing. Users can decided whether they want it enforced or not with a build flag (if it’s not enforced, an __assume()-type hint could still remain). I don’t think we need a fourth mode to implement that. Compilers can see it and use it to optimize if they can/choose to. I think what we are aiming for is correct semantics here, with enough hints that compilers are free to do the right thing.

include/gsl/gsl
CaseyCarter: No, the only "weirdness" here was that my pass looking for unintended arithmetic operators happened long enough after my pass to remove unneeded operator overloads that I didn't realize that unary `+` didn't need an overload.

I'm interpreting the consensus here as "remove this useless overload but don't `delete` unary `+`."

CaseyCarter:
For now, `not_null::get` simply has an `Ensure(ptr_ != nullptr);`. I'll change `GSL_UNENFORCED_ON_CONTRACT_VIOLATION` to use `__assume` in a follow-up, since it's very much a separate issue.

neilmacintosh:
Sounds perfect, thanks!

gdr-at-ms:
No, we don’t need to get dramatic or to inflame or to inflate the GSL_* macros.
The observation and the note are actually simple, when you look at them carefully.  We don’t need a fourth mode and that never was the suggestion.

MikeGitb:
Actually if what @CaseyCarter says is true (that compilers are bad to determine class invariants on their own) I would very much like to have to two different post condition macros:
- One that checks if I actually satisfied those conditions and terminates otherwise (This is what the contract macro `Ensures` currently does and can/should additionally contain something like gcc's `__builtin_expect` where available)
- One that allows me to tell the compiler that I'm 100% sure that my class/function will always satisfy a certain invariant and that it can optimize code based on that (just like `__assume` on MSVC).

gdr-at-ms:
I wouldn’t want the GSL design and implementation be driven by maligning compilers.
What I would like to see is a good example of what Casey says and have compiler writers do their jobs.
If we had to malign compilers, we shouldn’t even be using C++…

CaseyCarter:
> I wouldn’t want the GSL design and implementation be driven by maligning compilers.
What I would like to see is a good example of what Casey says and have compiler writers do their jobs.

Comment out the `Ensures` in `not_null::get`, and observe the assembly generated for this function by any compiler ([Here's the result for MSVC/GCC/Clang](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAKxAEZTUAHAvVAOwGdK0AbVAV2J4OmAIIcAtiADkABmmk0Exnh6YA8mwDCCAIZtgmGfICUpDgOLIj0gPT2Hjp85eu37j59sBSWaPu%2B/rYA1FpMAJ5CwAgEwRDIJsEATLK0AKzBALJ4yMSoFgBmsWHEjCS6LOwAdMGiPDzBUTEcjZgixABumOhVgQF%2B9sEAKgjCwWhYwWM8OZic3cH8bFjEwQQImFkAkkPBADKz83GZOya9Az4XwwAS2MEAyuoAYkMA6qIASndb98EACh91AA1LYAEWwoOC3iSSVEvx%2B0KSpGCrx213UAFVdu8Ph9RAA5IYATWCz1q%2BJJAGktvjQcjsAANAHYe6/dQfPohLaZP57LYQ5E0rR7DGgmkAcWCACEscF8epdnyTkMIcN1Dc7ji8YT%2BWynllsB8tNcCUNRFKtnziaROcEnjt8Szfk92bV/p8hlstBi9p9/hiPn91Pc7gTIfL8TSnh8JdhMthCTUaXL1dggQndvcTXs9hrbaIseiPmyPqF1H8iTHxdddui9uDi9K7nzzXs7i7SwSScLRNzkaDRJlROKWaTSwrbhyri2LVaicjXrchpPgsmCbUtJ71PjSfqtNuhnjN8ihuzdq6J4bUSHkZ8fhK7YDMjarpjz/rXcn9/jHZuttuUTRDUHmeN5PjuV1lzuDEQzHUkoNLcFRD5fFxXhfFbSgkCXhxbBziCTkvCI4iSNI/pRECaEAGY8AKZZMAKYI0L2AB9b9D1ETd7hY65KKSKisAKPA2E2Zi2IPI8hm43i/D4mi2GQHh%2BEmbwqK0TAAA9rGYVg2FU7A5OExTlM2VStA4AgsC0zBmH0yiLkwjZiE2XRnLWBBnM2NA2CE4BBAqXTgiYSpOGCAoSHcsZmKmZQ1AkOYCAC9gwGkFoACNMD0DpWGIW0AHcNjYYJGGc2wygs8Z2HwEKWnYdzRPuXMCHCRg2mCVzNmy1AeAqboQE5W1aBqMSVQ%2BE58VEFUWO3cTCUkliQXUX0t3xEBggs9AQBAAhMGICRhN64I8tUBoMvGXR6gWCBBN0fgeAIExbSSYbGpY5dAVeab8VmjjNwW/9lv/Vb2uCYAOB4LaCl0VQWKhirNO0kKjpO4IzvWPI8r0q4qJe1iMUdfEOy0CEvp%2B%2BbFsB7c1rYVB1mE4Bgj0RhWs4AbZP42jgjASBBOE7oIBG64PtJ9jyYByagcSbwAHYtBlrRgl5kT0AF17RvGybsBFiTOKGf6lol7cpdl%2BXgihNI5b8M3rcVhi%2BZVsT8YTImSZm0Xdf1yn8RMR6reopWGtY9WaU17W5o9inDYw9mqLmfACnsij%2BIDpjXvuQ8JS2J4iRY8EzUtCANKl/iNLklOxPTmNUKzolC8SCuM%2Br7Pc%2BwfO9jrxPqM5gOVZYljFP0YA%2B%2BNuXZdtoTlYgPvxXxDEtGHwIoWTu2RNT1i%2BUpbA9lrovgj7tL%2BFUFg2BYzTWuQWIIG57m6%2BRWhfdEJeBJXwOWPxjet53xJ98P%2B7hNPjS59L7X0gEXZEsgH7UUwDwEQi9/YvzXixD%2B2865xBAXXB%2BT9y6vXflsTeKDd5X2vhguScdaKJ3IoMMScJ7gYnjBAEq3Qci9QfpQkIQxoENHWJsYKeB9oAC9druQqMVZy%2BBkCHQkPwCqCBuo9FXLEMYSwODn1ongBYBVMDcNWBFGmBBbTcNEUwiRO0pgtF0BffgF0eDhGCJgDoF0rE7R6GzJONECiCT3pke4890xTkfvAier8aF0OwPEKq38WK6A4Bwfg8VwnLEgfxaB3cX6937j1AwC8/bLyCYgkJ9DvLoEiQfI%2B/9omxPiUUpJscZiMR7lPFiM857ZICbkvm%2BTWShIScUuIPSpZUX1BZAKyB%2B7RP0WpDoqA8DoH0hACBwQ1o/zKSfJYzkLF6DSmoCAPtSEwMwHA9pq9qFdMKRE9aiUWCjIkRZMyUyZlzPQdUmpZCE6yQclcZiVQKm7QIGtH5xAaquMCGwXQ8UVEWM2GDHglFpZSkCBZYg/AL5hWhjwWG4zFnFX4FsnIFzNogCRWwFg8VT7EDyLlWScLF7WzPjMZAhBUUwzhgQeIehVjeQsgAKkqpwWI4KOC6EMIkNaG0tpEpJZgMlFKIACqFZgY28LpaglhSqqi8KqUqveW41JQSHZqyFuoT6bsdZ/UjitSBOTn55LEu7P6xpsBaEpBAZqrVkTVJtp6r13qfW%2Br9f67wFsaVm1VuvPBn9%2BlQkGRckZYzbmTOmbMqiYSFlrX9emjNmazaBsto/b16NUB5VBuDSGaKMUWQgIiZia1ERrBaqZGETKlJuREbWvu9o2x9yXkkGtjaG5V3FDXRpKEtYsR9patxKT6lpNDW9Q0GspomvDma8WFrE5YIQba01esHVOpdfW915ys3HpPQGoNVtrazuQbXD1qkhmXJyHGiZWh7lJpTSK/FW0dp7QOjtHZE6oF1PHvbWdTtCbsmJqCMOv09bmsluuwJHSt3Lp3bcPdrrMCHsSZ02hZzEmd2Scsch2rEOr2wIAzAF8OCRuQzBtiaHnWIj%2BM5IphBdKIiw8UsuCDsCcEEG0Gjr07WocdYxmEfx8gEFYyFDjvKuMkcI/HYIVChPbukonUF4LGCQuLTCqlGqCJXD0SxNgd09OGaCMED4bQCBCCoyDMowlv1BVWJIVysRHPEqEUQIKbAbGMzkcEGmbAAC0pn6jBAcUpNo%2BE2HBGuNE4Igi8jrTwIIoKXRiAbF0OgDLQihixcuJZrYjFdhjF0MVaZXnVgQDwFUTANRdiqTVZCDEXLEjcKxpZkL7V6iFt5Yi5FSMCh5AkMENrtoev4EFX1otnLbNDcCiN1AY3ws8GYMQFi%2BirhTeENY/rN07qxHm0ii%2B7Gdt2L425E7i26rLdW2Z8bPKoaqA4JN3rfAi18MYPSxl3lMscECj5ibHzRA7ViodMyA8YnDDsn4aHLRjNrdhQZxgOL6X9QvWbYZVyokxN%2BRAMVIBhB48B8AUFWyDlqSGNCAAbMiIna2Ntbf0ltKL/BMPdt2BItgejUYuRiXgcnCwme2d6DCR66r13W3Bz9yHakMOac2BiZEiuwWbFBHEiQtjmufpAHMXQlOWK0RZ2pInJP/u/LwJTsyKvYfJrZ44qnBlk3BoGztQBqwkdmYgBiOndPFgfuZ4TyykMSB5Vckm70cz%2BA%2ByhHC4I5GgHUeZ1zMAuvRfEEl1KePWqKJY7WJgCHpizJq/iuN1X9alfBE1xIbXUbIRE4N0bk3z7zccH7uwTLLAbdqTtzTh3IB2fO7h3ms282z5e9piZn382gvT%2BR33uySRadBW0R%2B739QIC0yclUQwrK48yyVXnt3m%2BeAJIqmfgPO/doDMhId%2B6qkDM2yvyvoKrViAVBIM1i/sRX%2Br5vyzwb3Hlukfyl21Rtgn092GFBi0R2XdzdyPzdxtnsF41iWchT1sxYjTwzzMw22z2QOtmcgIEECKmZyf0QOVWl3H3YAskn3f12i/1WCGHgLnyP1aBIOICKn3x2Sf1zzdygJKhgKYEYKIGIBCzmUSDYIT2ININgIPz4JlhPwL0GEYS6GJRaCUBUB6iRk0SKgsHinYBcgIFlwIBaB8wqSFyKgqzW0qycyEXm30G2zH3nwIBny30ZzwKwIemAKwDUB2goIL3/wYM/zEJ/08PqGZx8N1z8K0SpwMzd0GCWAjy81yxENCJIA4CqGyM82/Vqj81sVyLWFQFSwMDUCCjSioEozMO5lPwXzM2vw/yYOhHhSSClHgJiOgTiMCJcOCPSKYJCxCw6Konvy6ICPAN6PqIi36LCLaJaNq2JTv1tn8PiLqLcNsJmJIEGIWOiJGOWO6ImJfymJhTf02NyjmJGJDwJQ23jgKC2yWNiPGOf2tj6KaLEIkMuKJxuNojuN2NGJWJ6JtlfRCOaLSHhTSFBCuK2m%2BIKF%2BKkNoNiE6IBImMCBKjwAcR2kxxcN2HIJRKoJRKuAkBIE2GSKcIWDOLez8Flx0NMjUgR2GHdR6hhz91dz8C%2BNs1uK2xBPeN/1cPcJhWp2X3p3dz5MX2j1dxXweLGNWKpKLzlxLzpKZJaAHwMj8Cv0FOTW5K2N5PVM3CFIZ1DxABhLhOzT2MeJlLBzlJpKjS0HpJVJBWOLMhVK1POPaLn11PtJXwNOuI5J%2BPuN8OlJ6OpPl1tKVPt1VNEA9P0hdPmPZOIE5IICw0v0dI1IMklIDORISM1WUxCCVwhWsF00TmhS2l1Kcy5WjJ2gshYmcjyk/0YGgBLRAFLOJXLM1MYEVTd1kK4OKj3zgIIM1UCGkDMAhmkDSAUDYBkFoAUFQBkDljmLaPWksALOoinNID%2BTkCHLMAAGsQAkg0hSARyAAWcc4wacmQBQLgWQNc08swOAWAJALQ1QXacgbgFbbQ3aFAHgWgAAThYlp0PNICEnul2i4AgDSlPMUBW3imJU0BsQgvwBYxYC6C4A3PIC8xHNMAYB0loJkFsDFWQBaJaJpIqlsGwGuA4GQGU3UA0mU1cmQCQFECBAZGU1eGPOHJkDHNIAnNQpnOkBXORnWGCA0gAA5acQt/zQhcwfyrMtAFB1zTBtzdz9yjyTyeLzzzAQArz5KhzSA7zEAUA3ynyyAKBwllAjKQBgBpYOBQVGAOBZFEygK8jKBwLUKlBoKCBYLwh4K8BEL0S2gIL7CMKdLeEcLpAQsNpmsCL5ypRaB%2BFggQt1AWV4qJA6KEBmtQUkLNgQtXhrF4rXgdKRzOLuL5AGAZB%2BLjpBKRKxKJLgBkBKLpY4hrLdBbL7KTA5KbzFKkhDyqhvzeq%2Br%2Br%2Br6cVKuKILeKLzNLryNzbyYB9LHy1BjLXyzL5rPzB4tsiUtzAKj4QLnKIK3KEpPLvLfLkKAr0KbysKaoZBwrLJIrCK2jYr4rErMUQsUqrA0qRiMq/K8rcrsqCqOLVKSreLyrCAEAhLRLxLDzxhMkGYXV1q2rJqFLSAdy9yDyZBjyRq1LpBxqtKOqUbpAkh/qzzMb4bNzSAAddIQBDygA%3D%3D%3D)):
```c++
gsl::not_null<int*> test_rewrap(gsl::not_null<int*> p) {
    return p.get();
}
```

Despite that it's obvious to a human on cursory inspection, compilers don't recognize the class invariant `ptr_ != nullptr` and re-validate when constructing the return object in `test_rewrap`. If you uncomment the `Ensures` in `not_null::get`, you'll get the same generated code ([Demo](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAKxAEZTUAHAvVAOwGdK0AbVAV2J4OmAIIcAtiADkABmmk0Exnh6YA8mwDCCAIZtgmGfICUpDgOLIj0gPT2Hjp85eu37j59sBSWaPu%2B/rYA1FpMAJ5CwAgEwRDIJsEATLK0AKzBALJ4yMSoFgBmsWHEjCS6LOwAdMGiPDzBUTEcjZgixABumOhVgQF%2B9sEAKgjCwWhYwWM8OZic3cH8bFjEwQQImFkAkkPBADKz83GZOya9Az4XwwAS2MEAyuoAYkMA6qIASndb98EACh91AA1LYAEWwoOC3iSSVEvx%2B0KSpGCrx213UAFVdu8Ph9RAA5IYATWCz1q%2BJJAGktvjQcjsAANAHYe6/dQfPohLaZP57LYQ5E0rR7DGgmkAcWCACEscF8epdnyTkMIcN1Dc7ji8YT%2BWynllsB8tNcCUNRFKtnziaROcEnjt8Szfk92bV/p8hlstBi9p9/hiPn91Pc7gTIfL8TSnh8JdhMthCTUaXL1dggQndvcTXs9hrbaIseiPmyPqF1H8iTHxdddui9uDi9K7nzzXs7i7SwSScLRNzkaDRJlROKWaTSwrbhyri2LVaicjXrchpPgsmCbUtJ71PjSfqtNuhnjN8ihuzdq6J4bUSHkZ8fhK7YDMjarpjz/rXcn9/jHZuttuUTRDUHmeN5PjuV1lzuDEQzHUkoNLcFRD5fFxXhfFbSgkCXhxbBziCTkvCI4iSNI/pRECaEAGY8AKZZMAKYI0L2AB9b9D1ETd7hY65KKSKisAKPA2E2Zi2IPI8hm43i/D4mi2GQHh%2BEmbwqK0TAAA9rGYVg2FU7A5OExTlM2VStA4AgsC0zBmH0yiLkwjZiE2XRnLWBBnM2NA2CE4BBAqXTgiYSpOGCAoSHcsZmKmZQ1AkOYCAC9gwGkFoACNMD0DpWGIW0AHcNjYYJGGc2wygs8Z2HwEKWnYdzRPuXMCHCRg2mCVzNmy1AeAqboQE5W1aBqMSVQ%2BE58VEFUWO3cTCUkliQXUX0t3xEBggs9AQBAAhMGICRhN64I8tUBoMvGXR6gWCBBN0fgeAIExbSSYbGpY5dAVeab8VmjjNwW/9lv/Vb2uCYAOB4LaCl0VQWKhirNO0kKjpO4IzvWPI8r0q4qJe1iMUdfEOy0CEvp%2B%2BbFsB7c1rYVB1mE4Bgj0RhWs4AbZP42jgjASBBOE7oIBG64PtJ9jyYByagcSbwAHYtBlrRgl5kT0AF17RvGybsBFiTOKGf6lol7cpdl%2BXgihNI5b8M3rcVhi%2BZVsT8YTImSZm0Xdf1yn8RMR6reopWGtY9WaU17W5o9inDYw9mqLmfACnsij%2BIDpjXvuQ8JS2J4iRY8EzUtCANKl/iNLklOxPTmNUKzolC8SCuM%2Br7Pc%2BwfO9jrxPqM5gOVZYljFP0YA%2B%2BNuXZdtoTlYgPvxXxDEtGHwIoWTu2RNT1i%2BUpbA9lrovgj7tL%2BFUFg2BYzTWuQWIIG57m6%2BRWhfdEJeBJXwOWPxjet53xJ98P%2B7hNPjS59L7X0gEXZEsgH7UUwDwEQi9/YvzXixD%2B2865xBAXXB%2BT9y6vXflsTeKDd5X2vhguScdaKJ3IoMMScJ7gYnjBAEq3Qci9QfpQkIQxoENHWJsYKeB9oAC9druQqMVZy%2BBkCHQkPwCqCBuo9FXLEMYSwODn1ongBYBVMDcNWBFGmBBbTcNEUwiRO0pgtF0BffgF0eDhGCJgDoF0rE7R6GzJONECiCT3pke4890xTkfvAier8aF0OwPEKq38WK6A4Bwfg8VwnLEgfxaB3cX6937j1AwC8/bLyCYgkJ9DvLoEiQfI%2B/9omxPiUUpJscZiMR7lPFiM857ZICbkvm%2BTWShIScUuIPSpZUX1BZAKyB%2B7RP0WpDoqA8DoH0hACBwQ1o/zKSfJYzkLF6DSmoCAPtSEwMwHA9pq9qFdMKRE9aiUWCjIkRZMyUyZlzPQdUmpZCE6yQclcZiVQKm7QIGtH5xAaquMCGwXQ8UVEWM2GDHglFpZSkCBZYg/AL5hWhjwWG4zFnFX4FsnIFzNogCRWwFg8VT7EDyLlWScLF7WzPjMZAhBUUwzhgQeIehVjeQsgAKkqpwWI4KOC6EMIkNaG0tpEpJZgMlFKIACqFZgY28LpaglhSqqi8KqUqveW41JQSHZqyFuoT6bsdZ/UjitSBOTn55LEu7P6xpsBaEpBAZqrVkTVJtp6r13qfW%2Br9f67wFsaVm1VuvPBn9%2BlQkGRckZYzbmTOmbMqiYSFlrX9emjNmazaBsto/b16NUB5VBuDSGaKMUWQgIiZia1ERrBaqZGETKlJuREbWvu9o2x9yXkkGtjaG5V3FDXRpKEtYsR9patxKT6lpNDW9Q0GspomvDma8WFrE5YIQba01esHVOpdfW915ys3HpPQGoNVtrazuQbXD1qkhmXJyHGiZWh7lJpTSK/FW0dp7QOjtHZE6oF1PHvbWdTtCbsmJqCMOv09bmsluuwJHSt3Lp3bcPdrrMCHsSZ02hZzEmd2Scsch2rEOr2wIAzAF8OCRuQzBtiaHnWIj%2BM5IphBdKIiw8UsuCDsCcEEG0Gjr07WocdYxmEfx8gEFYyFDjvKuMkcI/HYIVChPbukonUF4LGCQuLTCqlGqCJXD0SxNgd09OGaCMED4bQCBCCoyDMowlv1BVWJIVysRHPEqEUQIKbAbGMzkcEGmbAAC0pn6jBAcUpNo%2BE2HBGuNE4Igi8jrTwIIoKXRiAbF0OgDLQihixcuJZrYjFdhjF0MVaZXnVgQDwFUTANRdiqTVZCDEXLEjcKxpZkL7V6iFt5Yi5FSMCh5AkMENrtoev4EFX1otnLbNDcCiN1AY3ws8GYMQFi%2BirhTeENY/rN07qxHm0ii%2B7Gdt2L425E7i26rLdW2Z8bPKoaqA4JN3rfAi18MYPSxl3lMscECj5ibHzRA7ViodMyA8YnDDsn4aHLRjNrdhQZxgOL6X9QvWbYZVyokxN%2BRAMVIBhB48B8AUFWyDlqSGNCAAbMiIna2Ntbf0ltKL/BMPdt2BItgejUYuRiXgcnCwme2d6DCR66r13W3Bz9yHakMOac2BiZEiuwWbFBHEiQtjmufpAHMXQlOWK0RZ2pInJP/u/LwJTsyKvYfJrZ44qnBlk3BoGztQBqwkdmYgBiOndPFgfuZ4TyykMSB5Vckm70cz%2BA%2ByhHC4I5GgHUeZ1zMAuvRfEEl1KePWqKJY7WJgCHpizJq/iuN1X9alfBE1xIbXUbIRE4N0bk3z7zccH7uwTLLAbdqTtzTh3IB2fO7h3ms282z5e9piZn382gvT%2BR33uySRadBW0R%2B739QIC0yclUQwrK48yyVXnt3m%2BeAJIqmfgPO/doDMhId%2B6qkDM2yvyvoKrViAVBIM1i/sRX%2Br5vyzwb3Hlukfyl21Rtgn092GFBi0R2XdzdyPzdxtl41iWchT1sxYjTwzzMw22z2QOtmcgIEECKmZyf0QOVWl3H3YAskn3f12i/1WCGHgLnyP1aGIOICKn3x2Sf1zzdygJKhgKYAYKIGIBCzmUSFYITyIJINgIP14JlhPwL0GEYS6GJRaCUBUB6iRk0SKgsHinYBcgIFlwIBaB8wqSFyKgqzW0qycyEXm30G2zH3nwIBny30Z1wMwIemAKwDUB2nIIL3/3oM/1EJ/w8PqGZ28N118K0SpwMzd0GCWAjy81y2EJCJIA4CqCyM82/Vqj81sRyLWFQFSwMDUCCjSioEo1MO5lPwXzM2vw/0YOhHhSSClHgOiOgViICOcKCLSMYJCxC3aKonv06P8PAJ6LqIiz6NCNaOaNq2JTv1tj8LiNqNcJsOmJIAGPmKiOGKWK6PGJf0mJhTfw2NylmOGJDwJQ23jgKC20WJiLGOf2tl6MaNEPEIuKJ2uNoluJ2JGOWO6JtlfWCKaLSHhTSFBEuK2i%2BIKB%2BMkJoNiA6P%2BPGMCBKjwAcR2kx2cN2DIORMoORKuAkBIE2CSMcIWFOLez8Fl20NMjUgR2GHdR6hhz91dz8E%2BNsxuK22BLeN/xcLcJhWp2X3p3d15MX2j1dxX3uNGJWMpKLzlxL1pMZJaAHwMj8CvwFOTS5M2J5LVM3EFIZ1DxAGhNhOzV2IeOlLB1lOpKjS0DpOVJBSOLMmVM1LOLaLnx1LtJX31KuPZO%2BLuJ8KlO6KpPlxtMVPtxVNEHdP0mdLmLZOIA5IICw0vwdPVIMglP9KRPiM1WUxCCVwhWsF00TmhS2h1Kcy5SjJ2gshYmcjyk/0YGgBLRABLOJTLI1MYEVTdxkM4OKj3zgPwM1UCGkDMAhmkDSAUDYBkFoAUFQBkDllmNaPWksHzOoknNID%2BTkEHLMAAGsQAkg0hSBhyAAWMc4wKcmQBQLgWQVck8swOAWAJATQ1QXacgbgFbLQ3aFAHgWgAAThYlpwPNICEnul2i4AgDShPMUBW3imJU0BsXAvwBYxYC6C4HXPIC82HNMAYB0hoJkFsDFWQGaOaOpIqlsGwGuA4GQGU3UA0mU1cmQCQFECBAZGU1eCPKHJkFHNIHHJQunOkGXORnWGCA0gAA5acQs/zQhcxvyrMtAFA1zTAtydy9zDzjzuKzzzAQBLy5LBzSBbzEAUBXzHyyAKBwllBDKQBgBpYOBQVGAOBZEEzALcjKAwKUKlAoKCAYLwg4K8AEK0S2hwK7D0LtLeFsLpAQsNpmt8K5ypRaB%2BFggQt1AWU4qJBaKEBmtQVELNgQtXhrE4rXhtLhyOKuL5AGAZA%2BLjoBLhLRLxLgBkAKLpY4grLdAbK7KTBZLryFKkgDyqgvyereq%2Bq%2Br6dlLOLwKeLzyNKrz1ybyYA9KHy1AjKXzTK5qPzB4tsiVNyAKj5gKnLwLXKEoPKvKfKkL/K0LrzMKaoZAwrLIIqCLWiYq4qErMUQtkqrBUrhj0rfLcqcqsr8r2KVLiqeKyrCAEBBKRKxKDzxhMkGYXU1rWqJr5LSBtzdz9yZAjzhrVLpAxrNL2rkbpAkg/rTyMa4aNzSAAddIQADygA%3D)). The compilers are smart enough to only validate the pointer value once. Now change the `Ensures` to a `GSL_ASSUME` ([Demo](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAKxAEZTUAHAvVAOwGdK0AbVAV2J4OmAIIcAtiADkABmmk0Exnh6YA8mwDCCAIZtgmGfICUpDgOLIj0gPT2Hjp85eu37j59sBSWaPu%2B/rYA1FpMAJ5CwAgEwRDIJsEATLK0AKzBALJ4yMSoFgBmsWHEjCS6LOwAdMGiPDzBUTEcjZgixABumOhVgQF%2B9sEAKgjCwWhYwWM8OZic3cH8bFjEwQQImFkAkkPBADKz83GZOya9Az4XwwAS2MEAyuoAYkMA6qIASndb98EACh91AA1LYAEWwoOC3iSSVEvx%2B0KSpGCrx213UAFVdu8Ph9RAA5IYATWCz1q%2BJJAGktvjQcjsAANAHYe6/dQfPohLaZP57LYQ5E0rR7DGgmkAcWCACEscF8epdnyTkMIcN1Dc7ji8YT%2BWynllsB8tNcCUNRFKtnziaROcEnjt8Szfk92bV/p8hlstBi9p9/hiPn91Pc7gTIfL8TSnh8JdhMthCTUaXL1dggQndvcTXs9hrbaIseiPmyPqF1H8iTHxdddui9uDi9K7nzzXs7i7SwSScLRNzkaDRJlROKWaTSwrbhyri2LVaicjXrchpPgsmCbUtJ71PjSfqtNuhnjN8ihuzdq6J4bUSHkZ8fhK7YDMjarpjz/rXcn9/jHZuttuUTRDUHmeN5PjuV1lzuDEQzHUkoNLcFRD5fFxXhfFbSgkCXhxbBziCTkvCI4iSNI/pRECaEAGY8AKZZMAKYI0L2AB9b9D1ETd7hY65KKSKisAKPA2E2Zi2IPI8hm43i/D4mi2GQHh%2BEmbwqK0TAAA9rGYVg2FU7A5OExTlM2VStA4AgsC0zBmH0yiLkwjZiE2XRnLWBBnM2NA2CE4BBAqXTgiYSpOGCAoSHcsZmKmZQ1AkOYCAC9gwGkFoACNMD0DpWGIW0AHcNjYYJGGc2wygs8Z2HwEKWnYdzRPuXMCHCRg2mCVzNmy1AeAqboQE5W1aBqMSVQ%2BE58VEFUWO3cTCUkliQXUX0t3xEBggs9AQBAAhMGICRhN64I8tUBoMvGXR6gWCBBN0fgeAIExbSSYbGpY5dAVeab8VmjjNwW/9lv/Vb2uCYAOB4LaCl0VQWKhirNO0kKjpO4IzvWPI8r0q4qJe1iMUdfEOy0CEvp%2B%2BbFsB7c1rYVB1mE4Bgj0RhWs4AbZP42jgjASBBOE7oIBG64PtJ9jyYByagcSbwAHYtBlrRgl5kT0AF17RvGybsBFiTOKGf6lol7cpdl%2BXgihNI5b8M3rcVhi%2BZVsT8YTImSZm0Xdf1yn8RMR6reopWGtY9WaU17W5o9inDYw9mqLmfACnsij%2BIDpjXvuQ8JS2J4iRY8EzUtCANKl/iNLklOxPTmNUKzolC8SCuM%2Br7Pc%2BwfO9jrxPqM5gOVZYljFP0YA%2B%2BNuXZdtoTlYgPvxXxDEtGHwIoWTu2RNT1i%2BUpbA9lrovgj7tL%2BFUFg2BYzTWuQWIIG57m6%2BRWhfdEJeBJXwOWPxjet53xJ98P%2B7hNPjS59L7X0gEXZEsgH7UUwDwEQi9/YvzXixD%2B2865xBAXXB%2BT9y6vXflsTeKDd5X2vhguScdaKJ3IoMMScJ7gYnjBAEq3Qci9QfpQkIQxoENHWJsYKeB9oAC9druQqMVZy%2BBkCHQkPwCqCBuo9FXLEMYSwODn1ongBYBVMDcNWBFGmBBbTcNEUwiRO0pgtF0BffgF0eDhGCJgDoF0rE7R6GzJONECiCT3pke4890xTkfvAier8aF0OwPEKq38WK6A4Bwfg8VwnLEgfxaB3cX6937j1AwC8/bLyCYgkJ9DvLoEiQfI%2B/9omxPiUUpJscZiMR7lPFiM857ZICbkvm%2BTWShIScUuIPSpZUX1BZAKyB%2B7RP0WpDoqA8DoH0hACBwQ1o/zKSfJYzkLF6DSmoCAPtSEwMwHA9pq9qFdMKRE9aiUWCjIkRZMyUyZlzPQdUmpZCE6yQclcZiVQKm7QIGtH5xAaquMCGwXQ8UVEWM2GDHglFpZSkCBZYg/AL5hWhjwWG4zFnFX4FsnIFzNogCRWwFg8VT7EDyLlWScLF7WzPjMZAhBUUwzhgQeIehVjeQsgAKkqpwWI4KOC6EMIkNaG0tpEpJZgMlFKIACqFZgY28LpaglhSqqi8KqUqveW41JQSHZqyFuoT6bsdZ/UjitSBOTn55LEu7P6xpsBaEpBAZqrVkTVJtp6r13qfW%2Br9f67wFsaVm1VuvPBn9%2BlQkGRckZYzbmTOmbMqiYSFlrX9emjNmazaBsto/b16NUB5VBuDSGaKMUWQgIiZia1ERrBaqZGETKlJuREbWvu9o2x9yXkkGtjaG5V3FDXRpKEtYsR9patxKT6lpNDW9Q0GspomvDma8WFrE5YIQba01esHVOpdfW915ys3HpPQGoNVtrazuQbXD1qkhmXJyHGiZWh7lJpTSK/FW0dp7QOjtHZE6oF1PHvbWdTtCbsmJqCMOv09bmsluuwJHSt3Lp3bcPdrrMCHsSZ02hZzEmd2Scsch2rEOr2wIAzAF8OCRuQzBtiaHnWIj%2BM5IphBdKIiw8UsuCDsCcEEG0Gjr07WocdYxmEfx8gEFYyFDjvKuMkcI/HYIVChPbukonUF4LGCQuLTCqlGqCJXD0SxNgd09OGaCMED4bQCBCCoyDMowlv1BVWJIVysRHPEqEUQIKbAbGMzkcEGmbAAC0pn6jBAcUpNo%2BE2HBGuNE4Igi8jrTwIIoKXRiAbF0OgDLQihixcuJZrYjFdhjF0MVaZXnVgQDwFUTANRdiqTVZCDEXLEjcKxpZkL7V6iFt5Yi5FSMCh5AkMENrtoev4EFX1otnLbNDcCiN1AY3ws8GYMQFi%2BirhTeENY/rN07qxHm0ii%2B7Gdt2L425E7i26rLdW2Z8bPKoaqA4JN3rfAi18MYPSxl3lMscECj5ibHzRA7ViodMyA8YnDDsn4aHLRjNrdhQZxgOL6X9QvWbYZVyokxN%2BRAMVIBhB48B8AUFWyDlqSGNCAAbMiIna2Ntbf0ltKL/BMPdt2BItgejUYuRiXgcnCwme2d6DCR66r13W3Bz9yHakMOac2BiZEiuwWbFBHEiQtjmufpAHMXQlOWK0RZ2pInJP/u/LwJTsyKvYfJrZ44qnBlk3BoGztQBqwkdmYgBiOndPFgfuZ4TyykMSB5Vckm70cz%2BA%2ByhHC4I5GgHUeZ1zMAuvRfEEl1KePWqKJY7WJgCHpizJq/iuN1X9alfBE1xIbXUbIRE4N0bk3z7zccH7uwTLLAbdqTtzTh3IB2fO7h3ms282z5e9piZn382gvT%2BR33uySRadBW0R%2B739QIC0yclUQwrK48yyVXnt3m%2BeAJIqmfgPO/doDMhId%2B6qkDM2yvyvoKrViAVBIM1i/sRX%2Br5vyzwb3Hlukfyl21Rtgn092GFBi0R2XdzdyPzdxthOVwzCVT25gzzMw22z2QOtmcgIEECKmZyf0QOVWl3H3YAskn3f12i/1WCGHgLnyP1aEIOICKn3x2Sf1zzdygJKhgKYDoKIGIBCzmUSGYITwIKINgIP24JlhPwL0GEYS6GJRaCUBUB6iRk0SKgsHinYBcgIFlwIBaB8wqSFyKgqzW0qycyEXm30G2zH3nwIBny30Z2wNsy2zv1tjUB2lIIL3/1oM/2EJ/zcPqGZwemAKwB8KpwMzd0GCWAjy81y0EKCJIA4CqAyM82/Vqj81sSyLWFQFSwMDUCCjSioEo2MMwP8IXzM2vw/3oOhHhSSClHgN1yiK0RiNPxqPqDqKEJIBCxC1aKonv2gQ6L8McLP0CIaOaMaNq2JS8PaN8PAImO6IaBSPoIGLmIiLaNGKWOf2tgCPWOCJmOGJDwJQ23jgKE8MiN2M6OqOcORzfyOP6JCNDxAAuNoiuO2OGO8LGOWJtlfSmOCLSHhTSFBDOK2g%2BIKC%2BPEKoNiB2OiPGMCBKjwAcR2kx0cN2BIOWPkKRKuAkBIE2ASPsIWGeOIDez8Fl00NMjUgR2GHdR6hhz91dz8CJyhK%2BKBP6N/ycJcJhWp2X3p3dx5MX2j1dxXwWNuPGKpPly0DpIHwMj8Cv35OTU5JEO5KVM3AFIZzePZOuJuMROWOlJL1pMZJaHlJBVWLMnlNVNmLnw1PNJX21PONs0uOuIRL%2BIMyNJpNlNNPtwVNEHtP0htOaIhPeJdM%2BK2yw0v0tOVIMnFP1I9M7mVWUxCCVwhWsF00TmhS2g1Kcy5SDJ2gshYmcjyk/0YGgBLRAFzOJXzJVMYEVTdykPYOKj3zgNwM1UCGkDMAhmkDSAUDYBkFoAUFQBkDlhmOaPWksAzOoiHNID%2BTkC7LMAAGsQAkg0hSAeyAAWfs4wYcmQBQLgWQOc3cswOAWAJAdQ1QXacgbgFbDQ3aFAHgWgAAThYlp03NICEnul2i4AgDSl3MUBW3imJU0BsQAvwBYxYC6C4AXPIC8x7NMAYB0ioJkFsDFWQEaMaOpIqlsGwGuA4GQGU3UA0mU1cmQCQFECBAZGU1eG3O7JkD7NIAHNgpHOkBnORnWGCA0gAA5acQt3zQhcwXyrMtAFB5zTBlzVz1ytydyWL9zzAQAjzxKuzSAzzEAUA7yryyAKBwllAtKQBgBpYOBQVGAOBZECBPyj4fzKB/zYKlBgKCBQLwhwK8BILUS2gAKbCEKVLeEULpAQsNpmsMLxypRaB%2BFggQt1AWUIqJAyKEBmtQUoLNgQtXhrEIrXgVKezGLmL5AGAZB2LjpOKeK%2BKBLgBkBCLpY4hjLdBTLzKTAxKTzJKkhNyqhny2r2qOqOr6cZKmKALWKDzFLjyFzTyYB1LLy1BtLby9KJrHzB4tsiUlzLLvzySbKAL7KEonKXK3LoLPL4KTykKaoZAArLIgrMLmiwqIqorMUQtYqrB4rhjEr3L0q0qUrMqGLZLcrWKCrCAEAuLeL%2BLNzxhMkGYXUFr6qhqJLSAVy1yNyZBtzeq5LpABqlLGrYbpAkgPq9ykaIbFzSAAddIQBNygA%3D)). We've successfully informed at least MSVC and GCC of the class invariant, and they realize that the pointer needs no validation in `test_rewrap`. (Oddly enough, Clang does not. But if we change `test_rewrap` to accept `gsl::not_null<int*> const&` ([Demoo](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAKxAEZTUAHAvVAOwGdK0AbVAV2J4OmAIIcAtiADkABmmk0Exnh6YA8mwDCCAIZtgmGfICUpDgOLIj0gPT2Hjp85eu37j59sBSWaPu%2B/rYA1FpMAJ5CwAgEwRDIJsEATLK0AKzBALJ4yMSoFgBmsWHEjCS6LOwAdMGiPDzBUTEcjZgixABumOhVgQF%2B9sEAKgjCwWhYwWM8OZic3cH8bFjEwQQImFkAkkPBADKz83GZOya9Az4XwwAS2MEAyuoAYkMA6qIASndb98EACh91AA1LYAEWwoOC3iSSVEvx%2B0KSpGCrx213UAFVdu8Ph9RAA5IYATWCz1q%2BJJAGktvjQcjsAANAHYe6/dQfPohLaZP57LYQ5E0rR7DGgmkAcWCACEscF8epdnyTkMIcN1Dc7ji8YT%2BWynllsB8tNcCUNRFKtnziaROcEnjt8Szfk92bV/p8hlstBi9p9/hiPn91Pc7gTIfL8TSnh8JdhMthCTUaXL1dggQndvcTXs9hrbaIseiPmyPqF1H8iTHxdddui9uDi9K7nzzXs7i7SwSScLRNzkaDRJlROKWaTSwrbhyri2LVaicjXrchpPgsmCbUtJ71PjSfqtNuhnjN8ihuzdq6J4bUSHkZ8fhK7YDMjarpjz/rXcn9/jHZuttuUTRDUHmeN5PjuV1lzuDEQzHUkoNLcFRD5fFxXhfFbSgkCXhxbBziCTkvCI4iSNI/pRECaEAGY8AKZZMAKYI0L2AB9b9D1ETd7hY65KKSKisAKPA2E2Zi2IPI8hm43i/D4mi2GQHh%2BEmbwqK0TAAA9rGYVg2FU7A5OExTlM2VStA4AgsC0zBmH0yiLkwjZiE2XRnLWBBnM2NA2CE4BBAqXTgiYSpOGCAoSHcsZmKmZQ1AkOYCAC9gwGkFoACNMD0DpWGIW0AHcNjYYJGGc2wygs8Z2HwEKWnYdzRPuXMCHCRg2mCVzNmy1AeAqboQE5W1aBqMSVQ%2BE58VEFUWO3cTCUkliQXUX0t3xEBggs9AQBAAhMGICRhN64I8tUBoMvGXR6gWCBBN0fgeAIExbSSYbGpY5dAVeab8VmjjNwW/9lv/Vb2uCYAOB4LaCl0VQWKhirNO0kKjpO4IzvWPI8r0q4qJe1iMUdfEOy0CEvp%2B%2BbFsB7c1rYVB1mE4Bgj0RhWs4AbZP42jgjASBBOE7oIBG64PtJ9jyYByagcSbwAHYtBlrRgl5kT0AF17RvGybsBFiTOKGf6lol7cpdl%2BXgihNI5b8M3rcVhi%2BZVsT8YTImSZm0Xdf1yn8RMR6reopWGtY9WaU17W5o9inDYw9mqLmfACnsij%2BIDpjXvuQ8JS2J4iRY8EzUtCANKl/iNLklOxPTmNUKzolC8SCuM%2Br7Pc%2BwfO9jrxPqM5gOVZYljFP0YA%2B%2BNuXZdtoTlYgPvxXxDEtGHwIoWTu2RNT1i%2BUpbA9lrovgj7tL%2BFUFg2BYzTWuQWIIG57m6%2BRWhfdEJeBJXwOWPxjet53xJ98P%2B7hNPjS59L7X0gEXZEsgH7UUwDwEQi9/YvzXixD%2B2865xBAXXB%2BT9y6vXflsTeKDd5X2vhguScdaKJ3IoMMScJ7gYnjBAEq3Qci9QfpQkIQxoENHWJsYKeB9oAC9druQqMVZy%2BBkCHQkPwCqCBuo9FXLEMYSwODn1ongBYBVMDcNWBFGmBBbTcNEUwiRO0pgtF0BffgF0eDhGCJgDoF0rE7R6GzJONECiCT3pke4890xTkfvAier8aF0OwPEKq38WK6A4Bwfg8VwnLEgfxaB3cX6937j1AwC8/bLyCYgkJ9DvLoEiQfI%2B/9omxPiUUpJscZiMR7lPFiM857ZICbkvm%2BTWShIScUuIPSpZUX1BZAKyB%2B7RP0WpDoqA8DoH0hACBwQ1o/zKSfJYzkLF6DSmoCAPtSEwMwHA9pq9qFdMKRE9aiUWCjIkRZMyUyZlzPQdUmpZCE6yQclcZiVQKm7QIGtH5xAaquMCGwXQ8UVEWM2GDHglFpZSkCBZYg/AL5hWhjwWG4zFnFX4FsnIFzNogCRWwFg8VT7EDyLlWScLF7WzPjMZAhBUUwzhgQeIehVjeQsgAKkqpwWI4KOC6EMIkNaG0tpEpJZgMlFKIACqFZgY28LpaglhSqqi8KqUqveW41JQSHZqyFuoT6bsdZ/UjitSBOTn55LEu7P6xpsBaEpBAZqrVkTVJtp6r13qfW%2Br9f67wFsaVm1VuvPBn9%2BlQkGRckZYzbmTOmbMqiYSFlrX9emjNmazaBsto/b16NUB5VBuDSGaKMUWQgIiZia1ERrBaqZGETKlJuREbWvu9o2x9yXkkGtjaG5V3FDXRpKEtYsR9patxKT6lpNDW9Q0GspomvDma8WFrE5YIQba01esHVOpdfW915ys3HpPQGoNVtrazuQbXD1qkhmXJyHGiZWh7lJpTSK/FW0dp7QOjtHZE6oF1PHvbWdTtCbsmJqCMOv09bmsluuwJHSt3Lp3bcPdrrMCHsSZ02hZzEmd2Scsch2rEOr2wIAzAF8OCRuQzBtiaHnWIj%2BM5IphBdKIiw8UsuCDsCcEEG0Gjr07WocdYxmEfx8gEFYyFDjvKuMkcI/HYIVChPbukonUF4LGCQuLTCqlGqCJXD0SxNgd09OGaCMED4bQCBCCoyDMowlv1BVWJIVysRHPEqEUQIKbAbGMzkcEGmbAAC0pn6jBAcUpNo%2BE2HBGuNE4Igi8jrTwIIoKXRiAbF0OgDLQihixcuJZrYjFdhjF0MVaZXnVgQDwFUTANRdiqTVZCDEXLEjcKxpZkL7V6iFt5Yi5FSMCh5AkMENrtoev4EFX1otnLbNDcCiN1AY3ws8GYMQFi%2BirhTeENY/rN07qxHm0ii%2B7Gdt2L425E7i26rLdW2Z8bPKoaqA4JN3rfAi18MYPSxl3lMscECj5ibHzRA7ViodMyA8YnDDsn4aHLRjNrdhQZxgOL6X9QvWbYZVyokxN%2BRAMVIBhB48B8AUFWyDlqSGNCAAbMiIna2Ntbf0ltKL/BMPdt2BItgejUYuRiXgcnCwme2d6DCR66r13W3Bz9yHakMOac2BiZEiuwWbFBHEiQtjmufpAHMXQlOWK0RZ2pInJP/u/LwJTsyKvYfJrZ44qnBlk3BoGztQBqwkdmYgBiOndPFgfuZ4TyykMSB5Vckm70cz%2BA%2ByhHC4I5GgHUeZ1zMAuvRfEEl1KePWqKJY7WJgCHpizJq/iuN1X9alfBE1xIbXUbIRE4N0bk3z7zccH7uwTLLAbdqTtzTh3IB2fO7h3ms282z5e9piZn382gvT%2BR33uySRadBW0R%2B739QIC0yclUQwrK48yyVXnt3m%2BeAJIqmfgPO/doDMhId%2B6qkDM2yvyvoKrViAVBIM1i/sRX%2Br5vyzwb3Hlukfyl21Rtgn092GFBi0R2XdzdyPzdxthOVwzCVT25gzzMw22z2QOtmcgIEECKmZyf0QOVWl3H3YAskn3f12i/1WCGHgLnyP1aEIOICKn3x2Sf1zzdygJKhgKYDoKIGIBCzmUSGYITwIKINgIP24JlhPwL0GEYS6GJRaCUBUB6iRk0SKgsHinYBcgIFlwIBaB8wqSFyKgqzW0qycyEXm30G2zH3nwIBny30Z2wNsy2zv1tjUB2lIIL3/1oM/2EJ/zcPqGZwemAKwB8KpwMzd0GCWAjy81y0EKCJIA4CqAyM82/Vqj81sSyLWFQFSwMDUCCjSioEo2MMwP8IXzM2vw/3oOhHhSSClHgN1yiK0RiNPxqPqDqKEJIBCxC1aKonv2gQ6L8McLP0CIaOaMaNq2JS8PaN8PAImO6IaBSPoIGLmIiLaNGKWOf2tgCPWOCJmOGJDwJQ23jgKE8MiN2M6OqOcORzfyOP6JCNDxAAuNoiuO2OGO8LGOWJtlfSmOCLSHhTSFBDOK2g%2BIKC%2BPEKoNiB2OiPGMCBKjwAcR2kx0cN2BIOWPkKRKuAkBIE2ASPsIWGeOIDez8Fl00NMjUgR2GHdR6hhz91dz8CJyhK%2BKBP6N/ycJcJhWp2X3p3dx5MX2j1dxXwWNuPGKpPly0DpIHwMj8Cv35OTU5JEO5KVM3AFIZzePZOuJuMROWOlJL1pMZJaHlJBVWLMnlNVNmLnw1PNJX21PONs0uOuIRL%2BIMyNJpNlNNPtwVNEHtP0htOaIhPeJdM%2BK2yw0v0tOVIMnFP1I9M7mVWUxCCVwhWsF00TmhS2g1Kcy5SDJ2gshYmcjyk/0YGgBLRAFzOJXzJVPmwD0YEVTdykPYOKj3zgNwM1UCGkDMAhmkDSAUDYBkFoAUFQBkDlhmOaPWksAzOohHNID%2BTkB7LMAAGsQAkg0hSA%2ByAAWQc4wUcmQBQLgWQBc/cswOAWAJAdQ1QXacgbgFbDQ3aFAHgWgAAThYlp23NICEnul2i4AgDSn3MUBW3imJU0BsSAvwBYxYC6C4CXPIC8z7NMAYB0ioJkFsDFWQEaMaOpIqlsGwGuA4GQGU3UA0mU1cmQCQFECBAZGU1eF3N7JkAHNICHPgrHOkDnORnWGCA0gAA5acQtPzQhcw3yrMtAFBFzTBVz1zNydy9y2LDzzAQATzJKezSALzEAUAHybyyAKBwllAdKQBgBpYOBQVGAOBZECBvyj4/zKBAL4KlBQKCBwLwhIK8BoLUS2ggKbCkK1LeE0LpAQsNpmssLJypRaB%2BFggQt1AWUoqJAKKEBmtQUYLNgQtXhrEorXg1K%2BzmLWL5AGAZBOLjpuK%2BKBKhLgBkBiLpY4hTLdBzLLKTAJKzzpKkhtyqhXyOrOququr6c5KWKgL2KjzlLTylzzyYBNLry1BdL7yDKprnzB4tsiUVzrLfzyS7KgLHKEoXK3KPLYLvLEKzyUKaoZAgrLIQrsLmiIqoqYrMUQt4qrBErhjkrPLMqMq0rsqmL5L8r2KirCAEAeL%2BLBLtzxhMkGYXUlrGqRqpLSA1yNytyZBdz%2BqFLpAhqVLmr4bpAkgvqDyUaoblzSAAddIQBtygA%3D)) then clang won't revalidate but MSVC will. \*shrug\* optimizers.)

gdr-at-ms:
That is a good test to drive compiler writer to do their job – not the other way around :-)
Thanks for producing this.  Could I convince you to fill at least a bug against MSVC for better support for “Idiomatic C++”?

neilmacintosh:
Thanks @CaseyCarter for making the changes that were requested. This batch of cleanup and improvements for `not_null` looks good to me.

And thanks to @gdr-at-ms for saying all the things I was thinking...just more concisely ;-)

PR #450, opened 2017-02-11
Revert change to the definition of constexpr macro in gsl_byte from #446

CaseyCarter:
I apparently lied when I said I had reverted all of these changes to the `constexpr` macros. Causes warning spew when building with VS2013.


neilmacintosh:
Hahaha. Let's undo the lie!

PR #452, opened 2017-02-13
gsl_assert changes:

CaseyCarter:
* `GSL_LIKELY`/`_UNLIKELY` always contextually convert their argument to `bool`.

* Add macro `GSL_ASSUME(cond)` to inform the optimizer that the predicate `cond` must hold.

* Reimplement `Expects` & `Ensures` as expressions for flexibility and correctness. (Think about `if (cond1) Expects(cond2); else goto fail;`)

* `Expects` and `Ensures` `GSL_ASSUME` their predicate when `GSL_UNENFORCED_ON_CONTRACT_VIOLATION` is defined.

### DIFF
--- a/include/gsl/gsl_assert
+++ b/include/gsl/gsl_assert
@@ -30,20 +30,36 @@
 // 2. GSL_THROW_ON_CONTRACT_VIOLATION: a gsl::fail_fast exception will be thrown
 // 3. GSL_UNENFORCED_ON_CONTRACT_VIOLATION: nothing happens
 //
-#if !(defined(GSL_THROW_ON_CONTRACT_VIOLATION) || defined(GSL_TERMINATE_ON_CONTRACT_VIOLATION) ||    \
+#if !(defined(GSL_THROW_ON_CONTRACT_VIOLATION) || defined(GSL_TERMINATE_ON_CONTRACT_VIOLATION) ||  \
       defined(GSL_UNENFORCED_ON_CONTRACT_VIOLATION))
-#define GSL_TERMINATE_ON_CONTRACT_VIOLATION
+ #define GSL_TERMINATE_ON_CONTRACT_VIOLATION
 #endif
 
 #define GSL_STRINGIFY_DETAIL(x) #x
 #define GSL_STRINGIFY(x) GSL_STRINGIFY_DETAIL(x)
 
 #if defined(__clang__) || defined(__GNUC__)
-#define GSL_LIKELY(x) __builtin_expect (!!(x), 1)
-#define GSL_UNLIKELY(x) __builtin_expect (!!(x), 0)
+ #define GSL_LIKELY(x) __builtin_expect (!!(x), 1)
+ #define GSL_UNLIKELY(x) __builtin_expect (!!(x), 0)
 #else
-#define GSL_LIKELY(x) (x)
-#define GSL_UNLIKELY(x) (x)
+ #define GSL_LIKELY(x) (!!(x))
+ #define GSL_UNLIKELY(x) (!!(x))
+#endif
+
+//
+// GSL_ASSUME(cond)
+//
+// Tell the optimizer that the predicate cond must hold. It is unspecified
+// whether or not cond is actually evaluated.
+//
+#ifdef _MSC_VER
+ #define GSL_ASSUME(cond) __assume(cond)
+#elif defined(__clang__)
+ #define GSL_ASSUME(cond) __builtin_assume(cond)
+#elif defined(__GNUC__)
+ #define GSL_ASSUME(cond) ((cond) ? static_cast<void>(0) : __builtin_unreachable())
+#else
+ #define GSL_ASSUME(cond) static_cast<void>(!!(cond))
 #endif
 
 //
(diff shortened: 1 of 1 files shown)


MikeGitb:
Is using `GSL_ASSUME` in `Expects` really a good Idea? 

Under `GSL_UNENFORCED_ON_CONTRACT_VIOLATION` this turns a mechanism that usually increases safety/security (by preventing a contract violation to cause undefined behavior) into something that makes a contract violation even more dangerous (by essentially guaranteeing undefined behavior). In particular it even prevents a second independent check of the condition at another point in the code.

CaseyCarter:
Selecting `GSL_UNENFORCED_ON_CONTRACT_VIOLATION` does not remove the contracts from the program, and it does not magically rewrite the code to not rely on the guarantees those contracts provide. By selecting `GSL_UNENFORCED_ON_CONTRACT_VIOLATION` you explicitly opt in to undefined behavior on contract violation. Putting `GSL_ASSUME` into the contracts does not make the behavior of the program any more or less defined; all it does is communicate to the compiler your choice to allow UB in the presence of contract violations, so the compiler can take advantage of that choice to improve code generation.

MikeGitb:
Ues, but before your patch, not every contract violation necessarily leads to language level UB. Sometimes it might just result in wrong data, sometimes the error might be handled by another mechanism further down the callstack and sometimes the implementation might actually be able to "correctly" deal with inputs that were conservatively not allowed when the interface was designed. 

I don't have a strong opinion on the matter, because I'm lacking experience, but it seems somewhat dangerous to me, if an check of the caller's behavior is replaced by a promise that isn't even visible from the caller's side. Especially as this is a more or less project wide setting. 

Maybe there could be two settings: 
 - Take `Expects` as a hint (use likely)
 - Take `Expects` as a promise (use assume).

gdr-at-ms:
I don't think there needs to be variations of `Expects`.  Note that whether the runtime checks are active or not, the implementation always assumes that the preconditions hold.  

`GSL_UNENFORCED_ON_CONTRACT_VIOLATION` does NOT means "oh, forget function has preconditions".

MikeGitb:
> GSL_UNENFORCED_ON_CONTRACT_VIOLATION does NOT means "oh, forget function has preconditions".

I didn't say that. What I said was that the consequences of violating a contract might become more severe this way than when not using Expects at all.

> Note that whether the runtime checks are active or not, the implementation always assumes that the preconditions hold.
Not sure what you mean by implementation here. AFAIK the compiler doesn't know about contracts yet and I explained above why function implementations might not actually rely on preconditions.

gdr-at-ms:
But that is the whole point of having a runtime check in the first place!  You can't take it away and scream "oh, consequences of violation is more severe now."

I meant the implementation of the function -- not the compiler.

MikeGitb:
@gdr-at-ms: My concern is not that the consequences of violating a contract under `GSL_UNENFORCED_ON_CONTRACT_VIOLATION` are worse than under  `GSL_TERMINATE_ON_CONTRACT_VIOLATION` but that they are even worse than if I wouldn't use `Expects` at all.

Think about the following implementation of `std::memcpy`:

	void* memcpy(void* dest, const void* src, std::size_t count) {
		auto* d = static_cast<char*>(dest);
		auto* s = static_cast<const char*>(src);
		for (size_t i = 0; i < count; ++i) {
			*d++ = *s++;
		}
		return dest;
	}

calling memcpy as `std::memcpy(pd,ps,0);` is - according to the standard - undefined behavior if either `pd` or `ps` are null pointers. So far, my particular implementation however (and AFAIK many real world implementations out there) doesn't rely on that fact in any way and a program that violates that particular contract runs without any trouble. 

Now I introduce contract checking with 

	void* memcpy(void* dest, const void* src, std::size_t count) {
		Expects(dest != nullptr);
		Expects(src != nullptr);

But the user of my function compiles the function with  `GSL_UNENFORCED_ON_CONTRACT_VIOLATION` (e.g. in release mode). Now the compiler can (after inlining) assume that any pointer that is passed to that function is not a nullptr and can eliminate all following (and maybe even previous) nullptr checks on that variable leading to symptoms that are potentially very hard to analyze.

Of course, the program had a bug even before I started using `Expects`, but only after using `Expects` had that bug any consequences. 
Without the changes in this PR, I can at least be sure that - whether or not `GSL_UNENFORCED_ON_CONTRACT_VIOLATION` is used - using `Expects` doesn't introduce any additional undefined behavior on the language-level. With this PR I can't.

Btw.: I hope my remarks didn't feel like screaming - if they did, I apologize. Just because I raise concerns doesn't mean I demand that other people see them as actual problems, but I like to put them on the table, so they can be discussed and either addressed or dismissed.

neilmacintosh:
@MikeGitb I think the argument you've made has been addressed here by both Casey and Gaby and is not found to be convincing.

Contracts are just that: contracts between caller and callee. They exist whether or not they are enforced at runtime. Trying to minimize difficulty for code that chose to break a contract, and also chose to not enforce contracts seems completely at odds with the aim of writing more reliable and robust C++.

gdr-at-ms:
@MikeGitb : That is confused and mistaken.  Confused because you aren't supposed to have "your own" implementation of `std::memcpy` to begin with, if you were doing anything interesting with your program. So that is already undefined behavior. Mistaken because https://news.ycombinator.com/item?id=12002746

neilmacintosh:
Thanks for the PR @CaseyCarter, this is much tidier!

CaseyCarter:
> calling memcpy as std::memcpy(pd,ps,0); is - according to the standard - undefined behavior if either pd or ps are null pointers. So far, my particular implementation however (and AFAIK many real world implementations out there) doesn't rely on that fact in any way and a program that violates that particular contract runs without any trouble.

Not necessarily: if you name this function `std::memcpy`, the compiler is free to assume that the first two arguments always point to objects. That contract is specified in the C standard, whether or not it appears in your implementation is immaterial. The standard does not define the behavior of programs that violate that contract, so callers that did not satisfy the contract had undefined behavior even *before* you added it to the program.

I understand that it looks like I'm nitpicking at the particular example you chose, but in fairness you chose that example *because* it had a pre-existing contract.

PR #453, opened 2017-02-13
Silence warning in algorithm_tests with VS2017

CaseyCarter:
Now that the STL respects /W4, this test that uses `std::copy_n` to copy a span of ints to a span of chars triggers "warning C4244: '=': conversion from 'int' to 'char', possible loss of data". Switch the source & destination spans to `short` and `int` to maintain the test's cross-type nature but without narrowing.

### DIFF
--- a/tests/algorithm_tests.cpp
+++ b/tests/algorithm_tests.cpp
@@ -100,11 +100,11 @@ SUITE(copy_tests)
     {
         // dynamic source and destination span
         {
-            std::array<int, 5> src{1, 2, 3, 4, 5};
-            std::array<char, 10> dst{};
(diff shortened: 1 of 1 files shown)


gdr-at-ms:
Does this work with VS2015 series too?  Support for VS2013 will drop once VS2017 comes online, but VS2015 will persist until it doesn't.

CaseyCarter:
> Does this work with VS2015 series too?

Yes; the CI checks validate everything against VS2015/2013 (Appveyor), Clang3.6/3.7/3.8, and GCC5/6 (Travis).

gdr-at-ms:
Awesome!
Looks good to me.

neilmacintosh:
Nice cleanup @CaseyCarter, thanks!

Issue #451, opened 2017-02-13
length_func<CharT>::operator() fails CppCoreCheck


duncanhorn:
Running CppCoreCheck with VS2017 RC fails with:

    warning C26461: The pointer argument 'ptr' for function gsl::details::length_func<char>::operator() can be marked as a pointer to const. (con.3: https://go.microsoft.com/fwlink/p/?LinkID=786684)
    warning C26461: The pointer argument 'ptr' for function gsl::details::length_func<wchar_t>::operator() can be marked as a pointer to const. (con.3: https://go.microsoft.com/fwlink/p/?LinkID=786684)

Seems like the placement of `const` is incorrect:

    template <>
    struct length_func<char>
    {
        std::ptrdiff_t operator()(char* const ptr, std::ptrdiff_t length) GSL_NOEXCEPT
        {
            return details::string_length(ptr, length);
        }
    };

Should likely be declared as `const char* ptr` or perhaps `const char* const ptr`.

neilmacintosh:
Nice catch @duncanhorn. Would you like to offer a PR?

duncanhorn:
Sure thing. #458

neilmacintosh:
This was closed thanks to the #458 from @duncanhorn.

PR #458, opened 2017-02-28
[Issue #451] Updating length_func<CharT>::operator() to accept pointer-to-const

duncanhorn:
Leaving the arguments as const-pointer since it's not really harming anything. Compilation of tests succeeds & all tests pass (though this change shouldn't have any runtime effects). CppCoreCheck with VS 2017 RC also passes without issue.

### DIFF
--- a/include/gsl/string_span
+++ b/include/gsl/string_span
@@ -217,7 +217,7 @@ namespace details
     template <>
     struct length_func<char>
     {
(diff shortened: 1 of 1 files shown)


neilmacintosh:
Thanks for the PR @duncanhorn! This is a nice bit of clean-up.

PR #460, opened 2017-03-04
Mark headers as C++ code with vim filetype

ranisalt:
This adds syntax highlighting to the Github file browser and might add auto highlighting to some editors, namely vim and derivatives.

Example before: https://github.com/Microsoft/GSL/blob/3819df6e378ffccf0e29465afe99c3b324c2aa70/include/gsl/gsl
Example after: https://github.com/ranisalt/GSL/blob/28d6fd1d50dc9b1e09cdac42301860b481a26315/include/gsl/gsl

As it's not code changes, I didn't open an issue before, but I can do.

### DIFF
--- a/include/gsl/gsl
+++ b/include/gsl/gsl
@@ -144,3 +144,5 @@ struct hash<gsl::not_null<T>>
 #endif // defined(_MSC_VER) && _MSC_VER < 1910
 
 #endif // GSL_GSL_H
+
+// vim: set filetype=cpp:
--- a/include/gsl/gsl_algorithm
+++ b/include/gsl/gsl_algorithm
@@ -57,3 +57,5 @@ void copy(span<SrcElementType, SrcExtent> src, span<DestElementType, DestExtent>
 #endif // _MSC_VER
 
 #endif // GSL_ALGORITHM_H
+
+// vim: set filetype=cpp:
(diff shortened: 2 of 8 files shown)


neilmacintosh:
@ranisalt I appreciate the syntax highlighting and thanks for teaching me that GitHub recognizes the vim filetype comment-tag. However, I'm not sure that I want to start adding comment-tags to the GSL headers to help configure people's tools - that seems like an issue for users to deal with in their own configurations. See #465 for a similar discussion.

Also, I obviously don't want to get drawn into the vi/emacs war inadvertently by appearing to favor either side. :-) (even though I may be known to use vim sometimes...)

ranisalt:
@neilmacintosh I tried to use gitattributes before but Github seems to ignore it's own tool and follow vim modeline instead. 

(your surname is macintosh and this is so funny)

neilmacintosh:
Interesting. I note the GitHub documentation says they will use the `.gitattributes` to recognize the source language for the purpose of gathering statistics, but not doing syntax highlighting. I'm not sure why that would be. Perhaps it would be worth asking them or opening an issue?

(yes, my name has been remarked upon before. especially as I am often one of the few people in a room *without* a MacBook.)

PR #462, opened 2017-03-07
Refactor some duplication out of gsl_byte

pierrechevalier83:


neilmacintosh:
@pierrechevalier83 thanks for the PR, but in this case, it doesn't really add anything to the library. I don't see a bug being fixed (please let me know if I've misread) and I don't see a feature being added. It's simply a matter of style, and in this case, I'm not particularly convinced it is an important stylistic improvement.

PR #464, opened 2017-03-09
add gsl::not_null operator<<

rianquinn:
Hippomocks, which is a popular mocking engine for C++ uses operator<< on pointers and gets confused about gsl::not_null not having this operator.

### DIFF
--- a/include/gsl/gsl
+++ b/include/gsl/gsl
@@ -109,6 +109,13 @@ private:
     T ptr_;
 };
 
+template <class T>
+std::ostream& operator<<(std::ostream& os, const not_null<T>& val)
+{
+    os << val.get();
+    return os;
+}
+
 // more unwanted operators
 template <class T, class U>
 std::ptrdiff_t operator-(const not_null<T>&, const not_null<U>&) = delete;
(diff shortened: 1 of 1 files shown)


include/gsl/gsl
```diff
@@ -109,6 +109,13 @@ private:
     T ptr_;
 };
 
+template <class T>
```
neilmacintosh: You should add `#include <iostream>` in order to use `std::ostream` reliably here.

neilmacintosh:
Thanks for the PR @rianquinn! I have no objection to this, especially as it makes using the type slightly easier. I've added a comment asking for a small tweak.

include/gsl/gsl
rianquinn: oppsss... ha. I have that in my patch and forgot to port it to this one. Sorry about that

MikeGitb: isn't `<ostream>` enough?

neilmacintosh:
Thanks @rianquinn that looks good to me now.

include/gsl/gsl
rianquinn: from what I can tell, ostream seems to be enough. @neilmacintosh would you like me to change that? I made the change locally and it comes and tests fine

PR #465, opened 2017-03-09
register GSL as a system header

rianquinn:
GSL currently causes Clang Tidy to go nuts for several reasons: 
1) there are several checks that fail because of UnitTest++ 
2) there are several checks that fail simply because we need to clean up some issues 
3) the Core Guideline checks clearly fail since we are implementing the GSL

Adding this macro prevents Clang Tidy and other tools (like coverage tools) from registering the GSL as code that belongs to the project.

### DIFF
--- a/include/gsl/gsl
+++ b/include/gsl/gsl
@@ -19,7 +19,9 @@
 #ifndef GSL_GSL_H
 #define GSL_GSL_H
 
+#if defined(__clang__) || defined(__GNUC__)
 #pragma GCC system_header
+#endif
 
 #include <gsl/gsl_assert>  // Ensures/Expects
 #include <gsl/gsl_util>    // finally()/narrow()/narrow_cast()...
--- a/include/gsl/gsl
+++ b/include/gsl/gsl
@@ -19,6 +19,8 @@
 #ifndef GSL_GSL_H
 #define GSL_GSL_H
 
+#pragma GCC system_header
+
 #include <gsl/gsl_assert>  // Ensures/Expects
 #include <gsl/gsl_util>    // finally()/narrow()/narrow_cast()...
 #include <gsl/multi_span>  // multi_span, strided_span...


neilmacintosh:
@rianquinn I'm not sure I like the idea of using a vendor-specific extension to just push off the problems here. Here's my thinking:

If the GSL fails checks for the Core Guidelines in places as it implements core concepts like `span`, then we should add targeted suppressions at the places where it does. 

If the GSL has issues we should clean up, then we should clean them up ;-)

If UnitTest++ has issues to be suppressed/cleaned up then we should do so and contribute them back.

rianquinn:
@neilmacintosh it's actually more than just Clang Tidy. Coveralls also falsely reports the GSL as not being tested because it thinks it's a user defined header file. In the coveralls case I can manually exclude, but it's kind of ugly. 

Sadly the only way to suppress Clang Tidy is to tell it not to check the code at all (similar to CppChecker) which is what this line is doing also. Actually, I think Clang Tidy uses the same comment tag that CppChecker has, an issue we brought up at CppCon last year. 

I guess one solution is to install the GSL into a system header location, which should have the same effect. In my case, since I use a prefix, it's not quiet as clean, but we already maintain other patches to the GSL so this is one we can keep as well if needed. 

https://github.com/Bareflank/GSL/commit/6693fe38f0b3d147902e702aa8c5ddc9b1be5625

I'll leave it up to you. Either way, I'm willing to help with cleaning up the GSL with respect to Clang Tidy as there is a lot of things that it's complaining about.

neilmacintosh:
@rianquinn I think the principle here is that tool configuration is really an issue for users to deal with either in patches/forks or (preferably) local configuration.

I think it is worth making an exception for suppressing checks for Core Guidelines enforcement (for obvious reasons), but enforcement of those rules can be suppressed using a standard C++ annotation-based suppression mechanism (`[[gsl::suppress]]`).

I'd be happy to look at PRs to clean up things that clang-tidy warns about.

PR #370, opened 2016-09-20
Make GSL easier consumed by another project.

sezaru:
This PR is related to this issue: [#363](https://github.com/Microsoft/GSL/issues/363)

Changes:
- Switched old include_directories function for the newer target_include_directories.
- Added GSL as a INTERFACE library
- Added GSL project to the tests CMakeLists.txt so It can include it automatically.

Issue #363, opened 2016-09-16
GSL as sub-project


sezaru:
Currently gsl's CmakeLists.txt doesn't support it's addition as a sub-project from another cmake project, is that by design?

If not, adding something like this:

```
add_library(${PROJECT_NAME} INTERFACE)

target_include_directories(${PROJECT_NAME} INTERFACE ${CMAKE_CURRENT_SOURCE_DIR}/gsl)
```

would make it possible to simply using gsl as follow:

```
add_subdirectory(GSL);

target_link_libraries(${PROJECT_NAME} GSL)
```

MikeGitb:
Just in case this is not clear: 
As this is a header only library, there is nothing you have to build and link to. The cpp files in this repo are just tests. The only thing you have to do to include the gsl in your project is to copy the gsl folder and add the folder to your include paths.

sezaru:
I know that, but this means that I need to copy the folder to my project and losing the source control. so every time I want to update to the master of gsl repository, I would need to clone it and copy it again..

These changes make possible that I can only add the gsl repository as a submodule to my project and keep it updated.

Note that this create a interface library, it will not create a real library (since it doesn't have cpp files).

MikeGitb:
Maybe I'm missing something, but you can already clone the gsl into a subfolder of your project and add `include_directories(<path-to-repo>)` to your cmake file to use it, no?
I'm not opposed to the change, just trying to understand what the difference is.

godbyk:
@MikeGitb The [interface library](https://cmake.org/cmake/help/v3.5/manual/cmake-buildsystem.7.html#interface-libraries) method is the more modern way to do it. It's more flexible as you can set GSL-specific preprocessor definitions, compiler flags, etc. that will be included automatically whenever you use `target_link_library(MyTarget GSL)`.

Conversely, using the `include_directories()` approach has a couple drawbacks: (1) `include_directories()` is global to the entire project (though you could work around this easily enough with `target_include_directories()`, and (2) if using GSL entails any additional compiler flags and the like, the programmer would have to know about them and add them to each and every target that relies on GSL. If GSL provides an interface library, this is handled automatically.

neilmacintosh:
@sezaru to answer the original question, no, it's not by design. We simply built the CMake file as a way of building and running the test suite relatively easily.

I'd be very happy to look at a PR if this small change would leave the existing functionality unchanged but make the library easier to consume in CMake environments.

sezaru:
[...]

Issue #392, opened 2016-10-11
'make install' should not install UnitTest++ headers


marc-kdab:

neilmacintosh:
Thanks for the issue @marc-kdab. Interesting! I've never run that target, so I never noticed. Would you like to offer a PR to correct the problem?

marc-kdab:
The GSL 'make install' should just install the gsl headers, and not the UnitTest++ library headers. Current output of make install:

  Install the project...
  -- Install configuration: ""
  -- Installing: /usr/local/include/gsl/gsl
  -- Installing: /usr/local/include/gsl/gsl_assert
  -- Installing: /usr/local/include/gsl/gsl_byte
  -- Installing: /usr/local/include/gsl/gsl_util
  -- Installing: /usr/local/include/gsl/multi_span
  -- Installing: /usr/local/include/gsl/span
  -- Installing: /usr/local/include/gsl/string_span

It should end there, but continues:

  -- Installing: /usr/local/bin/span_tests
  -- Installing: /usr/local/bin/multi_span_tests
  -- Installing: /usr/local/bin/strided_span_tests
  -- Installing: /usr/local/bin/string_span_tests
  -- Installing: /usr/local/bin/at_tests
  -- Installing: /usr/local/bin/bounds_tests
  -- Installing: /usr/local/bin/notnull_tests
  -- Installing: /usr/local/bin/assertion_tests
  -- Installing: /usr/local/bin/utils_tests
  -- Installing: /usr/local/bin/owner_tests
  -- Installing: /usr/local/bin/byte_tests
  -- Installing: /usr/local/lib/libUnitTest++.a
  -- Installing: /usr/local/include/UnitTest++/TestMacros.h
  -- Installing: /usr/local/include/UnitTest++/TestResults.h
  -- Installing: /usr/local/include/UnitTest++/UnitTestPP.h
  -- Installing: /usr/local/include/UnitTest++/HelperMacros.h
  -- Installing: /usr/local/include/UnitTest++/TestRunner.h
  -- Installing: /usr/local/include/UnitTest++/ReportAssert.h
  -- Installing: /usr/local/include/UnitTest++/ExceptionMacros.h
  -- Installing: /usr/local/include/UnitTest++/XmlTestReporter.h
  -- Installing: /usr/local/include/UnitTest++/Config.h
  -- Installing: /usr/local/include/UnitTest++/TestList.h
  -- Installing: /usr/local/include/UnitTest++/TestReporterStdout.h
[...]

PR #467, opened 2017-03-12
Make GSL a CMake consumable header-only library

johelegp:
Supersedes #370:
- Uses correct CMake version.
- Can be used in CMake without requiring it to be a sub-project.
- Can be used in CMake after installation without a local fork.

The library has been named as suggested [here](https://github.com/Microsoft/GSL/issues/363#issuecomment-276821586) to prevent collision with GNU's GSL. FYI, I had name clashes with it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,11 +1,17 @@
-cmake_minimum_required(VERSION 2.8.7)
+cmake_minimum_required(VERSION 3.0.0)
 
 project(GSL CXX)
 
-install(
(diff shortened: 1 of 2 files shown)


johelegp:
c9fde56 fixes the other part of #392.

neilmacintosh:
Thanks for the PR @johelegp but we don't want to rename the GSL. However, we have accepted a PR to add a header-only library for the GSL to the CMake file. I'll close this PR out now.

PR #470, opened 2017-03-19
Fix some corechecker warnings

MikeGitb:
This PR fixes some of the warnings that the c++ core checker throws in VS2017. Mostly const correctnes "issues" and a few suppressions around pointer arithmetic.

### DIFF
--- a/include/gsl/string_span
+++ b/include/gsl/string_span
@@ -577,7 +577,7 @@ template <class CharT, std::ptrdiff_t Extent, class T,
               std::is_convertible<T, gsl::basic_string_span<std::add_const_t<CharT>>>::value>>
 bool operator==(const gsl::basic_string_span<CharT, Extent>& one, const T& other) GSL_NOEXCEPT
 {
-    gsl::basic_string_span<std::add_const_t<CharT>> tmp(other);
+    const gsl::basic_string_span<std::add_const_t<CharT>> tmp(other);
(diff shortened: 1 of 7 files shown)


MikeGitb:
Forgot that attribute syntax is not supported by all supported compilers

include/gsl/multi_span
```diff
@@ -1494,7 +1494,9 @@ public:
 
     constexpr reference operator[](const index_type& idx) const GSL_NOEXCEPT
     {
-        return data_[bounds_.linearize(idx)];
```
neilmacintosh: From memory, `linearize()` already checks the condition that `idx` is within bounds. That is the point of the `bounds_` object. Why would we double-check the calculation here?

MikeGitb: @neilmacintosh; No idea. Apparently I wasn't fully awake when makeing this change ;) Sorry for that. I removed the last two commits as they where useless without compiler support for attribute syntax anyway.

neilmacintosh:
Thanks @MikeGitb this is helpful cleanup! As a side note, now we've dropped support for VS 2013, you should be free to add `[[suppress(...)]]` annotations in the future if needed. :-)

MikeGitb:
Yes, that was a bit of unfortunate timing on my part. If no one takes care of it first I might provide another PR when I've the time

Issue #468, opened 2017-03-17
Compile without exception support


LucaFulchir:
I am trying to use this in an other project which compiles without exception support.  
I would not mind substituting the exceptions with asserts, and let it fire only in debug.

currently I use only the `<gsl/span>` header, my problem lies in `<gsl/gsl_util>`.  
Could we get something like:
```
#if defined ( __cpp_exceptions) || \
        (defined (_MSC_VER) && defined (__CPPUNWIND)) || \
        (defined (__GNUC__) && defined (__EXCEPTIONS))
    #define ABORT_THROW (x) throw x
#else
    #define ABORT_THROW (x) (x, std::abort())
#endif
```
and then use `ABORT_THROW` when needed?

neilmacintosh:
@LucaFulchir can you help me understand why does using the `#define` of `GSL_TERMINATE_ON_CONTRACT_VIOLATION` not achieve what you want?

LucaFulchir:
No real reason, I just did not know enough the codebase, was in a bit of a rush and I assumed that it had not been tested for non-exception.

Sorry, but If I had the code ready I would have used a pull request :p

neilmacintosh:
Ok, thanks. In that case, can I close out this issue as resolved?

Simply build your GSL-using project with `GSL_TERMINATE_ON_CONTRACT_VIOLATION' set and you will get contract-checking that does not use exceptions. (Note: some functions in the GSL like `narrow()` explicitly do throw exceptions, by-design. I would just expect you do not use those in a no-exception environment).

LucaFulchir:
uhm...
That seems to work with g++, I was using clang++ which still gives error just by including `<gsl/span>`

neilmacintosh:
Can you give details of the error?

LucaFulchir:
yeah, sorry.
example file:
```
#define GSL_TERMINATE_ON_CONTRACT_VIOLATION
#include <gsl/span>
int main (void)
{
    return 0;
}
```
`gcc version 6.3.1 20170306 (GCC)`:
g++ -Wall -std=c++14 -fno-exceptions -I ./gsl/include/ span.cpp
all ok. no warning, 
`clang version 3.9.1 (tags/RELEASE_391/final)`
clang++ -Wall -std=c++14 -fno-exceptions -I ./gsl/include/ span.cpp
```
In file included from span.cpp:2:
In file included from ./gsl/include/gsl/span:25:
[...]

PR #472, opened 2017-03-28
Enable usage of gsl::narrow with exceptions disabled

djarek:
This solution uses the approach of boost::asio to enabling usage of the library in environments where exception usage is either prohibited or not feasible (due to code size constraints). A function template gsl::throw_exception has been added, which in a normal environment just throws the exception. However, when ```GSL_TERMINATE_ON_CONTRACT_VIOLATION``` is defined the function is only declared by gsl and the definition of this function template must be supplied by the library's user.

I could also modify the macro ```GSL_CONTRACT_CHECK``` so that it's exactly the same when exceptions are either enabled and disabled, however this would cause existing, working code that uses ```GSL_TERMINATE_ON_CONTRACT_VIOLATION``` to break. If this approach is chosen then this macro might as well be renamed to ```GSL_NOTHROW_ON_CONTRACT_VIOLATION``` and the behavior on contract violation would be user-defined.

Closes: #468

Signed-off-by: Damian Jarek <damian.jarek93@gmail.com>

### DIFF
--- a/include/gsl/gsl_assert
+++ b/include/gsl/gsl_assert
@@ -72,18 +72,54 @@ struct fail_fast : public std::logic_error
 {
     explicit fail_fast(char const* const message) : std::logic_error(message) {}
 };
-}
+namespace details
+{
+template <typename Exception>
+[[noreturn]] void throw_exception(Exception&& exception);
+
+} //namespace details
+} //namespace gsl
 
 #if defined(GSL_THROW_ON_CONTRACT_VIOLATION)
 
+namespace gsl
+{
+namespace details
+{
+template <typename Exception>
+[[noreturn]] void throw_exception(Exception&& exception)
+{
+    throw exception;
+}
+
+} //namespace details
+} //namespace gsl
+
 #define GSL_CONTRACT_CHECK(type, cond)                                                             \
     (GSL_LIKELY(cond) ? static_cast<void>(0)                                                       \
-                      : throw gsl::fail_fast("GSL: " type " failure at " __FILE__                  \
-                                             ": " GSL_STRINGIFY(__LINE__)))
+                      : gsl::throw_exception(gsl::fail_fast("GSL: " type " failure at " __FILE__   \
+                                             ": " GSL_STRINGIFY(__LINE__))))
 
 #elif defined(GSL_TERMINATE_ON_CONTRACT_VIOLATION)
 
 #define GSL_CONTRACT_CHECK(type, cond) (GSL_LIKELY(cond) ? static_cast<void>(0) : std::terminate())
+
+namespace gsl
+{
+namespace details
+{
+template <typename Exception>
+[[noreturn]] void throw_exception(Exception&& exception)
+{
+    static_cast<void>(exception);
+    std::terminate();
+}
+
+} //namespace details
+} //namespace gsl
+
+ #define GSL_CONTRACT_CHECK(type, cond)                                                            \
+    (GSL_LIKELY(cond) ? static_cast<void>(0) : std::terminate())
 
 #elif defined(GSL_UNENFORCED_ON_CONTRACT_VIOLATION)
 
--- a/include/gsl/gsl_util
+++ b/include/gsl/gsl_util
@@ -121,9 +121,9 @@ template <class T, class U>
 inline T narrow(U u)
 {
     T t = narrow_cast<T>(u);
-    if (static_cast<U>(t) != u) throw narrowing_error();
+    if (static_cast<U>(t) != u) gsl::details::throw_exception(narrowing_error());
     if (!details::is_same_signedness<T, U>::value && ((t < T{}) != (u < U{})))
-        throw narrowing_error();
+        gsl::details::throw_exception(narrowing_error());
     return t;
 }
 


include/gsl/gsl_assert
```diff
@@ -68,17 +68,32 @@
...
 {
+
 struct fail_fast : public std::runtime_error
 {
     explicit fail_fast(char const* const message) : std::runtime_error(message) {}
 };
-}
+
+template <typename Exception>
+void throw_exception(Exception&& exception);
```
neilmacintosh: Given this is implementation detail it should really be in the `gsl::details` namespace (we do that for other helpers around the place).

neilmacintosh: Doesn't this also need an implementation that does std::terminate() in the case that some compiler flag that specifies no-exceptions mode is on?

djarek: Done

include/gsl/gsl_assert
```diff
@@ -68,20 +68,54 @@
...
+
 struct fail_fast : public std::runtime_error
 {
     explicit fail_fast(char const* const message) : std::runtime_error(message) {}
 };
-}
+namespace details
+{
+template <typename Exception>
+void throw_exception(Exception&& exception);
```
CaseyCarter: This forward declaration - and the declaration on line 91 - should be `[[noreturn]]` as well to make it clear that users may not define a `throw_exception` that returns.

include/gsl/gsl_assert
```diff
@@ -68,20 +68,54 @@
 
 namespace gsl
 {
+
```
CaseyCarter: Extraneous blank line here.

include/gsl/gsl_assert
MikeGitb: I think attribute syntax is not supported by vs2013 - I ran into a similar problem with a different PR. Now that VS2017 is out I hope support for vs2013 will be dropped soon.

CaseyCarter: Thanks for the reminder. I never used 2013 significantly, so I forget which features were new in 2015.  I think `[[noreturn]]` for `throw_exception` is significant enough to deserve a `GSL_NORETURN` portability macro until VS2013 support is dropped.

MikeGitb: @CaseyCarter: Yeah, I also needed the CI to remind me...

neilmacintosh: As previously advertised, I'm dropping VS 2013 support as VS 2017 has been released. I'm using Casey's conveniently timed PR to fix CI to drive this (sorry to piggyback Casey :-)).

So feel free to use the attributes normally and I'll make sure this goes in after the CI changes.

include/gsl/gsl_assert
```diff
@@ -72,16 +72,49 @@ struct fail_fast : public std::runtime_error
 {
     explicit fail_fast(char const* const message) : std::runtime_error(message) {}
 };
-}
+namespace details
+{
+template <typename Exception>
+[[noreturn]] void throw_exception(Exception&& exception);
```
neilmacintosh: One final thing. Can we ensure that when the "no enforcement" macro is defined, `throw_exception` will still throw an exception? That won't affect `Ensures` or `Expects` but means the contract for `narrow()` (or any other functions that come in the future like it) can be maintained correctly.

djarek: As it is now, the build will fail at the linking stage if the user defines ```GSL_UNENFORCED_ON_CONTRACT_VIOLATION``` and uses ```narrow()```. My assumption was that if the user does not want "normal" enforcement with exceptions, they will have to provide their own implementation of ```throw_exception()```. What could be done instead is that we provide a specialization of ```throw_exception``` for ```gsl::narrowing_error``` when ```GSL_UNENFORCED_ON_CONTRACT_VIOLATION``` is defined.

neilmacintosh: I'm happy with `throw_exception()` throwing an exception under `GSL_UNENFORCED_ON_CONTRACT_VIOLATION`. That way we can use it for cases like `narrow()` where the only options a user gets are: exception or terminate (and we are allowing people to choose terminate because it is strictly more severe and allows non-exception environments). 

Saying "no contract enforcement" should apply only to contract enforcement, not to a function like `narrow()`, whose purpose for existence is to always check-and-ensure.

ligfx:
Any updates on this? I'd love to start using `gsl::span` everywhere (seriously, it's amazing), but the requirement on exceptions is a deal-breaker in some cases.

djarek:
@ligfx 
I almost forgot about this PR ;). I'll try to resolve the conflicts later today.

ericLemanissier:
+1 !

JayKickliter:
Is this PR dead? I'd really like to see it get merged so I can use `gsl::span` in an embedded project.

neilmacintosh:
I'll try to look at resolving the conflicts and accepting this tomorrow.

ttshughes:
@neilmacintosh Its nearly a year later, can you resolve the conflicts and merge it, only you have the power its out of our hands.

djarek:
@ttshughes 

This patch has been merged https://github.com/Microsoft/GSL/pull/640

PR #473, opened 2017-03-29
Added support of not_null<smart_ptr> comparison

alexey-malov:
This pull requests adds comparison operators support for gsl::not_null on smart pointers such as shared_ptr.

Comparison already worked for `not_null<raw pointer>`. Now not_null smart pointers can also be compared.

### DIFF
--- a/include/gsl/gsl
+++ b/include/gsl/gsl
@@ -117,6 +117,42 @@ std::ostream& operator<<(std::ostream& os, const not_null<T>& val)
     return os;
 }
 
+template <class T, class U>
+bool operator==(const not_null<T>& lhs, const not_null<U>& rhs)
+{
+    return lhs.get() == rhs.get();
+}
+
+template <class T, class U>
+bool operator!=(const not_null<T>& lhs, const not_null<U>& rhs)
+{
+    return lhs.get() != rhs.get();
+}
+
+template <class T, class U>
+bool operator<(const not_null<T>& lhs, const not_null<U>& rhs)
+{
+    return lhs.get() < rhs.get();
+}
+
+template <class T, class U>
+bool operator<=(const not_null<T>& lhs, const not_null<U>& rhs)
+{
+    return lhs.get() <= rhs.get();
+}
+
+template <class T, class U>
+bool operator>(const not_null<T>& lhs, const not_null<U>& rhs)
+{
+    return lhs.get() > rhs.get();
+}
+
+template <class T, class U>
+bool operator>=(const not_null<T>& lhs, const not_null<U>& rhs)
+{
+    return lhs.get() >= rhs.get();
+}
+
 // more unwanted operators
 template <class T, class U>
 std::ptrdiff_t operator-(const not_null<T>&, const not_null<U>&) = delete;
--- a/include/gsl/gsl
+++ b/include/gsl/gsl
@@ -118,37 +118,37 @@ std::ostream& operator<<(std::ostream& os, const not_null<T>& val)
 }
 
 template <class T, class U>
-bool operator==(const not_null<T>& lhs, const not_null<U>& rhs)
+auto operator==(const not_null<T>& lhs, const not_null<U>& rhs) -> decltype(lhs.get() == rhs.get())
 {
     return lhs.get() == rhs.get();
 }
 
 template <class T, class U>
-bool operator!=(const not_null<T>& lhs, const not_null<U>& rhs)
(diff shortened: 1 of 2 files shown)


include/gsl/gsl
```diff
@@ -117,6 +117,42 @@ std::ostream& operator<<(std::ostream& os, const not_null<T>& val)
     return os;
 }
 
+template <class T, class U>
+bool operator==(const not_null<T>& lhs, const not_null<U>& rhs)
+{
+    return lhs.get() == rhs.get();
+}
```
CaseyCarter: These would be better with return-type expression SFINAE:
```c++
template <class T, class U>
auto operator==(const not_null<T>& t, const not_null<U>& u) -> decltype(t.get() == u.get())
{
    return t.get() == u.get();
}
```
although I suspect VS2013 would explode. @neilmacintosh, is it time yet to drop VS2013 support?

alexey-malov: Shall I re-submit the pull request?

neilmacintosh: I agree with @CaseyCarter, it would probably be best to SFINAE here (roll on Concepts, sigh). You can just push more changes to this PR if you like.

alexey-malov: I added some pulls. Could you please review them?

tests/notnull_tests.cpp
```diff
@@ -95,6 +145,104 @@ SUITE(NotNullTests)
...
+        using NotNull2 = not_null<decltype(p2)>;
+
+        CHECK((NotNull1(p1) == NotNull1(p1)) == true);
+        CHECK((NotNull1(p1) == NotNull2(p2)) == false);
+
+        CHECK((NotNull1(p1) != NotNull1(p1)) == false);
+        CHECK((NotNull1(p1) != NotNull2(p2)) == true);
+
+        CHECK((NotNull1(p1) < NotNull1(p1)) == false);
+        CHECK((NotNull1(p1) < NotNull2(p2)) == (p1 < p2));
```
CaseyCarter: The result of comparing `&x` and `&y` with `<`, '>`, `<=`, and `>=` is unspecified, and hence might not even be deterministic. The standard allows you to evaluate this condition, but says nothing about what its value will be.

CaseyCarter: I suggest replacing `x` and `y` with two array elements:
```c++
int some_ints[] = {1, 2};
int * p1 = &some_ints[0];
const int * p2 = &some_ints[1];
```

tests/notnull_tests.cpp
```diff
@@ -95,6 +145,104 @@ SUITE(NotNullTests)
...
+        CHECK((NotNullSp1(sp1) >= NotNullSp2(sp2)) == (sp1 >= sp2));
+        CHECK((NotNullSp2(sp2) >= NotNullSp1(sp1)) == (sp2 >= sp1));
+    }
+
+	TEST(TestNotNullCustomPtrComparison)
+	{
+		int i = 42;
+		CustomPtr<int> p1(&i);
+		const int ci = 42;
+		CustomPtr<const int> p2(&ci);
```
CaseyCarter: Same problem here: `&i` and `&ci` are unrelated, comparing them with `<`, `>`, `<=` or `>=` has unspecified results.

tests/notnull_tests.cpp
alexey-malov: Done

tests/notnull_tests.cpp
```diff
@@ -181,8 +180,11 @@ SUITE(NotNullTests)
 
     TEST(TestNotNullSharedPtrComparison)
     {
-        auto sp1 = std::make_shared<int>(1);
-        auto sp2 = std::make_shared<const int>(3);
+        int ints[2] = {42, 43};
+        auto emptyDeleter = [](const int*){};
+
+        shared_ptr<int> sp1(&ints[0], emptyDeleter);
+        shared_ptr<const int> sp2(&ints[1], emptyDeleter);
```
CaseyCarter: This change to the `shared_ptr` tests is not necessary for correctness: `shared_ptr` comparisons are defined in terms of `owner_before`, which is defined in terms of `less`, which is total (valid for all pointer values). That said, the change doesn't make it incorrect.

alexey-malov: Thanks for clarification. I've simplified the test

neilmacintosh:
Thanks @alexey-malov, these changes look good.

PR #475, opened 2017-03-29
CI updates:

CaseyCarter:
Travis: use cmake 3.7
Appveyor: Add .appveyor.yml config, test VS2013/VS2015/VS2017 x86/x64 Debug/Release.

I think @neilmacintosh may need to fix his AppVeyor configuration and tell it to simply use appveyor.yml.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -93,7 +93,7 @@ install:
   - |
     if [[ "${TRAVIS_OS_NAME}" == "linux" ]]; then
       if [[ -z "$(ls -A ${DEPS_DIR}/cmake/bin)" ]]; then
-        CMAKE_URL="https://cmake.org/files/v3.6/cmake-3.6.2-Linux-x86_64.tar.gz"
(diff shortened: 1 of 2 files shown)


CaseyCarter:
> I think @neilmacintosh may need to fix his AppVeyor configuration and tell it to simply use appveyor.yml.

Nope - looks like AppVeyor picked the yml right up.

neilmacintosh:
Thanks Casey, this looks good. But can I ask you to make one small change?

Can you drop VS 2013 from the AppVeyor config? I think you asked somewhere else about whether 2013 is still supported.

World: As per my announcement at CppCon and reply to an issue on here...VS 2013 is unsupported now that VS 2017 has been released. So dropping CI support will enact "phase 1" of discountinuing support. That will then allow us to go do the nice source cleanup as "phase 2".

CaseyCarter:
> Can you drop VS 2013 from the AppVeyor config?

Done. (and there was much rejoicing.)

neilmacintosh:
Thanks @CaseyCarter you have helped move us into the future ;-)

PR #479, opened 2017-04-03
gsl::at updates:

### LINKED ISSUE #357: gsl::at in const member functions
The following does not compile because gsl::at does not seem to have a const overload:

```
uintptr_t size(uintptr_t addr) const noexcept
{
    return gsl::at(m_allocated, addr);
}
```

In this example, `m_allocated` is a `std::array< uintptr_t >`. I get this really long winded (and totally useless) error which basically amounts to having to remove `const` from the function. If I do that, it compiles fine. I am using GCC 5.x. 

I tried adding my own, but the constexpr / template insanity 😄 has me a bit confused on how to do it properly.

CaseyCarter:
* `initializer_list` overload returns by value to avoid lifetime issues

* generic overload uses expression SFINAE to work with any type that has member `size()` and operator `[]`, which notably includes `const`/non-`const` `vector` and `array`.

* Add test coverage for const objects, rvalue `initializer_list`s, and `constexpr` usage.

Fixes #357.

### DIFF
--- a/include/gsl/gsl_util
+++ b/include/gsl/gsl_util
@@ -127,31 +127,25 @@ inline T narrow(U u)
 }
 
 //
-// at() - Bounds-checked way of accessing static arrays, std::array, std::vector
+// at() - Bounds-checked way of accessing builtin arrays, std::array, std::vector
 //
 template <class T, std::size_t N>
-inline constexpr T& at(T (&arr)[N], std::ptrdiff_t index)
-{
-    Expects(index >= 0 && index < narrow_cast<std::ptrdiff_t>(N));
-    return arr[static_cast<std::size_t>(index)];
-}
-
-template <class T, std::size_t N>
-inline constexpr T& at(std::array<T, N>& arr, std::ptrdiff_t index)
+inline constexpr T& at(T (&arr)[N], const std::ptrdiff_t index)
 {
     Expects(index >= 0 && index < narrow_cast<std::ptrdiff_t>(N));
     return arr[static_cast<std::size_t>(index)];
 }
 
 template <class Cont>
-inline constexpr typename Cont::value_type& at(Cont& cont, std::ptrdiff_t index)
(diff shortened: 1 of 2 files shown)


include/gsl/gsl_util
```diff
@@ -127,31 +127,25 @@ inline T narrow(U u)
...
 }
 
 template <class Cont>
-inline constexpr typename Cont::value_type& at(Cont& cont, std::ptrdiff_t index)
+inline constexpr auto at(Cont& cont, const std::ptrdiff_t index) -> decltype(cont[cont.size()])
 {
     Expects(index >= 0 && index < narrow_cast<std::ptrdiff_t>(cont.size()));
-    return cont[static_cast<typename Cont::size_type>(index)];
+    using size_type = decltype(cont.size());
+    return cont[static_cast<size_type>(index)];
```
neilmacintosh: Shouldn't this `static_cast` be a `narrow_cast`, like in the `Expects()`?

neilmacintosh:
Thanks @CaseyCarter, I appreciate the tidy-up and test improvements!

include/gsl/gsl_util
CaseyCarter: The `Expects` requires that `index` is in `[0, cont.size())`. Since `cont.size()` must be representable by its own type `size_type`, everything in `[0, cont,size()]` - and `index` in particular - is known to be representable by `size_type`. I speculate this generates substantially better code for debug compiles.

CaseyCarter: ... which is obviously a low-ish priority concern. Let me know if you think `narrow_cast` would provide clearer code.

neilmacintosh: I have no problem with the cast itself. I think the principle here is that we want to use the GSL types/functions within the GSL wherever appropriate (and possible). Precisely for the reason you give: to make code clear. Also to model the desired usage of the types.

With all that said, I suspect your point about the impact on debug builds is worth noting. Plus, in this case, the computation is already guaranteed in-range. So let's leave this alone for now. Thanks!

PR #480, opened 2017-04-03
fail_fast should derive from logic_error

### LINKED ISSUE #298: Differentiating exception type for Expects and Ensures for option [GSL_THROW_ON_CONTRACT_VIOLATION]

CaseyCarter:
Fixes #298.

Contract violations are logic errors.

### DIFF
--- a/include/gsl/gsl_assert
+++ b/include/gsl/gsl_assert
@@ -68,9 +68,9 @@
 
 namespace gsl
 {
-struct fail_fast : public std::runtime_error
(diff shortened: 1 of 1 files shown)


neilmacintosh:
Thanks @CaseyCarter, this fixes a long-standing issue.

Issue #471, opened 2017-03-20
missing raii_thread, detached_thread


cubbimew:
CppCoreGuidelines rules [CP.25: Prefer gsl::raii_thread over std::thread unless you plan to detach()](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rconc-raii_thread), [CP.30: Do not pass pointers to local variables to non-raii_threads](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rconc-pass) and [CP.26: Prefer gsl::detached_thread over std::thread if you plan to detach()](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rconc-detached_thread) mention `gsl::raii_thread` and `gsl::detached_thread` that appear to be unimplemented.

neilmacintosh:
Thanks for pointing that out @cubbimew. It's nice to have an issue to track getting the implementation done.

They are somewhat down the implementation priority list at the moment, unless anyone wants to submit a PR.

galik:
I happen to have these laying about that have not been tested to any real degree yet. I don't know if this is what they were intended to be. Real world template classes seem to be a whole lot more complicated looking than the ones I make...
```cpp
class detached_thread
{
public:
	template<typename Callable, typename... Args>
	explicit detached_thread(Callable&& f, Args&&... args)
	: t(std::forward<Callable>(f), std::forward<Args>(args)...) {t.detach();}

private:
	std::thread t;
};

class raii_thread
{
public:
	raii_thread(raii_thread const&) = delete;
	raii_thread(raii_thread&& other): t(std::move(other.t)) {}

	template<typename Callable, typename... Args>
	raii_thread(Callable&& f, Args&&... args)
	: t(std::forward<Callable>(f), std::forward<Args>(args)...) {}

	~raii_thread() { if(t.joinable()) t.join(); }

private:
	std::thread t;
};
```
If there is more to it than this then you probably want to have  *real* expert provide something more appropriate. 😄

neilmacintosh:
@galik I appreciate any starting point ;-)

Would it be possible to throw together a PR containing these? That way they pass through the CLA process, so we get the legal clearance to use the code (not that I'm accusing you of trying to do anything shady, just trying to dot my i's and cross my t's). Having them as a PR would also make it easier for people to review\contribute to their evolution as necessary.

galik:
[...]

PR #481, opened 2017-04-04
PR for Issue #471 Added Thread Classes: gsl::raii_thread gsl::detached_thread

galik:
This PR adds <gsl/gsl_thread> and tests/thread_tests.cpp
Fixes for Issue #471 

**Notes:**
I have copied the interface of `std::thread` except for `std::thread::detach`. The reason being that `gsl::detached_thread` is already detached and `gsl::raii_thread` should not be detached by design.


include/gsl/gsl_thread
```diff
@@ -0,0 +1,169 @@
...
+        t.detach();
+    }
+
+    bool joinable() const { return t.joinable(); }
+
+    std::thread::id get_id() const noexcept { return t.get_id(); }
+
+    std::thread::native_handle_type native_handle() { return t.native_handle(); }
+
+    void join() { t.join(); }
```
mlalic: If the `std::thread` is actually detached, this will throw. I don't think there's a point in exposing a method that will _always_ just throw no matter what.

include/gsl/gsl_thread
```diff
@@ -0,0 +1,169 @@
...
+    raii_thread(raii_thread const&) = delete;
+    raii_thread(raii_thread&& other) : t(std::move(other.t)) {}
+
+    raii_thread(std::thread const&) = delete;
+    raii_thread(std::thread&& other) noexcept : t(std::move(other)) {}
+
+    raii_thread& operator=(raii_thread const&) = delete;
+    raii_thread& operator=(raii_thread&& other) noexcept
+    {
+        t = std::move(other.t);
```
mlalic: This `std::terminate`s if `this->t` hasn't been joined. This isn't in the spirit of an RAII wrapper. It should likely first properly "reset" `this` and then move the `rhs` into it, similarly to what other RAII managers, like `unique_ptr` would do.

A check `if (this != &rhs)` on top of it would be needed too.

galik: When you say "reset" `this` do you mean perform a `join()`? Because I am not entirely sure about that. 

I could be wrong but my feeling is that the main use for this function is when you are *moving* into a default initialized `raii_thread`. For example when returning a `raii_thread` from a function or accepting one as a parameter.

If the client actually assigns a `raii_thread` to an already running one then performing a join might introduce hidden, unexpected serialization delays and I suspect it would rarely be the intention to *move* a running `raii_thread`onto another already running one.

Would that not always be an error?

I am willing to be persuaded either way. I think I would generally rather it trigger a terminal error than silently join because (with an `raii_thread`) I would probably always want to do the join explicitly using the natural scope.

hsutter:
The main thing is to get `raii_thread` whose destructor unconditionally joins, which is what `std::thread` should have done. That part looks fine. Then we can recommend using `raii_thread` everywhere.

I'm less positive about adding a `detached_thread`... I'm not sure we should be encouraging detached threads.

Suggestion: Could we separate the PR into two so that we can adopt the `raii_thread` and further discuss the advisability of `detached_thread`?

galik:
See new PR #504

PR #483, opened 2017-04-13
Fix Bad Paths On Windows In UnitTest-Cpp Submodule

### LINKED ISSUE #438: Can't Clone Microsoft/GSL Cleanly
The version of unittest-cpp has wiki files that have invalid names for the windows platform.

When cloning the repo the files can't be created on windows which means git reports that the submodule contains changes. This can cause some CI systems to fail as it believes the checkout was unsuccessful. I changed unittest-cpp to have valid wiki page names, but I noticed that the gsl points to the Microsoft/unittest-cpp which is far behind the source repo. I'm not sure what the best way to fix this in the gsl would be, should we just make it point to unittest-cpp/unittest-cpp then advance to the tip? Are there changes in Microsoft/unittest-cpp that are necessary for the gsl? Was it a conscious decision to stay on this specific unittest-cpp hash? 

Thanks,

Julien.

jumonatr:
- unittest-cpp's wiki submodule used to contain items that had invalid path names on windows.
- updated to latest unittest-cpp that includes a fix for that and made the gsl submodule point back to the origin.
- resolves #438

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "tests/unittest-cpp"]
 	path = tests/unittest-cpp
-	url = https://github.com/Microsoft/unittest-cpp.git
+	url = https://github.com/unittest-cpp/unittest-cpp.git


neilmacintosh:
Thanks @jumonatr, looks good to me.

Issue #486, opened 2017-04-16
project not formatted as per .clang-format


tiagomacarios:
Looks like a lot of the files are not formatted according to the .clang_format. Is this by design?
Some projects prefer to format as they go as not to "pollute" git blame. If this is not the case I can create a PR fixing the formatting issues.

neilmacintosh:
The .clang-format file came along after we already had some source, and not everyone (me included) always remembers to format before check-in.

Personally, I'm ok with things drifting a bit here and there - it's just formatting ;-). If you are the sort of person who would like to see things tidied up, feel free to offer a PR that just fixes the formatting.

neilmacintosh:
Thanks for the PR @tiagomacarios. I think this can now be closed out.

PR #490, opened 2017-04-18
Reformat files to follow clang-format style

tiagomacarios:
Project files were not following the clang-format style. For people
using IDEs were clang-format is always run after a save this would
cause unwanted changes.

This commit only applies "clang-format -i" to files.

"Fixes" #486

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,10 +1,20 @@
-cmake_minimum_required(VERSION 2.8.7)
+cmake_minimum_required(VERSION 3.8.0)
 
 project(GSL CXX)
 
-install(
-    DIRECTORY include/gsl
-    DESTINATION include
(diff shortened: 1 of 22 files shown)


rianquinn:
Is there a reason that the PR process doesn't check for formatting issues via Travis CI? See here for an example (would be easy to do the same thing for Clang Format):
https://github.com/Bareflank/hypervisor/blob/master/.travis.yml#L57

This way, there isn't an issue with Git history moving forward.

neilmacintosh:
@tiagomacarios it looks as though this PR also contains changes to move to CMake 3.8. Can you separate those out from the formatting changes?

tiagomacarios:
@rianquinn @neilmacintosh I'll close this PR and open a new one with the correct commits.

Issue #495, opened 2017-04-21
Update to Catch instead of UnitTest-cpp


rianquinn:
I have an updated Travis CI script that adds Clang Format, Clang 3.9, Clang 4.0, and Git Check. Apparently, Clang 3.9 and Clang 4.0 have been broken for some time now as UnitTest-cpp segfaults with these compilers on it's own unit tests. 

To add to the issues with UnitTest-cpp, on my own project, UnitTest-cpp cannot compile at all on Cygwin, which means neither can the GSL which prevents automated testing with the GSL on AppVeyor Cygwin builds. We currently patch the GSL to remove UnitTest-cpp so that the GSL works in all environments (GCC, Clang, Cygwin and Windows). 

The simple solution is to move to Catch.hpp which is a simple, header only library that works in all of these environments. Furthermore, the syntax is very similar to UnitTest-cpp, so the diff will not be that large. Catch.hpp is also a much larger project with regular updates, while UnitTest-cpp hasn't been updated in 3 months. 

My Proposal:
- PR#1: Move to Catch.cpp, update unit tests
- PR#2: Update Travis CI script 

I already have PR#2 done, but I don't want to start PR#1 without a blessing from @neilmacintosh

neilmacintosh:
Thanks for the issue @rianquinn! I'll do a little thinking on this and get back to you.

CaseyCarter:
Just for another data point: I can also reproduce unittest's unit tests crashing with a double free for release compiles with clang 3.9, 4, and 5. I haven't had time to investigate the bug and file a proper report, but that is why I didn't enable clang 3.9 and 4 in my most recent update to the Travis CI script.

tiagomacarios:
I faced the same issue last week and opened a ticket on the unittest-cpp github: https://github.com/unittest-cpp/unittest-cpp/issues/154

I thought about suggesting catch too. But restrained myself =)

rianquinn:
yeah, I posted an issue with respect to this same code not compiling on Cygwin, and haven't gotten any feedback yet. The project doesn't seem to be be as supported as Catch.

tiagomacarios:
I tried to figure out why it was crashing and from what I gathered it was crashing on Test::RunImpl(), but I did not figure out why exactly. Anyway, I agree with you. Even if we figured the problem I am not sure the PR process would go through. Seems like there is not much action going on over there.

If we move to a newer clang version I have a PR to add clang-tidy support to the build.

[...]

PR #501, opened 2017-04-28
Replace UnitTest-cpp with Catch

rianquinn:
This PR replaces UnitTest-cpp with Catch. Catch is downloaded
using cmake's ExternalProject_Add and installed into the
include directory as it's a header only library.

This also enables the Clang 3.9 tests as they no longer segfault. Clang 4.0 has not been added yet because it is missing from Travis CI. Progress on that can be seen here:
https://github.com/travis-ci/apt-package-whitelist/issues/3296

Signed-off-by: Rian Quinn <“rianquinn@gmail.com”>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ cache:
     - ${TRAVIS_BUILD_DIR}/deps/llvm-3.6.2/install
     - ${TRAVIS_BUILD_DIR}/deps/llvm-3.7.1/install
     - ${TRAVIS_BUILD_DIR}/deps/llvm-3.8.1/install
-    - ${TRAVIS_BUILD_DIR}/deps/llvm-3.9.0/install
+    - ${TRAVIS_BUILD_DIR}/deps/llvm-3.9.1/install
 
 matrix:
   include:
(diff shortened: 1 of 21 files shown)


rianquinn:
https://github.com/Microsoft/GSL/issues/495

.travis.yml
```diff
@@ -61,6 +61,19 @@ matrix:
     - env: CLANG_VERSION=3.8 BUILD_TYPE=Release
       os: linux
       addons: *clang38
+    - env: CLANG_VERSION=3.9 BUILD_TYPE=Debug
```
CaseyCarter: We should replace the occurrences of "3.9.0" on lines 13 and 126 with "3.9.1" to get the more recent libc++.

rianquinn: Fixed 👍

rianquinn:
@neilmacintosh any needed for this PR?

neilmacintosh:
@rianquinn Only enough time at my end to do the legal things that will allow this change to happen. I should get to it soon. Thanks for the PR and to everybody for their patience here.

neilmacintosh:
Thanks very much for putting this together @rianquinn! I had to do a manual merge as the source had moved a bit underneath you. Also I had to add some extra legal notices to the repo to reflect the use of Catch.

PR #506, opened 2017-05-11
Add GSLConfig.cmake and export GSL target

scheibel:
This PR should ease the usage of the GSL library for other cmake-based projects.
Ideally, It enables users to find the GSL package without providing their own find script for GSL using the following, standard cmake snippet:
```cmake
find_package(GSL) // unfortunately, not working as expected
```
Unfortunately, cmake comes with a find script for another GSL library (the GNU Scientific Library) so a user must declare to use the *config* mode of the ```find_package``` call:
```cmake
find_package(GSL CONFIG) // this should do the trick
```

Later in the CMakeLists, GSL can be added as dependency:
```cmake
target_link_libraries(${target}
    PUBLIC
    GSL::GSL
)
```
Propagation of the include directory of GSL to the target is handled here.

These changes were tested on an Ubuntu 16.04, cmake 3.5.1 using both an in-source build
```shell
cd GSL
cmake .
```
and a standard system install
```shell
cd GSL
cmake .
sudo make install
```


CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.1.3)
+cmake_minimum_required(VERSION 2.8.7)
```
Sarcasm: Is this correct?

`target_sources()` is used below, and requires CMake >= 3.1.0, I think:
- https://cmake.org/cmake/help/v3.1/release/3.1.0.html?highlight=target_sources#commands

CMakeLists.txt
```diff
@@ -44,21 +44,52 @@ target_compile_definitions(GSL INTERFACE
...
 )
 
+# Check for system dir install
+set(SYSTEM_DIR_INSTALL FALSE)
+if("${CMAKE_INSTALL_PREFIX}" STREQUAL "/usr" OR "${CMAKE_INSTALL_PREFIX}" STREQUAL "/usr/local")
+    set(SYSTEM_DIR_INSTALL TRUE)
+endif()
+
+# Installation paths
+if(UNIX AND SYSTEM_DIR_INSTALL)
```
Sarcasm: Why is this necessary?
I don't think you need any special case (anymore?) to install on Windows/Linux or the system.

The CMake files will be found in `share/MicrosoftGSL` whether or not they are installed on the system.
Also, distinguishing between Windows on Linux does not seem necessary for the CMake files.

On a PR I have for google/cctz, I have the following:

```cmake
include(GNUInstallDirs)
...
set(CMAKE_CONFIG_INSTALL_DIR ${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME})

install(DIRECTORY include/cctz DESTINATION include)
install(TARGETS cctz
        EXPORT ${PROJECT_NAME}-targets
        LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
        ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
        RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
install(EXPORT ${PROJECT_NAME}-targets
        NAMESPACE cctz::
        DESTINATION ${CMAKE_CONFIG_INSTALL_DIR})
install(FILES cmake/${PROJECT_NAME}-config.cmake
        DESTINATION ${CMAKE_CONFIG_INSTALL_DIR})
```

IMHO, the less differences between platforms, the better.

CMakeLists.txt
```diff
@@ -44,21 +44,52 @@ target_compile_definitions(GSL INTERFACE
...
-        ${CMAKE_CURRENT_SOURCE_DIR}/include
-    >
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+    $<INSTALL_INTERFACE:include>
 )
 
 # add natvis file to the library so it will automatically be loaded into Visual Studio
 target_sources(GSL INTERFACE
-    ${CMAKE_CURRENT_SOURCE_DIR}/GSL.natvis
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/GSL.natvis>
```
Sarcasm: Don't think it is needed, and it's just complicate things.

What is did not work with the previous method?

CMakeLists.txt
```diff
@@ -44,21 +44,52 @@ target_compile_definitions(GSL INTERFACE
...
 )
 
+# Check for system dir install
+set(SYSTEM_DIR_INSTALL FALSE)
```
Sarcasm: I don't think you need to check this, please don't.

Many people and build tool use -DCMAKE_INSTALL_PREFIX= to some custom location but expect the same layout.

JordanMaples:
Closing this PR, find_package support was added with #784. If anything is missing from that change, please file a follow up Issue or PR

Issue #477, opened 2017-03-31
Restrict T in owner<T> to only be raw pointer.


neilmacintosh:
Anything else is nonsensical. `owner<T>` is designed as a bridge for code that must deal directly with owning pointers for some reason.

neilmacintosh:
Thanks for the PR @bashrc-real, this has now been fixed.

PR #507, opened 2017-05-13
Restricting usage of owner<T> to pointer types

bashrc-real:
Also added basic test for a simple case for integer

### DIFF
--- a/include/gsl/gsl
+++ b/include/gsl/gsl
@@ -57,6 +57,7 @@ using std::shared_ptr;
 //
 // owner
 //
+//
 // owner<T> is designed as a bridge for code that must deal directly with owning pointers for some reason
 //
 // T must be a pointer type
@@ -107,7 +108,6 @@ public:
 
     constexpr operator T() const { return get(); }
     constexpr T operator->() const { return get(); }
-    constexpr auto operator*() const { return *get(); }
 
     // prevents compilation when someone attempts to assign a null pointer constant
     not_null(std::nullptr_t) = delete;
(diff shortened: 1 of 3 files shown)


bashrc-real:
Fix for issue #477

neilmacintosh:
Thanks for the PR @bashrc-real! This contains two unrelated changes, however. I'd prefer to separate out the changes to `owner<T>` and `not_null<T>` so that they are separate. Would that be possible?

bashrc-real:
@neilmacintosh  I have seperated the PRs for the two issues. This one only addresses #477 
Seems like the failure in tests is due to platform issue unrelated to this change.

neilmacintosh:
Thanks @bashrc-real, I span off a new build and it did indeed pass. I have a minor comment for you to consider.

bashrc-real:
@neilmacintosh  I dont see your comment in the review. Am I missing something

neilmacintosh:
@bashrc-real sorry, that's probably my fault...it appears I forgot to click "Complete Review" so that it became visible to you. I have done so now!

include/gsl/gsl
```diff
@@ -53,7 +54,16 @@ namespace gsl
...
+//
+// owner
+//
+//
+// owner<T> is designed as a bridge for code that must deal directly with owning pointers for some reason
+//
+// T must be a pointer type
+// - disallow construction from any type other than pointer type
+//
+template <class T, typename = std::enable_if_t<std::is_pointer<T>::value>>
```
bashrc-real: Fixed the NIT. Couldn't understand the comment about ::value?

neilmacintosh:
Thanks @bashrc-real this all looks good to me now!

PR #511, opened 2017-05-22
span_iterator: converting constructor isn't a copy constructor

### LINKED ISSUE #510: Compiling `span_tests.cpp` fails on Linux with `clang++` using `libstdc++`
Creating a `std::match_result<BiIter>` fails (line 1530) when using a `gsl::span<char>::iterator`.

**error: exception specification of explicitly defaulted move constructor does not match the calculated one**

```cpp
    std::match_results<span<char>::iterator> match; // ERROR!!

    std::regex_match(s.begin(), s.end(), match, std::regex(".*"));
    CHECK(match.ready());
    CHECK(!match.empty());
```

```
In file included from /home/galik/tmp/GSL/tests/span_tests.cpp:25:
In file included from /usr/bin/../lib/gcc/x86_64-redhat-linux/6.3.1/../../../../include/c++/6.3.1/regex:62:
/usr/bin/../lib/gcc/x86_64-redhat-linux/6.3.1/../../../../include/c++/6.3.1/bits/regex.h:1577:7: error: exception
      specification of explicitly defaulted move constructor does not match the calculated one
      match_results(match_results&& __rhs) noexcept = default;
      ^
/home/galik/tmp/GSL/tests/span_tests.cpp:1530:46: note: in instantiation of template class
      'std::__cxx11::match_results<gsl::details::span_iterator<gsl::span<char, -1>, false>,
      std::allocator<std::__cxx11::sub_match<gsl::details::span_iterator<gsl::span<char, -1>, false> > > >' requested
      here
    std::match_results<span<char>::iterator> match;
                                             ^
1 error generated.

```
Compiler version info.
```
clang version 3.9.1 (tags/RELEASE_391/final)
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

CaseyCarter:
Fixes #510.

* Constrain the converting constructor to not participate in overload resolution when `IsConst` is true, so that it is never a copy constructor.
* Use Default Member Initializers for `span_iterator`'s data members so that the default constructor can be explicitly defaulted.
* Declare all members of `span_iterator` `GSL_NOEXCEPT`: they only throw when contract violations throw.
* Don't use `&` in `operator->` since evil types may overload it.

### DIFF
--- a/include/gsl/span
+++ b/include/gsl/span
@@ -141,36 +141,34 @@ namespace details
         using reference = std::conditional_t<IsConst, const element_type_, element_type_>&;
         using pointer = std::add_pointer_t<reference>;
 
-        constexpr span_iterator() GSL_NOEXCEPT : span_iterator(nullptr, 0) {}
+        span_iterator() = default;
 
         constexpr span_iterator(const Span* span, typename Span::index_type index)
-            : span_(span), index_(index)
+            : span_(span), index_(index) GSL_NOEXCEPT
         {
             Expects(span == nullptr || (index_ >= 0 && index <= span_->length()));
         }
 
-        friend class span_iterator<Span, true>;
-        constexpr span_iterator(const span_iterator<Span, false>& other) GSL_NOEXCEPT
+        friend span_iterator<Span, true>;
+        template<bool B, std::enable_if_t<!B && IsConst>* = nullptr>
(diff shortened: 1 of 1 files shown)


neilmacintosh:
@CaseyCarter I love it! Thanks very much for solving this.

Issue #275, opened 2016-03-24
Compiler warning if gsl.h is included after windows.h


dripfeed:
1. Use Visual Studio 2015.
2. Create new Win32 Console app.
3. Target platform is auto-set to 8.1 (but it doesn't matter if I use v140 or v120).

In **stdafx.h**

### Good
```
#pragma once
#include "GSL-master/include/gsl.h"
#include <windows.h>
```

### Bad
```
#pragma once
#include <windows.h>
#include "GSL-master/include/gsl.h"
```

d:\tmp\gsl-test-1\gsl-test-1\gsl-master\include\span.h(90): warning C4003: not enough actual parameters for macro 'max'

neilmacintosh:
This is not a deficiency in the GSL headers, rather this is a known issue with using `windows.h` from C++. Please refer to #253 for some possible solutions to the problem.

dripfeed:
Wow. In all my years of win32 dev I've never encountered that. Thanks Neil.

grahamreeds:
Really? How long have you been programming Win32 for?

It was a daily occurrence back in the Visual Studio 6 days...

GR

dripfeed:
LOL - I've been on win32 since 1995. I guess I never used min or max very much.

Issue #253, opened 2016-01-22
max misinterpreted in MFC project


daixtrose:
Hi!
I am not sure if you want to fix this, but MFC has some odd definition for max. Therefore 

```c++
static const SizeType max_value = std::numeric_limits<SizeType>::max();
```
fails with an error message simmilar to this one:

packages\microsoft.gsl.0.1.1.0\build\native\include\span.h(90): warning C4003: not enough actual parameters for macro 'max'

This is easily fixed with a dirty trick, namely extra parentheses
```c++
static const SizeType max_value = (std::numeric_limits<SizeType>::max)();
```
Your decision if you want to support the usage of GSL in such an old environment ...

KInd Regards,

Markus

gdr-at-ms:
This is an issue with Windows headers.  The best fix is for you to #undef that macro after including Windows headers.  Because it is likely to cause problems elsewhere.

neilmacintosh:
@gdr-at-ms  has it right here @daixtrose. The issue is not with GSL source but with Windows headers. A popular solution is to `#define NOMINMAX` before including any headers that include `windows.h`. That way `windows.h` will not `#define` macros for min and max. [This relevant thread ](http://stackoverflow.com/questions/4913922/possible-problems-with-nominmax-on-visual-c)on StackOverflow may be useful.

PR #512, opened 2017-05-25
Fixes conflict with max macro from windows.h

alexey-malov:
The fix for conflict between std::numeric_limits<T>::max and 'max' macro defined in <windows.h>.

The following code cound reproduce compilation warning C4003.

```c++
#include <windows.h>
#include <gsl/gsl>

int main()
{
}
```
```
c:\sdk\gsl\include\gsl\multi_span(82): warning C4003: not enough actual parameters for macro 'max'
```
It fixes the following issues #291, #253, #275 without need to `#define NOMINMAX` before including windows.h

BTW, the gsl::detail::SizeTypeTraits structure from `multi_span` that caused this warning is not used anywhere in GSL. Is it possible to remove this structure from GSL if it is not used?

### DIFF
--- a/include/gsl/multi_span
+++ b/include/gsl/multi_span
@@ -76,12 +76,6 @@ namespace gsl
 */
 namespace details
 {
-    template <typename SizeType>
-    struct SizeTypeTraits
-    {
-        static const SizeType max_value = (std::numeric_limits<SizeType>::max)();
-    };
-
     template <typename... Ts>
     class are_integral : public std::integral_constant<bool, true>
     {
--- a/include/gsl/multi_span
+++ b/include/gsl/multi_span
@@ -79,7 +79,7 @@ namespace details
     template <typename SizeType>
     struct SizeTypeTraits
     {
-        static const SizeType max_value = std::numeric_limits<SizeType>::max();
+        static const SizeType max_value = (std::numeric_limits<SizeType>::max)();
     };
 
     template <typename... Ts>


include/gsl/multi_span
```diff
@@ -79,7 +79,7 @@ namespace details
     template <typename SizeType>
     struct SizeTypeTraits
     {
-        static const SizeType max_value = std::numeric_limits<SizeType>::max();
+        static const SizeType max_value = (std::numeric_limits<SizeType>::max)();
```
gdr-at-ms: Hmm, you don't really want to transmoglify the source code like that.
There are other ways to work around this.  For example, add `-DNOMINMAX` to the command line when compiling and you know you're dragging `windows.h` into the source code.

alexey-malov: I know about `#define NOMINMAX` trick. And I completely understand that it is windows.h fault. However some libraries for windows (for example, GDI+) rely on presence of such macro definition.
What do you think about removing SizeTypeTraits from GSL? It is not anymore.

gdr-at-ms: I think the problem should be fixed outside the source file -- not necessarily by code removal.

neilmacintosh:
@alexey-malov we have been through this discussion a number of times before and the consistent position is that we are not going to unnecessarily modify the GSL code to work around what is essentially a mistake in a particular platform's header files. So while I appreciate the effort of creating a PR, I'm not going to accept it at this time.

alexey-malov:
Thank you for clarification.

PR #514, opened 2017-05-27
GSL_NOEXCEPT should appear before the initialization list

galik:
Moved it to the correct position

### DIFF
--- a/include/gsl/span
+++ b/include/gsl/span
@@ -143,8 +143,8 @@ namespace details
 
         span_iterator() = default;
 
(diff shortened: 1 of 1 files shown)


CaseyCarter:
Ouch, my fault. @neilmacintosh: the fact that this passed CI indicates you may want to think about adding some smoke tests that don't define `GSL_THROW_ON_CONTRACT_VIOLATION`, even if it's just an "include all the headers" or "instantiate all the templates but don't actually do anything" test.

neilmacintosh:
Thanks for fixing this @galik, this looks good to me.

PR #515, opened 2017-05-30
Fix use of 0 instead of nullptr warning

garyfurnish:
This causes an error on clang with all warnings.


garyfurnish:
Is there a way to force tests to rerun?  This failure is pretty mysterious given this is literally swap 0 for nullptr.

neilmacintosh:
Yes, there is a way to re-run the tests and I've spun off a new round as it looks like a timing failure in the test infrastructure.

neilmacintosh:
Thanks @garyfurnish, nice cleanup!

PR #513, opened 2017-05-26
Adding derference operator to not_null

bashrc-real:
Fix for issue #491 
Also added relevant test

PR #516, opened 2017-05-31
Fixes dereferencing operator issue #517 introduced in PR #513

alexey-malov:
dereferencing operator added in PR #513 returned a copy of the object
instead of reference to it.
Adding decltype(auto) as return type of operator* fixes this issue.


neilmacintosh:
Nice catch @alexey-malov!

PR #519, opened 2017-06-01
Use std::byte when available.

neilmacintosh:
This small change allows the GSL to use the std::byte implementation from MSVC when it is available.

### DIFF
--- a/include/gsl/gsl_byte
+++ b/include/gsl/gsl_byte
@@ -37,10 +37,30 @@
 #pragma push_macro("noexcept")
 #define noexcept /*noexcept*/
 #endif           // _MSC_VER <= 1800
+
+#if _HAS_STD_BYTE 
+
+#define GSL_USE_STD_BYTE 1
+
+#else
+
+#define GSL_USE_STD_BYTE 0
+
+#endif // _HAS_STD_BYTE
+
 #endif           // _MSC_VER
 
 namespace gsl
 {
+#if GSL_USE_STD_BYTE
+
+#include <cstddef>
+
+using byte = std::byte;
+using std::to_integer;
+
+#else // GSL_USE_STD_BYTE
+
 // This is a simple definition for now that allows
 // use of byte within span<> to be standards-compliant
 enum class byte : unsigned char
@@ -108,6 +128,8 @@ inline constexpr IntegerType to_integer(byte b) noexcept
 {
     return static_cast<IntegerType>(b);
 }
+
+#endif // GSL_USE_STD_BYTE
 
 template <bool E, typename T>
 inline constexpr byte to_byte_impl(T t) noexcept
--- a/include/gsl/gsl_byte
+++ b/include/gsl/gsl_byte
@@ -38,15 +38,29 @@
 #define noexcept /*noexcept*/
 #endif           // _MSC_VER <= 1800
 
-#if _HAS_STD_BYTE 
+// this tests if we are under MSVC and the standard lib has std::byte and it is enabled
+#if _MSC_VER >= 1911 && (!defined(_HAS_STD_BYTE) || _HAS_STD_BYTE) 
 
 #define GSL_USE_STD_BYTE 1
 
-#else
+#else // _MSC_VER >= 1911 && (!defined(_HAS_STD_BYTE) || _HAS_STD_BYTE) 
 
 #define GSL_USE_STD_BYTE 0
 
-#endif // _HAS_STD_BYTE
+#endif // _MSC_VER >= 1911 && (!defined(_HAS_STD_BYTE) || _HAS_STD_BYTE) 
+
+#else // _MSC_VER
+
+// this tests if we are under GCC or Clang with enough -std:c++1z power to get us std::byte
+#if defined(__cplusplus) && (__cplusplus > 201703L)
+
+#define GSL_USE_STD_BYTE 1
+
+#else // defined(__cplusplus) && (__cplusplus > 201703L)
+
+#define GSL_USE_STD_BYTE 0
+
+#endif //defined(__cplusplus) && (__cplusplus > 201703L)
 
 #endif           // _MSC_VER
 
@@ -56,7 +70,7 @@ namespace gsl
 
 #include <cstddef>
 
-using byte = std::byte;
+using std::byte;
 using std::to_integer;
 
 #else // GSL_USE_STD_BYTE
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -21,6 +21,7 @@ add_library(gsl_tests_config INTERFACE)
 target_compile_options(gsl_tests_config INTERFACE
     $<$<CXX_COMPILER_ID:MSVC>:
         /EHsc
+        /std:c++latest
         /W4
         /WX
     >
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -21,7 +21,6 @@ add_library(gsl_tests_config INTERFACE)
 target_compile_options(gsl_tests_config INTERFACE
     $<$<CXX_COMPILER_ID:MSVC>:
         /EHsc
-        /std:c++latest
         /W4
         /WX
     >


include/gsl/gsl_byte
```diff
@@ -37,10 +37,30 @@
 #pragma push_macro("noexcept")
 #define noexcept /*noexcept*/
 #endif           // _MSC_VER <= 1800
+
+#if _HAS_STD_BYTE 
+
```
gdr-at-ms: Would it make sense to test for _MSC_VER instead?

gdr-at-ms: and/or also test for `__cplusplus`?  For example, both GCC and Clang provide `std::byte` when compiling with `-std=c++1z` which also implies a value of `__cplusplus` over 201703L.

CaseyCarter: If one knew the `_MSC_FULL_VER` that corresponds to 15.3, yes.

neilmacintosh: @gdr-at-ms That particular test is already inside a test for `_MSC_VER` (or at least I thought it was). So the intent was to test for both the presence of MSVC (via `_MSC_VER`) and the presence of `std::byte` (via the MSVC STL feature flag).

include/gsl/gsl_byte
```diff
@@ -37,10 +37,30 @@
...
+#define GSL_USE_STD_BYTE 1
+
+#else
+
+#define GSL_USE_STD_BYTE 0
+
+#endif // _HAS_STD_BYTE
+
 #endif           // _MSC_VER
```
CaseyCarter: I would put a:
```c++
#else
#define GSL_USE_STD_BYTE 0
```
here, to avoid clang users getting `-Wundef` warnings when we test `GSL_USE_STD_BYTE` on line 55.

CaseyCarter: (...and leave an obvious place for people to tack on support for other standard libraries if they so desire.)

include/gsl/gsl_byte
neilmacintosh: I agree it would be nice to support GCC and Clang's `byte`s too. I thought I'd just add the support for MSVC `std::byte` in this change and do them in a separate pass. This is the one I need now internally ;-).

gdr-at-ms: OK.   I guess, I would feel more confortable if the test wasn't for `_HAS_STD_BYTE :-)
@Casey: _MSC_VER_FULL is too finegrained

include/gsl/gsl_byte
```diff
@@ -37,10 +37,30 @@
...
 
 namespace gsl
 {
+#if GSL_USE_STD_BYTE
+
+#include <cstddef>
+
+using byte = std::byte;
```
CaseyCarter: Why not simply `using std::byte` instead of defining an alias?

include/gsl/gsl_byte
neilmacintosh: Oooh, thanks! Will do.

include/gsl/gsl_byte
neilmacintosh: Because stupid. Thanks, will fix.

include/gsl/gsl_byte
CaseyCarter: Will `_MSC_VER` be bumped up from 1910 in 15,3?

gdr-at-ms: @neilmacintosh Sure.  I generally prefer and encourage testing for compiler version than feature test macros.  Compiler versions tend to be more reliable :-)

CaseyCarter: (I should point out that we added `std::byte` for 15.3; it was not in previous library releases.)

gdr-at-ms: @CaseyCarter yes.  We are now committed to bump at every new official release.  In fact, we've already bumped :-)

gdr-at-ms: @CaseyCarter See https://blogs.msdn.microsoft.com/vcblog/2016/10/05/visual-c-compiler-version/

CaseyCarter: >  yes. We are now committed to bump at every new official release. In fact, we've already bumped :-)

Wow, I'm way out of the loop. In that case, I agree that testing for `_MSC_VER >= XXX` is better than using `_HAS_STD_BYTE` which may go away again some day.

neilmacintosh: @gdr-at-ms I'm not sure what option I have left. `_HAS_STD_BYTE` is the symbol that indicates to us whether a user has opted into `std::byte` with MSVC STL or not. It is possible for people to opt-out to avoid conflicts with existing code (regardless of the version of their compiler). If we don't check that symbol in the MSVC case, then we would fail in those environments. At least this way people would get a fallback `gsl::byte` in those cases.

gdr-at-ms: @neilmacintosh I understand your conundrum, but I think on average the compiler version is the better version.  The feature test macro `_HAS_STD_BYTE` is a last resort escape hatch and may go away (at least that has been my understanding from internal discussion with libs folks.)

gdr-at-ms: Maybe `#if defined(_HAS_STD_BYTE) && _MSC_VER >= 1911` and then educate people that the `_HAS_STD_BYTE` might go away?

CaseyCarter: I suggest:
```c++
#if _MSC_VER >= 1911 && (!defined(_HAS_STD_BYTE) || _HAS_STD_BYTE)
```
Which will avoid trying to use `std::byte` when a user has defined `_HAS_STD_BYTE` to 0 to disable the feature, but will continue to work fine if we rip out the macro entirely.

gdr-at-ms: @CaseyCarter Better.  Thanks!

neilmacintosh: @gdr-at-ms I'm happy with that as a compromise. Personally, I started with the compiler version and then hit exactly the conundrum I described. So to avoid leaving behind an important set of internal users while they (necessarily) make use of `_HAS_STD_BYTE` in the short term, I'll make that change and add a comment pointing out that `_HAS_STD_BYTE` is likely to go away one day. I'll add in the GCC and Clang support now too, given I'm going to edit the %$#%# thing anyway.

gdr-at-ms: OK, see @CaseyCarter 's version -- which is a better condition :-)

CaseyCarter: There's unfortunately no SD-6 feature test macro for `std::byte`, which libc++ and libstdc++ would otherwise define appropriately. You should complain to the author of the `std::byte` proposal ;)

gdr-at-ms: I am volunteering myself as his attorney on that topic, pro bono :-)

neilmacintosh: @CaseyCarter I would, but I already do all the time, and it gets me nowhere. Sigh.

gdr-at-ms:
Looks good!  Thanks for your patience and taking the additional input.

neilmacintosh:
Of course! Thanks for the help everyone.

include/gsl/gsl_byte
```diff
@@ -37,10 +37,44 @@
...
+#else // _MSC_VER >= 1911 && (!defined(_HAS_STD_BYTE) || _HAS_STD_BYTE) 
+
+#define GSL_USE_STD_BYTE 0
+
+#endif // _MSC_VER >= 1911 && (!defined(_HAS_STD_BYTE) || _HAS_STD_BYTE) 
+
+#else // _MSC_VER
+
+// this tests if we are under GCC or Clang with enough -std:c++1z power to get us std::byte
+#if defined(__cplusplus) && (__cplusplus > 201703L)
```
CaseyCarter: This should be `>= 201703L`, no?

gdr-at-ms: @CaseyCarter Yes, eagle eye :-)

neilmacintosh: Aha. Yes..it should indeed! I am comparison-blind apparently. I'll fix that now.

PR #520, opened 2017-06-01
Fix STD_BYTE cstddef include in wrong place

garyfurnish:
This fails to compile on gcc 7.1 because it tries to include cstddef inside the gsl namespace.

### DIFF
--- a/include/gsl/gsl_byte
+++ b/include/gsl/gsl_byte
@@ -55,6 +55,7 @@
 #if defined(__cplusplus) && (__cplusplus >= 201703L)
 
 #define GSL_USE_STD_BYTE 1
+#include <cstddef>
 
 #else // defined(__cplusplus) && (__cplusplus >= 201703L)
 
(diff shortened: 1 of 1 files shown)


garyfurnish:
This seems to have the same timeout error as earlier.

tests/CMakeLists.txt
```diff
@@ -38,7 +38,7 @@ target_compile_options(gsl_tests_config INTERFACE
         -Woverloaded-virtual
         -Wpedantic
         -Wshadow
-        -Wsign-conversion
+#        -Wsign-conversion
```
neilmacintosh: Why did you comment out this warning for GCC\Clang?

neilmacintosh:
Thanks for fixing up this silly mistake Gary!, I just have one comment for you to consider (about the change to the `CMakeLists.txt`.

garyfurnish:
The change to CMakeLists.txt wasn't supposed to be left in and was a work around to fix #525 (which is a different issue).

neilmacintosh:
Thanks @garyfurnish and thanks for opening #525 to track the separate narrowing issue.

Issue #522, opened 2017-06-02
Wrong multi_span constructor called when passing const std::array


VincentZalzal:
When constructing a multi_span from a const std::array, the constructor for containers is called instead. This actually happens in the unit tests. This is caused by a wrong template parameter used in the function signature (compare it with the non-const std::array version).

VincentZalzal:
Pull request has been submitted.

neilmacintosh:
Thanks for the issue and PR @VincentZalzal!

PR #523, opened 2017-06-02
Fix multi_span constructor taking a const std::array (#522)

VincentZalzal:
The current constructor is not called even by the unit test (it is currently calling the "container" constructor instead). I mimicked the constructor taking a non-const std::array to fix the issue #522.


neilmacintosh:
Thanks @VincentZalzal, this looks like a pretty good fix for the problem to me.

PR #526, opened 2017-06-02
Fix issue #525 GSL tests broken with GCC 7.1.0

### LINKED ISSUE #525: GSL tests broken with GCC 7.1.0
```/mnt/fscratch0/pub/build/root-gcc-7.1.0-system/bin/g++
COLLECT_LTO_WRAPPER=/mnt/fscratch0/pub/build/root-gcc-7.1.0-system/libexec/gcc/x86_64-pc-linux-gnu/7.1.0/lto-wrapper
Target: x86_64-pc-linux-gnu
Configured with: ../../dependencies/gcc-7.1.0/configure --prefix=/mnt/fscratch0/pub/build/root-gcc-7.1.0-system --enable-languages=c,c++ --without-multilib
Thread model: posix
gcc version 7.1.0 (GCC)
```
This produces a ton of sign error spam when compiling tests.

Example:

```In file included from /mnt/devdiskvault/gfurnish/ap/GSL/include/gsl/gsl_algorithm:22:0,
                 from /mnt/devdiskvault/gfurnish/ap/GSL/tests/algorithm_tests.cpp:19:
/mnt/devdiskvault/gfurnish/ap/GSL/include/gsl/span: In constructor ‘constexpr gsl::span<ElementType, Extent>::span(gsl::span<ElementType, Extent>::element_type (&)[N])’:
/mnt/devdiskvault/gfurnish/ap/GSL/include/gsl/span:373:51: error: conversion to ‘long int’ from ‘long unsigned int’ may change the sign of the result [-Werror=sign-conversion]
         : storage_(&arr[0], details::extent_type<N>())
```
See [log](https://www.dropbox.com/s/4e7xxsi8217q4qh/gsl_errors.txt?dl=0)
I haven't looked into why this is happening yet but removing the -Wsign-conversion warning from tests does fix the issue.

garyfurnish:
This fixes #525 test failures/sign errors.

### DIFF
--- a/include/gsl/multi_span
+++ b/include/gsl/multi_span
@@ -1186,8 +1186,11 @@ namespace details
         : std::true_type
     {
     };
-
+#if defined(_MSC_VER)
     template <typename ValueType, std::ptrdiff_t Rank>
+#else
+    template <typename ValueType, std::size_t Rank>
+#endif
     struct is_multi_span_oracle<strided_span<ValueType, Rank>> : std::true_type
     {
     };
@@ -1298,7 +1301,11 @@ public:
     }
 
     // construct from n-dimensions static array
+#ifdef _MSC_VER
     template <typename T, std::size_t N, typename Helper = details::SpanArrayTraits<T, N>>
+#else
+    template <typename T, std::size_t N, typename Helper = details::SpanArrayTraits<T, narrow_cast<std::ptrdiff_t>(N)>>
+#endif
     constexpr multi_span(T (&arr)[N])
         : multi_span(reinterpret_cast<pointer>(arr), bounds_type{typename Helper::bounds_type{}})
     {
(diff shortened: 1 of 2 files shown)


garyfurnish:
Fixing this is broken in 2015, seems to potentially work in 2017.  There may be a better way to handle this then whats in here.

garyfurnish:
Wanted to re-ping on this and ask if this is sufficient or if I should approach the problem a different way?

MikeGitb:
Why all the #ifdefs? Doesn't the new code with the narrow_cast also work with msvc? And why are you changing so many template parameters from std::ptrdiff to std::size?

garyfurnish:
I think the argument deduction fails in VS2015 on the narrow cast code. The template parameters changes are because rank is supposed to be size_t but some of the template parameters give a ptrdiff_t for rank.

MikeGitb:
@garyfurnish: Sorry, overlooked that. Any idea, why rank is an unsigned variable, while the gsl strives to use signed integers everywhere else?

neilmacintosh:
@garyfurnish @MikeGitb From memory, that is not by-design. We adapted `multi_span` from a previous codebase and I think rank may have been overlooked.

garyfurnish:
It should at least be internally consistent one way or another... Changing it from size_t to index_type would be a massive api-breaking patch.

neilmacintosh:
It should be consistent, I agree.

galik:
Those of us who have just upgraded to `GCC 7.1` are kind of blocked by this. Is this patch ready to go or are you going a different route?

garyfurnish:
I consider it ready to go (as in tests cleanly and fixes the problem), there were two questions:
1) Should rank be changed to a ptrdiff_t (a different problem in my opinion)
2) Is this the "right" way to get around older MSVC type deduction issues (which is the ifdef controversy).  I defer to the experts on this one.

garyfurnish:
Ping on this.

neilmacintosh:
@garyfurnish sorry for the delay in getting back to you. I wouldn't want to take this patch with the `#ifdef`s in place. Do we understand why the `narrow_cast` changed doesn't work in MSVC 2015? That seems like the root cause for all the `#ifdef` work and so is worth investigating and fixing.

garyfurnish:
After a bit of study the issue is that prior to some version of 2017 it appears that narrow_cast can not be used in constexpr content in a template parameters.  I could make a constexpr_narrow_cast macro I guess?

tiagomacarios:
whenever someone fixes this please enable GCC7 in the .travis file (you just need to uncomment the section)

scribam:
Superseded by #652?

annagrin:
current fix that was accepted is in #690. Thanks all!

Issue #529, opened 2017-07-04
CMakeLists.txt requires 2.8.7 but uses feature introduced in 3.1.3


EdgarWahn:
target_sources is not available in 2.8.7 nor 3.0.2. Either change cmake_minimum_required to 3.1.3 or remove / replace that command.

neilmacintosh:
Thanks for the warning @EdgarWahn, I'll update the minimum requirement text.

neilmacintosh:
Fixed by #533.

Issue #494, opened 2017-04-20
Make install will also install UnitTest++


tiagomacarios:
Currently, if you run `make install` it will also install UnitTest++. AFAIU every install command in a cmake file is run during install. There is no way to select which one to install. Since GSL adds UnitTest++ through a subfolder and UnitTest++ has an install command it will also get installed.

I will try to patch the UnitTest++ cmake file to avoid this.

```
[ 32%] Built target UnitTest++
[ 35%] Built target algorithm_tests
[ 38%] Built target multi_span_tests
[ 41%] Built target span_tests
[ 44%] Built target strided_span_tests
[ 47%] Built target string_span_tests
[ 50%] Built target at_tests
[ 52%] Built target bounds_tests
[ 55%] Built target byte_tests
[ 58%] Built target owner_tests
[ 61%] Built target notnull_tests
[ 64%] Built target assertion_tests
[ 67%] Built target utils_tests
[100%] Built target TestUnitTest++
Install the project...
-- Install configuration: ""
-- Installing: /mnt/c/Users/Mac/GitHub/GSL/build_linux/include/gsl
-- Installing: /mnt/c/Users/Mac/GitHub/GSL/build_linux/include/gsl/gsl
-- Installing: /mnt/c/Users/Mac/GitHub/GSL/build_linux/include/gsl/gsl_algorithm
-- Installing: /mnt/c/Users/Mac/GitHub/GSL/build_linux/include/gsl/gsl_assert
-- Installing: /mnt/c/Users/Mac/GitHub/GSL/build_linux/include/gsl/gsl_byte
-- Installing: /mnt/c/Users/Mac/GitHub/GSL/build_linux/include/gsl/gsl_util
-- Installing: /mnt/c/Users/Mac/GitHub/GSL/build_linux/include/gsl/multi_span
-- Installing: /mnt/c/Users/Mac/GitHub/GSL/build_linux/include/gsl/span
-- Installing: /mnt/c/Users/Mac/GitHub/GSL/build_linux/include/gsl/string_span
-- Installing: /mnt/c/Users/Mac/GitHub/GSL/build_linux/lib/libUnitTest++.a
-- Installing: /mnt/c/Users/Mac/GitHub/GSL/build_linux/include/UnitTest++/AssertException.h
-- Installing: /mnt/c/Users/Mac/GitHub/GSL/build_linux/include/UnitTest++/CheckMacros.h
[...]

PR #533, opened 2017-07-13
Move from unittest-cpp to catch for unit testing.

### LINKED ISSUE #495: Update to Catch instead of UnitTest-cpp

neilmacintosh:
Many thanks to @rianquinn. This should fix #495, #494 and #529.


rianquinn:
Awesome. Once this is merged I will start work on getting the clang-format stuff integrated into Travis CI as we discussed.

rianquinn:
oh, and will also up-rev the version of catch to include the NO_SELFTEST work that we go upstreamed which should speed up Travis.

neilmacintosh:
Actually...you can't just up-rev the version of catch at will I'm afraid. Each time we change the version that we snap to, I need to update legal documentation to keep everyone happy.

Are there likely to be other upgrades/changes to Catch worth waiting on so we can batch up that effort?

rianquinn:
I'm not sure, but we are in no hurry either. As far as I know, the answer is no as we are not really using that much of Catch.hpp. Most of the cygwin issues that I know of have been fixed in the version you have, and it works fine with VS, GCC and Clang.

neilmacintosh:
Great. So this seems like the right version to start from. Let's sit with it for a little while (a month or so) and if nothing else worth waiting for appears to be on the horizon, we can bump the version.

rianquinn:
sounds great

PR #536, opened 2017-07-23
Allow users to override GSL_USE_STD_BYTE

CaseyCarter:
Because the comment "this tests if we are under GCC or Clang with enough -std:c++1z power to get us `std::byte`" is wrong: the correlation between compiler and standard library versions is at best loose in the open source world. There is no reasonable way for a third party library to determine if the implementation supports `std::byte`, so users must be able to override our guess when it is wrong.

Of course `byte` isn't implementable as a pure library feature in C++, so the fallback implementation is not portable. It would probably be better - at least the behavior would be portable - for the library to use `unsigned char` in place of `byte` absent implementation support for `std::byte`.

### DIFF
--- a/include/gsl/gsl_byte
+++ b/include/gsl/gsl_byte
@@ -28,19 +28,22 @@
 // don't warn about function style casts in byte related operators
 #pragma warning(disable : 26493)
 
+#ifndef GSL_USE_STD_BYTE
 // this tests if we are under MSVC and the standard lib has std::byte and it is enabled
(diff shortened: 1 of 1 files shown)


neilmacintosh:
Seems like a totally reasonable thing to do.

Yes, we should probably have an `unsigned char` fallback for `byte` on compilers that would otherwise be unhappy. MSVC accepts the existing fallback implementation, so it is probably thing to keep that as an exception.

MikeGitb:
FYI, Regarding the unsigned char fallback, there was some related discussion here: https://github.com/Microsoft/GSL/pull/369

PR #540, opened 2017-08-01
Fix catch installation issue

### LINKED ISSUE #538: CMake installs catch into top-level source dir
The `CMAKE_INSTALL_PREFIX` for catch [is set to](https://github.com/Microsoft/GSL/blob/be43c79742dc36ee55b21c5d531a5ff301d0ef8d/tests/CMakeLists.txt#L9) `${CMAKE_SOURCE_DIR}`. I believe the intention is to place the catch/catch.hpp file in the GSL/include directory. However, if GSL is built from another project with eg. `add_subdirectory`, `CMAKE_SOURCE_DIR` is the top-level source directory and not the GSL directory, so catch is installed in the wrong place (and can't be found by the tests, etc).

I'm using `${CMAKE_CURRENT_SOURCE_DIR}/..` instead.

tiagomacarios:
Catch was being installed to the root of the cmake project. This
violated the source code tree itself and would not work at all if GSL
was being consumed with add_subdirectory.

CMake will now install catch to the build tree under /dependencies.

This fixes #538

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 CMakeFiles
 build
-include/catch
 tests/CMakeFiles
 tests/Debug
 *.opensdf
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -6,7 +6,7 @@ project(GSLTests CXX)
 set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 
 list(APPEND CATCH_CMAKE_ARGS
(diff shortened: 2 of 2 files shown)


tiagomacarios:
@rianquinn wdyt?

rianquinn:
@tiagomacarios yeah, that looks perfect

neilmacintosh:
LGTM, thanks for getting this fixed!

Issue #542, opened 2017-08-05
Creating string_spans with `char16_t` & `char32_t` is not working.


galik:
The compiler complains when you try to make a span with `char16_t` & `char32_t`
```cpp
gsl::basic_string_span<const char> c("stuff");         // good
gsl::basic_string_span<const wchar_t> wc(L"stuff");    // good
gsl::basic_string_span<const char16_t> u16c(u"stuff"); // compile error!
gsl::basic_string_span<const char32_t> u32c(U"stuff"); // compile error!
```
On inspection this appears to be because various overloads and functor specializations are missing from `<gsl/string_span>`.

neilmacintosh:
Fixed by #543...thanks @galik!

PR #543, opened 2017-08-05
Made string_span details::string_length() generic (Fix issue #542)

galik:
Made `string_length()` generic (templated) Removed overloads & specialized classes

Creating *string spans* using `char16_t` and `char32_t` was not possible without creating new specializations and function overloads.

This patch makes `details::string_length()` generic removing the need to extend the overloads and specializations.

This PR intends to fix Issue #542

### DIFF
--- a/include/gsl/string_span
+++ b/include/gsl/string_span
@@ -123,76 +123,18 @@ span<T, dynamic_extent> ensure_sentinel(T* seq, std::ptrdiff_t max = PTRDIFF_MAX
 }
 
 //
-// ensure_z - creates a span for a czstring or cwzstring.
+// ensure_z - creates a span for a zero terminated strings.
 // Will fail fast if a null-terminator cannot be found before
 // the limit of size_type.
 //
-template <typename T>
-inline span<T, dynamic_extent> ensure_z(T* const& sz, std::ptrdiff_t max = PTRDIFF_MAX)
+template <typename CharT>
+inline span<CharT, dynamic_extent> ensure_z(CharT* const& sz, std::ptrdiff_t max = PTRDIFF_MAX)
 {
-    return ensure_sentinel<T, 0>(sz, max);
+    return ensure_sentinel<CharT, CharT(0)>(sz, max);
 }
 
-// TODO (neilmac) there is probably a better template-magic way to get the const and non-const
-// overloads to share an implementation
-inline span<char, dynamic_extent> ensure_z(char* const& sz, std::ptrdiff_t max)
-{
-    auto len = details::string_length(sz, max);
-    Ensures(sz[len] == 0);
-    return {sz, len};
-}
-
-inline span<const char, dynamic_extent> ensure_z(const char* const& sz, std::ptrdiff_t max)
-{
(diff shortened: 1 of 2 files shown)


neilmacintosh:
@galik This looks like a nice, simple solution. But could you add some tests to verify that construction of `string_span` with `char16_t` and `char32_t` work as intended?

galik:
@neilmacintosh I have the new commits but I can't compile the tests using `GCC 7.1.1` because of issue #525. The new commits work in `clang++` though. Should I submit it while `GCC`tests are still broken or will it fail the integration checks?

BTW I have a fix for `GCC 7` somewhere (it just suppresses the warnings) if your interested in that while a better fix comes along?

galik:
This turned out to be a little more than adding the tests because adding tests exposed the fact that other parts of `string_span` did not cater to `char16_t` and `char32_t` types also (mainly type aliases). So I added those in a way that seems consistent with current nomenclature.

include/gsl/string_span
```diff
@@ -150,14 +151,42 @@ inline span<const char, dynamic_extent> ensure_z(const char* const& sz, std::ptr
...
+}
+
+inline span<char32_t, dynamic_extent> ensure_z(char32_t* const& sz, std::ptrdiff_t max)
+{
+    auto len = details::string_length(sz, max);
+    Ensures(sz[len] == 0);
+    return {sz, len};
+}
+
+inline span<const char32_t, dynamic_extent> ensure_z(const char32_t* const& sz, std::ptrdiff_t max)
```
neilmacintosh: I'll confess I haven't looked at this code in a little while, but is there a reason we wouldn't just have a generic `ensure_z<CharT>()`? All these individual implementations appear to do the same thing and have no type-specific code.

galik: Yes that's the obvious thing to do. But I tried it and failed. For some reason I could not figure out how to get that working. From what I could tell it looked like `char16_t` and `char32_t` as template parameters didn't resolve the same way? I think I could get it working for `char16_t` but not `char32_t` or maybe the other way round? My template knowledge is lacking so  maybe someone with better template foo can fix that?

galik: At line 131 there is a template version of the function. It is followed by more specialized overloads for `char` types. I am not sure why that is. Are those char type overloads even necessary?

I am adding a new commit that simply deletes the char type overloads. I can't see that they do anything different from the main templated function. I also changed the template parameter from `T` to `CharT` to be inline with the other generic functions.

galik: As a side note is `ensure_z()` such a great name for this function? It feels more like a `make_span_from_zstring()` or something that suggests it creates a string_span?

neilmacintosh: Thanks @galik, deleting that old code was a nice bit of cleanup work (and removed a TODO that named me!). The name was originally chosen to distinguish between `assume_z()`: where it's a "trust-me" statement from the programmer. I would rather not rename at this point.

neilmacintosh:
LGTM, thanks very much for your work and patience here @galik!

PR #548, opened 2017-08-09
turning off warning

rianquinn:
using the GSL with MSVC 2017 causes the following warning"
`'std::copy_n::_Unchecked_iterators::_Deprecate': Call to 'std::copy_n'`

This patch removes this since the point of this code is to deal with this issue.


neilmacintosh:
Thanks @rianquinn, nice clean-up!

PR #549, opened 2017-08-11
Use system version of the Catch test framework

mymedia2:
This removes dependency on Git for tests. This eliminates CMake errors in environments without installed Git. Also an user can force to use the Catch library which is installed in `/usr/include` or something, if (s)he defines the _CMAKE_DISABLE_FIND_PACKAGE_Git_ variable for CMake invocation. Like this

    cmake /path/to/GSL -DCMAKE_DISABLE_FIND_PACKAGE_Git=1

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,7 @@ cmake_minimum_required(VERSION 3.1.3)
 project(GSL CXX)
 
 include(ExternalProject)
-find_package(Git REQUIRED)
+find_package(Git)
 
 # creates a library GSL which is an interface (header files only)
 add_library(GSL INTERFACE)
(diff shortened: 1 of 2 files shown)


mymedia2:
It looks Travis build failed due to a very old version of CMake :hushed:. I hope it's not my fault.

CaseyCarter:
> It looks Travis build failed due to a very old version of CMake 😯. I hope it's not my fault.

The travis script failed to download a newer cmake version from cmake.org, presumably due to a transient network issue. I suggest force pushing a new commit atop what's in your PR branch to trigger Travis to rerun. (Or you could wait for @neilmacintosh to notice this thread and go restart the job manually.)

rianquinn:
yeah, its back up and running, this just happened to me as well on our repo, but it's working now. Would be great if Travis updated CMake as everyone has to update for the same reason, which means Travis is hitting their website pretty hard everyday.

neilmacintosh:
This looks good to me. Thanks @mymedia2!

PR #551, opened 2017-08-17
Reimplement final_act using the scope_exit design from P0052R5

### LINKED ISSUE #125: finally() mishandles modifiable lvalues
Meng Zhu pointed this out to me.  gsl.h contains:

```
template <class F>
class final_act
{
[...]
private:
    F f_;
    bool invoke_;
};

// finally() - convenience function to generate a final_act
template <class F>
final_act<F> finally(const F &f) noexcept { return final_act<F>(f); }

template <class F>
final_act<F> finally(F &&f) noexcept { return final_act<F>(std::forward<F>(f)); }
```

Given a const lvalue of type `X`, `finally(const F&)` is selected, deducing `F` to be `X`, and it returns `final_act<X>`.

However, given a modifiable lvalue of type `X`, `finally(F&&)` is selected, deducing `F` to be `X&`, and it returns `final_act<X&>`. This appears to be completely undesired.

If `final_act<F>` assumes that `F` is an object type, it should `static_assert` so, and `finally()` should be fixed accordingly.  The most robust fix would be to provide a perfect forwarder only, and use `decay`.

Lastly, all other occurrences of perfect forwarding in the GSL should be audited for this problem, especially when perfect forwarders are overloaded with anything else of the same arity (perfect forwarders are extremely greedy, and will outcompete other overloads, often unintentionally).

CaseyCarter:
Fixes #125.

### DIFF
--- a/include/gsl/gsl_util
+++ b/include/gsl/gsl_util
@@ -37,6 +37,16 @@
 #endif                          // _MSC_VER < 1910
 #endif                          // _MSC_VER
 
+#ifndef GSL_USE_DEDUCTION_GUIDES
+#ifdef __cpp_deduction_guides
+#define GSL_USE_DEDUCTION_GUIDES __cpp_deduction_guides
+#elif !(defined(__GNUC__) || defined(__clang__)) && __cplusplus >= 201703L
+#define GSL_USE_DEDUCTION_GUIDES 201611L
+#else
+#define GSL_USE_DEDUCTION_GUIDES 0
+#endif
+#endif // GSL_USE_DEDUCTION_GUIDES
+
 namespace gsl
 {
 //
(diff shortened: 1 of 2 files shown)


neilmacintosh:
Thanks for the PR @CaseyCarter!

At the moment, we like the simplicity of the design of `final_act` as it stands (modulo bugs). We can revisit whether to harmonize `final_act` with a standardized facility that offers similar functionality once there is such a thing (P0052 is still in-flight for now).

So I'll leave thi PR open, but mark it "deferred" for now (because I wanted to use that label and have never had a chance before ;-)).

JordanMaples:
Maintainer's call: closing this as the code has diverged significantly. Additionally, it looks like most of the changes @CaseyCarter wanted to implement were added in other PRs since this. Please file an issue or re-open if there are additional improvements you'd like to add to `final_action`. Thanks!

PR #552, opened 2017-08-20
Adds an option to disable test generation

bfierz:
When integrating the GSL as an external dependency, it is usually unnecessary to build and run the unit-tests.


rianquinn:
This is very similar to what I did here with Catch. 
https://github.com/philsquared/Catch/issues/897

neilmacintosh:
Thanks @bfierz this seems pretty reasonable to me.

PR #557, opened 2017-09-17
Enable vectorization of common loops using iterators with range checks enabled with GCC and Clang

annagrin:
I noticed that when range checks are enabled, loops using iterators and range-for don't get vectorized. Rewriting inequalities in the same direction (using <= instead of >=), removing null checks for span in iterators (the program will AV now instead of calling terminate, but there will be no buffer overflow) and relaxing range checks while still making sure they check the ranges as well as original ones makes it easier for GCC and Clang to vectorize the following commonly used loops that were not vectorized before:

```cpp
void test_span_iter(span<int> s) {
    for (auto it = s.begin(); it != s.end(); ++it)
    {
        (*it) += 1;
    }
}

void test_span_rangeiter(span<int> s) {
    for (auto& e : s)
    {
        e += 1;
    }
}
```


neilmacintosh:
A really nice piece of work that should benefit lots of users, thanks!

Issue #521, opened 2017-06-01
not_null comparison


Amaroker:
This code causes compiler error C2678 in Visual Studio 2017. Why is that?

```
void test()
{
	struct A {};
	struct B : public A {} b;

	const B* pB{ &b }; 
	gsl::not_null<A*> pA{ &b };
	
	assert(pA == pB);
}
```

JordanMaples:
Maintainers' call: Sorry for the delay. You would write this as `pA.get() == pB`. Thanks

JordanMaples:
I'm providing a workaround. The bug is MSVC specific and isn't a problem with GSL.

PR #558, opened 2017-10-07
Adding not_null comparison operators for raw pointers

kernhanda:
Fixes issue #521

### DIFF
--- a/include/gsl/gsl
+++ b/include/gsl/gsl
@@ -159,6 +159,78 @@ auto operator>=(const not_null<T>& lhs, const not_null<U>& rhs) -> decltype(lhs.
     return lhs.get() >= rhs.get();
 }
 
+template <class T, class U>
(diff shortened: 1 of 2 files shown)


Amaroker:
I'm still curious as to why Visual C++ has this limitation. GCC and Clang works without this fix.

kernhanda:
Ping.

xaxxon:
This is a lot of complexity and code to add for a workaround for what is likely a compiler bug.   I'd suggest further looking into alternative solutions if this compiler is required to be supported.

annagrin:
@kernhanda thank you for reporting the compiler issue! I opened a bug for msvc, so this will get sorted out. If you need those to work for compiler versions with the bug, I suggest to add those operators on your side for now. Closing the issue.

Amaroker:
@annagrin  Hi, it was I who reported this issue here. I also reported it to Developer Community (id 389748, active with status "Under Investigation" for some time now).

PR #561, opened 2017-10-18
Update std::byte detection in gsl_byte for MSVC

CaseyCarter:
* Only check that `_HAS_STD_BYTE` is defined to a non-zero value


CaseyCarter:
This is necessary for MS-internal usage that wants to use recent compilers with older STL headers.

neilmacintosh:
Thanks @CaseyCarter !

PR #563, opened 2017-10-25
Infra

tiagomacarios:
Updates infra.

First commit:
- Remove trailing spaces
- Move notification setting from the bottom to the top
- Set Linux as default OS
- Simplify cache directory
- Add banners around different configurations

Second commit:
- Make it depend on variable (easier to update)
- Simplify code a bit

Third:
- make logic depend on a single COMPILER variable
- Update LLVM installation process
- Update linux to trusty

Fourth:
- Update compilers to latest stable version
GCC 7 is blocked by #525, so whoever fixes that one please uncomment that GCC 7 code added here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ notifications:
 
 # Use Linux unless specified otherwise
 os: linux
+dist: trusty
 
 cache:
   directories:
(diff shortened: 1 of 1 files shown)


neilmacintosh:
Thanks @tiagomacarios very helpful!!!

PR #564, opened 2017-10-27
Adds clang-format as a build target

tiagomacarios:
If code is not properly formatted the build will fail and display the changes that have to be applied.

@rianquinn wdyt?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,30 @@ matrix:
   include:
 
     ##########################################################################
+    # Check for code formatting errors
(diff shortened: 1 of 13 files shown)


tiagomacarios:
I guess this broke on my last rebase, but now you see how it will work =)

@rianquinn @neilmacintosh wdyt?

annagrin:
@tiagomacarios unfortunately clang-format does not always render the test code the best way, and fails to format attributes correctly. I am hesitant to have it enforced with those issues present.

PR #565, opened 2017-10-27
Makes string_length more strict about input

### LINKED ISSUE #409: string_length is not strict enough
`string_length` returns 0 if the pointer passed is nullptr or if the length passed is < 0. Based on the principle of being strict about memory safety, it seems like it should terminate if the length is < 0, or if the length is non-zero and the pointer is null.

tiagomacarios:
I would think this closes #409

### DIFF
--- a/include/gsl/string_span
+++ b/include/gsl/string_span
@@ -92,7 +92,8 @@ namespace details
     template <class CharT>
     std::ptrdiff_t string_length(const CharT* str, std::ptrdiff_t n)
     {
-        if (str == nullptr || n <= 0) return 0;
+        Expects(!(n < 0 || (str == nullptr && n >= 0)));
+        if (n == 0 || str == nullptr) return 0;
 
         span<const CharT> str_span{str, n};
 


include/gsl/string_span
```diff
@@ -94,7 +94,7 @@ namespace details
     template <class CharT>
     std::ptrdiff_t string_length(const CharT* str, std::ptrdiff_t n)
     {
-        if (str == nullptr || n <= 0) return 0;
+        Expects(str != nullptr && n > 0);
```
neilmacintosh: This looks close to me, but not quite following the spirit of the original issue that was raised. I think it's ok for `str` to be `nullptr` and in that case, `string_length` should return `0`. Similarly, the `n` can legitimately be `0`, regardless of whether or not `str` is `nullptr`. This change would disallow those cases.

I think what you want to prevent here is when `n < 0` (which should never happen) and when `str == nullptr && n > 0` which would also be problematic.

tiagomacarios:
@neilmacintosh wdyt?

ericLemanissier:
the boolean value `!(n < 0 || (str == nullptr && n >= 0))` is equivalent to `n >= 0 && (str != nullptr || n < 0)` which is equivalent to `(n >= 0 && str != nullptr) || (n >= 0 && n < 0)`. Obviously the last one is always false, so it is equivalent to `n >= 0 && str != nullptr`

MikeGitb:
In its current form, the if statement is not needed and as @ericLemanissier pointed out, the test in Expects can be simplified. 

Personally, I'd like to be able to call `string_length` on a nullptr, if if n is zero, but there are some prominent examples in the standard library, where similar operations aren't allowed (e.g. memcpy).

PR #566, opened 2017-10-27
Move GSL pointers to a separate header

### LINKED ISSUE #502: Relocate `not_null` and `owner` into `<gsl/gsl_util>`?
Both `not_null` and `owner` are defined in the header `<gsl/gsl>` which also *includes* every other `GSL` header. However people who want `owner` or `not_null` may not want everything else. For that reason it might be better to re-locate them elsewhere.

One way to fix this would be to simply move them into the `<gsl/gsl_util>` header which already contains a few things:

* final_act
* narrow_cast
* at

This should not be a breaking change as anyone including either `not_null` or `owner` using the `<gsl/gsl>` header will continue to get them because `<gsl/gsl>` already includes all the other headers.

tiagomacarios:
Should close #502


neilmacintosh:
Thanks for taking a look at this @tiagomacarios. We may yet decompose `<gsl/pointers>` further, but I think it's a pretty convenient "grouping header" to have anyway that you've come up with here. I hope people looking to minimize header dependencies (as we wait for a proper module system to arrive in the language) find this useful.

tiagomacarios:
@neilmacintosh if you want I can break it further. I was mirroring what std does, as in `#include <memory>` brings in all pointers.

PR #568, opened 2017-10-27
additional span tests (pointer length constructor)

menete:
extended test case for span/pointer length contructor


neilmacintosh:
Thanks @menete these look like a useful addition.

PR #569, opened 2017-10-28
Minimize checking in subspan

annagrin:
rewrite span subspan checks to help optimizations
    
    * Rewrote subspan range checks to help compilers optimize away
      provable conditions
    
    * Removed checking pointer for null for subspans. We would never
      check pointer for null in between ptr and ptr+size in the
      original span, so there seems to be  no reason to do so for
      subspans, provided that the subspan's boundaries are ensured
      to be within the range of the original span's boundaries.
    
    This change allows to simplify generated code, for example, to
    remove 5 out of 9 branches in code generated from the following
    by MSVC, and 4 out 8 branches in clang and gcc-generated code:
    

```
    span<int> mysubspan(int* p, std::ptrdiff_t size, std::ptrdiff_t i)
    {
      if (p != nullptr)
      {
        span<int> s = { p, size };
        return s.subspan(i);
      }
    
      return { nullptr };
    }
```
    Similar effects are achieved for dynamic subspans of static spans,
    where in the following code we remove 2 out of 4 branches in MSVC
    and GCC-generated code:
    
```
    int test_dynamic_subspan_of_static_span(std::ptrdiff_t i)
    {
      int x[] = { 0,1,2,3,4,5 };
      span<int, 6> s = { x };
      auto subspan = s.subspan(i);
      return subspan.size();
    }

```


include/gsl/span
```diff
@@ -463,18 +464,17 @@ public:
 
     constexpr span<element_type, dynamic_extent> last(index_type count) const
     {
-        Expects(count >= 0 && count <= size());
-        return {data() + (size() - count), count};
+        Expects(count >= 0 && size() - count >= 0);
```
tlanc007: Very minor nit.  Any particular reason you're still using Expects here?  It isn't needed as `make_span` covers it.  In your change for line 474 in `sub_span` did drop Expects.

annagrin: You are right, the checks are not needed. Will update the PR.

include/gsl/span
```diff
@@ -517,10 +516,13 @@ private:
     class storage_type : public ExtentType
     {
     public:
+
+        // basechecked needed to remove unnecessary null check in subspans
```
neilmacintosh: Minor thing: could we change the name of "basechecked" to something more descriptive. Maybe "null_checked" or just "checked". It's the "base" part I find confusing.

include/gsl/span
```diff
@@ -530,8 +532,38 @@ private:
     };
 
     storage_type<details::extent_type<Extent>> storage_;
+
+    // The rest is needed to remove unnecessary null check in subspans
+    constexpr span(pointer ptr, index_type count, bool basechecked) : storage_(ptr, count, basechecked) {}
+
+
+    template <bool Dependent = false,
```
neilmacintosh: Maybe you could achieve the same effect here with slightly fewer characters by using a single boolean template parameter that tells `make_subspan` which case it is operating in. Then you wouldn't need the dummy `basechecked` parameter to distinguish them. Instead you could just use specialization to differentiate the two implementations.

include/gsl/span
annagrin: done!

include/gsl/span
annagrin: please see the updated changes.

neilmacintosh:
Thanks @annagrin, a nice reduction in overhead for subspan operations across GCC, Clang and MSVC!

PR #571, opened 2017-11-01
Remove pragma once

tiagomacarios:
Remove pragmas and expanded guards to reduce chance of collision

### DIFF
--- a/include/gsl/multi_span
+++ b/include/gsl/multi_span
@@ -14,10 +14,8 @@
 //
 ///////////////////////////////////////////////////////////////////////////////
 
-#pragma once
-
-#ifndef GSL_MULTI_SPAN_H
-#define GSL_MULTI_SPAN_H
+#ifndef GUIDELINE_SUPPORT_LIBRARY_MULTI_SPAN_H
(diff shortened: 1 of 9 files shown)


include/gsl/multi_span
```diff
@@ -14,10 +14,8 @@
 //
 ///////////////////////////////////////////////////////////////////////////////
 
-#pragma once
-
-#ifndef GSL_MULTI_SPAN_H
-#define GSL_MULTI_SPAN_H
+#ifndef GUIDELINE_SUPPORT_LIBRARY_SPAN_H
```
neilmacintosh: Shouldn't this be `GUIDELINE_SUPPORT_LIBRARY_MULTI_SPAN_H`?

tiagomacarios: Fixed

tiagomacarios:
@neilmacintosh Ill withdrawn this so it is not lingering around. If we get collisions with the guard name I can put it back together

neilmacintosh:
Thanks @tiagomacarios sounds good.

PR #572, opened 2017-11-01
Drop "Dummy" from enable_if in templates

tiagomacarios:
From what I understand [from this discussion](https://github.com/Microsoft/GSL/pull/114) looks like Dummy was needed for VS2013. Since it seems like we don't support that anymore I removed the Dummies.


neilmacintosh:
Thanks @tiagomacarios, looks like nice cleanup.

PR #574, opened 2017-11-04
Exporting MicrosoftGSL::GSL CMake target during library installation

paweldac:
Modified cmake install process to support modular cmake architecture in project using GSL.

Changes allow to export GSL CMake target and its configuration into given prefix path. Currently gsl library has to be added to project using `include_directories`. After changes are merged it will be possible to do it with `find_package` and `target_link_libraries`. Example is added to readme.

*.natvis file support was added to cmake in version 3.7: https://cmake.org/cmake/help/v3.7/release/3.7.html#generators

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,9 +43,8 @@ target_compile_definitions(GSL INTERFACE
 
 # add include folders to the library and targets that consume it
 target_include_directories(GSL INTERFACE
-    $<BUILD_INTERFACE:
-        ${CMAKE_CURRENT_SOURCE_DIR}/include
-    >
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+    $<INSTALL_INTERFACE:include>
 )
 
 if ((CMAKE_VERSION GREATER 3.7.9) OR (CMAKE_VERSION EQUAL 3.7.9))
@@ -57,9 +56,9 @@ if ((CMAKE_VERSION GREATER 3.7.9) OR (CMAKE_VERSION EQUAL 3.7.9))
 
     # add natvis file to the library so it will automatically be loaded into Visual Studio
     if(VS_ADD_NATIVE_VISUALIZERS)
-        target_sources(GSL INTERFACE
-            ${CMAKE_CURRENT_SOURCE_DIR}/GSL.natvis
-        )
+    target_sources(GSL INTERFACE
+        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/GSL.natvis>
+    )
     endif()
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,7 @@ cmake_minimum_required(VERSION 3.1.3)
 project(GSL CXX)
 
 include(ExternalProject)
+include(GNUInstallDirs)
 find_package(Git)
 
 # creates a library GSL which is an interface (header files only)
@@ -62,13 +63,39 @@ if ((CMAKE_VERSION GREATER 3.7.9) OR (CMAKE_VERSION EQUAL 3.7.9))
     endif()
 endif()
 
-install(
-    DIRECTORY include/gsl
-    DESTINATION include
-)
-
 option(GSL_TEST "Generate tests." ${GSL_STANDALONE_PROJECT})
 if (GSL_TEST)
 	enable_testing()
 	add_subdirectory(tests)
 endif ()
+
+set(package_name "MicrosoftGSL")
+set(targets_export_name "${package_name}Targets")
+set(project_config "${package_name}Config.cmake")
+set(config_install_dir "${CMAKE_INSTALL_LIBDIR}/cmake/${package_name}")
+set(project_namespace "${package_name}::")
+install(
+    TARGETS GSL
+    EXPORT ${targets_export_name}
+    INCLUDES DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
+ )
+
+install(
+    DIRECTORY include/gsl
+    DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
+)
+
+install(
+    EXPORT ${targets_export_name}
+    NAMESPACE ${project_namespace}
+    DESTINATION ${config_install_dir}
+    FILE ${project_config}
+)
+
+export(
+    EXPORT ${targets_export_name}
+    NAMESPACE ${project_namespace}
+    FILE ${project_config}
+)
+export(PACKAGE ${package_name})
+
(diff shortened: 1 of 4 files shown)


paweldac:
I didn't use GSL::GSL to avoid name colisilon with GNU GSL library(https://www.gnu.org/software/gsl/) or its cmake finder FindGSL(https://cmake.org/cmake/help/v3.2/module/FindGSL.html).

@neilmacintosh namespace name can be changed to Microsoft::GSL. I leave the decision on namespace name to you.

johelegp:
> *.natvis file support was added to cmake in version 3.7: https://cmake.org/cmake/help/v3.7/release/3.7.html#generators
Consider leaving this out for another PR so that this one is more easily accepted.

paweldac:
@johelegp Ok. I've extracted *.natvis file support in: https://github.com/Microsoft/GSL/pull/575

johelegp:
Thank you. Hopefully now it's uncontroversial, unlike other PRs which have attempted this, but also tried to do too many other things at once.

paweldac:
@johelegp I cannot. CMake forbids exporting targets, which adds files with:
`target_sources` using `INTERFACE`

johelegp:
Sorry, I don't understand what you mean. What I originally meant to say was to separate the natvis support and CMake target aspects of the PR. Natvis support is now its own PR, but this PR still has that part, which is better removed so that this has more chances of being accepted.

paweldac:
@johelegp If I would go back to current repo version:
`target_sources(GSL INTERFACE
    ${CMAKE_CURRENT_SOURCE_DIR}/GSL.natvis
)`
CMake won't allow me to export such target.

> Targets with INTERFACE_SOURCES may not be exported with the export() or install(EXPORT) commands. This limitation may be lifted in a future version of CMake.
https://cmake.org/cmake/help/v3.1/command/target_sources.html

So I can split PR into 2, but those have to be delivered sequentially.

johelegp:
That restriction is gone with v3.3.2: https://cmake.org/cmake/help/v3.3/command/target_sources.html
From the changelog: https://cmake.org/cmake/help/latest/release/3.3.html
> The install(EXPORT) and export() commands learned to export targets that populate the INTERFACE_SOURCES target property.

You could raise the version to v3.3.2.

paweldac:
@johelegp thanks for pointing this out. I thought I've checked latest documentation version as well. PR updated as requested.

CMakeLists.txt
```diff
@@ -1,8 +1,9 @@
-cmake_minimum_required(VERSION 3.1.3)
+cmake_minimum_required(VERSION 3.3.1)
```
tiagomacarios: Do we need this version bump?

CMakeLists.txt
```diff
@@ -44,23 +45,47 @@ target_compile_definitions(GSL INTERFACE
...
 	enable_testing()
 	add_subdirectory(tests)
 endif ()
+
+set(package_name "MicrosoftGSL")
+set(targets_export_name "${package_name}Targets")
+set(project_config "${package_name}Config.cmake")
+set(config_install_dir "${CMAKE_INSTALL_LIBDIR}/cmake/${package_name}")
+set(project_namespace "${package_name}::")
+install(
```
tiagomacarios: What would you think about guarding this installs behind a variable?
AFAIR not guarding turns into a problem when people consume the library as a git submodule and their own product has install directives. So if you run install in the parent repo gsl also get installed which might not be wanted.

If you think this is out of scope of this change I can open an issue

CMakeLists.txt
```diff
@@ -44,23 +45,47 @@ target_compile_definitions(GSL INTERFACE
...
+    DESTINATION ${config_install_dir}
+    FILE ${project_config}
+)
+
+export(
+    EXPORT ${targets_export_name}
+    NAMESPACE ${project_namespace}
+    FILE ${project_config}
+)
+export(PACKAGE ${package_name})
```
tiagomacarios: new line on end of file

CMakeLists.txt
johelegp: If the parent doesn't want to install GSL, they should add [`EXCLUDE_FROM_ALL`](https://cmake.org/cmake/help/latest/command/add_subdirectory.html) to the `add_subdirectory` call.

CMakeLists.txt
johelegp: It is needed to export the GSL.natvis file, such that it is a dependency on consumers. I don't know what that file is or does, so I don't know if this is the correct way to go.

tiagomacarios: It "enhances" the visual studio debugging experience. See here: https://msdn.microsoft.com/en-us/library/jj620914.aspx

So you can only use the BUILD_INTERFACE generator after 3.3.1?

CMakeLists.txt
tiagomacarios: That would work, but I am not sure it is the best way. Here are some drawbacks:
https://stackoverflow.com/questions/16486785/cmake-exclude-custom-install-targets-from-make-install-all
Anyway not a big issue

CMakeLists.txt
johelegp: The problem is that the natvis file can't be exposed to consumers through the `MicrosoftGSL::GSL` target because it's not supported before the new version.

CMakeLists.txt
paweldac: thanks for comments. In current CMake solution `install` isn't guarded with variable. If it's needed I can rise seperate PR.

CMakeLists.txt
paweldac: version 3.3.2 dropped restriction on exporting configuration of targets containing files added with `target_sources(INTERFACE ...)`.
3.3.2 was released in September 2015. It's already quite old.

CMakeLists.txt
paweldac: done

CMakeLists.txt
johelegp: Did you test with an older CMake version to notice the error? It sure has gone unnoticed for a while.

paweldac: @tiagomacarios debugger is used during development of library not in deliveries right? That's why it's used in build not install.

CMakeLists.txt
muggenhor: Another option would be to specify `EXCLUDE_FROM_ALL` and a `COMPONENT` for `install` if this isn't the top-level CMakeLists.txt

paweldac:
@annagrin can you check? Since GSL has now versions it would be nice to have cmake exports with versions and pkgconfig files for easier package management.

johelegp:
The lack of exporting GSL, and therefore importing, blocks me from being able to export my own library that depends on it. It seems like I'll have to resort to dirty hacks or require the user to install the GSL and let it be found by my lib on the system headers and work by chance if the installed version is what I expect it to be.

JordanMaples:
Closing this PR, #784 was merged today and adds support for find_package.

PR #575, opened 2017-11-05
Visual Studio Debug Visualizers support

paweldac:
changes:
    - minimal cmake version updated to 3.7(first version with .natvis file handling)
    - updated cmake version in readme file
    - GSL.natvis file is added only to VS projects

*.natvis file support was added to cmake in version 3.7: https://cmake.org/cmake/help/v3.7/release/3.7.html#generators

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.1.3)
+cmake_minimum_required(VERSION 3.7)
 
 project(GSL CXX)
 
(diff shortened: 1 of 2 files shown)


tiagomacarios:
Bumping the cmake version can cause issues to other people, we should try to be conservative with that. Ubuntu for example still ships with some really only cmake.

My understand is that the special handling for natvis is already done by cmake (it understands that file). For example, (AFAIR) if you build with ninja on windows the natvis file does not show up on the build files. So I am not sure we should handle it separately.

paweldac:
@tiagomacarios thanks for comments. Currenlty natvis files is added to GSL target in cmake for all platforms. CMake introduced natvis file support in version 3.7 which was released in November last year, before that CMake didn't understand which tool should be used for this file extension. It's possible to tell CMake which Tool should be used for specific files, but I consider it a little hacky since we have already CMake version that has proper support.

Current stable version of Ubuntu supports CMake version 3.9.1 https://launchpad.net/ubuntu/+source/cmake

tiagomacarios:
Alright, makes sense.

MikeGitb:
Ubuntu's current LTS version (16.04) is shipping cmake 3.5. I'm not sure if a lot of people are using non-LTS versions and in particular, a lot of docker Images are based on 16.04 (including Ubuntu on WSL)

paweldac:
@MikeGitb LTS provides only security updates. It's not surprising that ubuntu from april 2016 stopped on 3.5. Most of Github projects including GSL library are installing newer CMake in .travis.yml from sources.

neilmacintosh:
This looks good to me and seems a nice improvement @paweldac, thanks for the PR! I agree with the reasoning here: given it is a year old, it seems reasonable to require people to use CMake 3.7 or better.

Issue #567, opened 2017-10-27
including <iostream> not always desired


menete:
When running on an embedded Platform (like ARM Cortex-M) including `<iostream>` is not always desired – it usually increases the code size dramatically (200 … 300 kB – on targets with usually less than 1 MB of memory).
The `<iostream>` include in the ‘gsl/gsl’ file hence prevents efficient use of the not_null class.
Some options might be

- isolate the not_null into a dedicated header (not having the `ostream& operator<<` function, and keep the `operator<<` within the ‘gsl/gsl’ file.
    This would allow to do `#include <gsl/not_null>` the same way it is currently possible for `#include <gsl/span>`).

- replace `#include <iostream>` by `#include <iosfwd>`
    While this seems to work (and solves the code bloat issue) I’ mot sure whether this is the right thing to do, with respect to the `ostream operator<<` definition within this file.

I can offer to do those changes, provided some agreement on which option to choose.

neilmacintosh:
I certainly appreciate the frustration @menete. But for my education: this can't just be a problem for `not_null`....doesn't this same problem present itself with use of `std` types such as `string` and `string_view`?

menete:
@neilmacintosh It is - but has a different order of magnitude.
Just adding a `std::string` to a simple  "Hello World" example will have quite some impact on the code size - as a  `std::string` will bring quite some dependencies (heap functions, ...).

Starting from a simple "Hello World" example, which just toggles an I/O pin. The code size of the Debug Build (gcc 7.2, C++14, -O0, newlib default) evolves like this.

- ~ 2.5 kB exceptions/rtti disabled

- ~ 14 kB exceptions/rtti enabled

- ~ 76 kB when including a `std::string` variable

- ~ 282 kB for `#include <iostream>`

This is just to provide some examples. Whether exceptions are suited for an embedded target or not would be a non-trivial discussion.
Nevertheless embedded systems usually avoid dynamic memory management (due to reasons of heap fragmentation) - hence `std::string`, `std::vector`, ... 
However therer is no reason not to use `gsl::span`, `gsl::not_null`, `std::array` (or `boost::static_vector`)

neilmacintosh:
@menete Thanks for the data: very interesting! I raised std::string as an example because it does include definitions (at least in our stdlib implementation) of `operator<<`, and hence should include <iostream> (or at least enough of it to hurt).

[...]

PR #576, opened 2017-11-06
possible solution to #567

menete:
since there have not been any further suggestions to #567 I propose isolating the iostream part into a dedicated file


neilmacintosh:
I think we can close this now the alternative solution has been merged.

PR #129, opened 2015-10-05
Fix #128 rename final_act to final_action

trebconnell:
Got all uses in comments as well

PR #579, opened 2017-11-08
rename final_act to final_action

### LINKED ISSUE #128: Rename final_act to final_action

tiagomacarios:
Should fix #128. There is an old PR #129 that does the same, but it never got merged and now has merge conflicts.

So if this gets merged we should probably close #128 and #129.


neilmacintosh:
Looks good to me. Thanks for cleaning this up @tiagomacarios.

PR #580, opened 2017-11-08
Updates catch to version 2.0

tiagomacarios:
Update Catch to version 2.

Version 1 would capture exceptions by value (which would make clang dislike code at times) version 2 fixes that.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -15,8 +15,8 @@ if(GIT_FOUND)
     ExternalProject_Add(
         catch
         PREFIX ${CMAKE_BINARY_DIR}/catch
-        GIT_REPOSITORY https://github.com/philsquared/Catch.git
-        GIT_TAG v1.9.6
+        GIT_REPOSITORY https://github.com/catchorg/Catch2.git
+        GIT_TAG v2.0.1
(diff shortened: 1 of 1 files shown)


martinmoene:
Correction: Catch 1 (Catch Classic) lets the user choose how exceptions are caught, apparently a feature not well-understood by many users (and tools). It's documented in [Assertion Macros, section Assertions](https://github.com/catchorg/Catch2/blob/Catch1.x/docs/assertions.md#exceptions):
>Note that the exception type is used verbatim and you should include (const) reference.

tiagomacarios:
@martinmoene you are correct, I was referring to the way our tests are currently set. Out of curiosity, why would one want to catch the exceptions by copy?

martinmoene:
I'm only the messenger ;) 

I don't know what underpins the decision to not include `const&` in Catch1, perhaps ask @philsquared.

philsquared:
IIRC the thinking at the time was that it would look the same as if you were writing a catch block.
In retrospect most people don't seem to have intuited it that way, so we revisited that decision for Catch2.
We didn't want to make the change in Catch1.x because it would be a breaking change (for anyone that has written there handlers with references).

neilmacintosh:
Looks good, seems completely reasonable. Thanks very much @tiagomacarios!

PR #581, opened 2017-11-08
Moves VS specific error handling to separate file

tiagomacarios:
Moves VS specific error handling to separate files gsl/detail/pre|post

Makes main code a bit easier to follow and centralizes warning/error suppression. One downside I see is that we might end up suppressing more than we should in certain files.

### DIFF
--- a/include/gsl/details/post
+++ b/include/gsl/details/post
@@ -1,0 +1,11 @@
+#ifdef _MSC_VER
+
+#if _MSC_VER < 1910
+#undef constexpr
+#pragma pop_macro("constexpr")
+#endif // _MSC_VER < 1910
+
+#pragma warning(pop)
+#endif // _MSC_VER
+
+#undef GSL_NOEXCEPT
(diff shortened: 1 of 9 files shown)


neilmacintosh:
thanks for the PR @tiagomacarios. However, this is a change I don't want to take for the library. The warning suppressions and keyword workarounds for MSVC are seen as short-term workarounds. I don't want to hide them away...having them there is a regular reminder to get them to disappear. ;-)

neilmacintosh:
thanks for the PR @tiagomacarios. However, this is a change I don't want to take for the library. The warning suppressions and keyword workarounds for MSVC are seen as short-term workarounds. I don't want to hide them away...having them there is a regular reminder to get them to disappear. ;-)

neilmacintosh:
thanks for the PR @tiagomacarios. However, this is a change I don't want to take for the library. The warning suppressions and keyword workarounds for MSVC are seen as short-term workarounds. I don't want to hide them away...having them there is a regular reminder to get them to disappear. ;-)

neilmacintosh:
thanks for the PR @tiagomacarios. However, this is a change I don't want to take for the library. The warning suppressions and keyword workarounds for MSVC are seen as short-term workarounds. I don't want to hide them away...having them there is a regular reminder to get them to disappear. ;-)

PR #582, opened 2017-11-08
joining_thread implementation

paweldac:
Partial fix for issue: https://github.com/Microsoft/GSL/issues/471 Only `joining_thread` is added. It's possible to create joining_thread as global object, it'll behave like detached thread.

https://github.com/isocpp/CppCoreGuidelines/issues/925

I also took into account @hsutter comment from previous attempt: 
"The main thing is to get raii_thread whose destructor unconditionally joins, which is what std::thread should have done. That part looks fine. Then we can recommend using raii_thread everywhere."
https://github.com/Microsoft/GSL/pull/481

### DIFF
--- a/include/gsl/joining_thread
+++ b/include/gsl/joining_thread
@@ -39,8 +39,8 @@ public:
     {
     }
 
-    joining_thread(const joining_thread&) = delete;
-    joining_thread& operator=(const joining_thread&) = delete;
+    joining_thread(const joining_thread&) = default;
+    joining_thread& operator=(const joining_thread&) = default;
 
     joining_thread(joining_thread&&) noexcept = default;
     joining_thread& operator=(joining_thread&&) noexcept = default;
--- a/include/gsl/joining_thread
+++ b/include/gsl/joining_thread
@@ -1,0 +1,78 @@
+///////////////////////////////////////////////////////////////////////////////
+//
+// Copyright (c) 2015 Microsoft Corporation. All rights reserved.
+//
+// This code is licensed under the MIT License (MIT).
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+///////////////////////////////////////////////////////////////////////////////
+
+#pragma once
+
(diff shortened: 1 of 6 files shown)


paweldac:
@galik @neilmacintosh what do you think?

paweldac:
@neilmacintosh any thoughts?

arteniioleg:
I hope this will be merged soon. 

I want to use `gsl::joining_thread` [as advices](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rconc-joining_thread) but when I search in all `gsl/` headers for `thread` there is no match

mpoullet:
I've a question about the implementation: why do you put the `std::thread` in a `std::unique_ptr`?

I know of 3 implementations of this design:
- Scott Meyers' `thread_raii`
- Anthony Williams' `scoped_thread`
- Bjarne Stroustrup's `guarded_thread`

I've put them all [here](https://github.com/mpoullet/cpp-snippets/blob/master/src/thread_raii.cpp) with the details from their books. They all use directly `std::thread` and no smart pointers. The implementation proposed by @galik [here](https://github.com/Microsoft/GSL/issues/471#issuecomment-289165710) uses `std::thread` directly too.

See this [SO question](https://stackoverflow.com/q/44832054/496459) too.

include/gsl/joining_thread
```diff
@@ -0,0 +1,78 @@
...
+ * RAII wrapper for std::thread.
+ * It's guaranteed that joining_thread will join when destructed.
+ */
+class joining_thread
+{
+    using JoiningThreadUPtr = std::unique_ptr<std::thread, std::function<void(std::thread*)>>;
+    template <typename... Args>
+    JoiningThreadUPtr create_thread_with_joining_deleter(Args&&... args)
+    {
+        return {new std::thread(std::forward<Args>(args)...), [](auto* th) { th->join(); }};
```
ericLemanissier: I see a `new` with no corresponding `delete`: Where is this `std::thread` instance deleted ? the deleter you provide to `std::unique_ptr` only joins without deleting

galik:
I actually still have a live PR for this (though it is hopelessly out of date) https://github.com/Microsoft/GSL/pull/596 I am not sure if it is exactly what is required. For example my implementation has *move constructor* and *move assignment* from normal `std::thread`. Is that useful? Or an unnecessary complication? Could it cause problems?

Still it's there if it's needed.

include/gsl/joining_thread
paweldac: you're right it's an oversight. I've removed unique_ptr completely.

include/gsl/joining_thread
```diff
@@ -0,0 +1,82 @@
...
+    joining_thread() noexcept {};
+
+    template <class Function, class... Args>
+    explicit joining_thread(Function&& function, Args&&... args)
+        : thread(std::forward<Function>(function), std::forward<Args>(args)...)
+    {
+    }
+
+    joining_thread(const joining_thread&) = delete;
+    joining_thread& operator=(const joining_thread&) = delete;
```
ericLemanissier: `= default` the copy constructor and copy assignment operator would be slightly better, because it would clearly indicate that this class has the same behavior as `std::thread`. These members would still end up being deleted because corresponding members of `std::thread` are deleted.

johelegp:
Maybe the `std::jthread` proposal https://wg21.link/p0660 could help guide this design. However, according to https://botondballo.wordpress.com/2018/11/30/trip-report-c-standards-meeting-in-san-diego-november-2018/, it's still undergoing design review.

JordanMaples:
Maintainers' call: This is a good idea, however C++20 being finalized prefer using `std::jthread`. Sorry this took so long.

PR #584, opened 2017-11-08
Include <type_traits> in <gsl/string_span>

### LINKED ISSUE #583: <gsl/string_span> uses std::add_const without including <type_traits>
Obviously the implementation is picking up <type_traits> indirectly from another thing that is included, but we should be explicit about including what we use in a header.

LegalizeAdulthood:
Fixes #583


neilmacintosh:
I believe this has been superseded  by #588, so we can close it out.

PR #585, opened 2017-11-09
Remove inline added for the now unsupported MSVC 2013.

johelegp:


paweldac:
is msvc implementing whole c++14 standard? It's not added in supported platform list.

johelegp:
No entry for it in the C++11 and 14 tables of http://en.cppreference.com/w/cpp/compiler_support is marked red, so I guess so.

neilmacintosh:
@johelegp thanks very much for this cleanup, it looks good to me! MSVC 2013 is definitely no longer a supported target.

PR #586, opened 2017-11-13
Iosfwd

menete:
Possible Solution to #576.

With restricting `gsl/pointer` to include `<iosfwd>` rather than `<iostream>` following application will no longer compile.
```
#include <gsl/gsl>
int main()
{
    int i;
    gsl::not_null<int*> p(&i);
    std::cout << p << std::endl;
}
```
Reason is simply because `gsl` no longer includes `<iostream>`, hence this needs to be done within the applications code as shown below.

```
#include <iostream>
#include <gsl/gsl>
int main()
{
    int i;
    gsl::not_null<int*> p(&i);
    std::cout << p << std::endl;
}
```

### DIFF
--- a/include/gsl/pointers
+++ b/include/gsl/pointers
@@ -21,7 +21,7 @@
 
 #include <gsl/gsl_assert>
 
-#include <iostream>
+#include <iosfwd>
 #include <memory>
 #include <type_traits>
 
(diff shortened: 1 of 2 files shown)


neilmacintosh:
Thanks for this PR @menete. Thanks also for a submitting well-described issue and taking the time to discuss solution options.

I like this approach because it allows a practical cost-reduction for many users who do not perform stream operations using `not_null` while not imposing an enormous burden on users that do. If someone is inserting into a stream, it's reasonable to assume they have already included `<iostream>`, so this change should have no effect on them.

PR #588, opened 2017-11-17
runs iwyu on the code base

### LINKED ISSUE #583: <gsl/string_span> uses std::add_const without including <type_traits>

tiagomacarios:
I ran it with the --comment flag to get clangs reasoning for the include.

Should fix #583 and supersedes #584.

@paweldac @LegalizeAdulthood wdyt?

### DIFF
--- a/include/gsl/gsl
+++ b/include/gsl/gsl
@@ -20,8 +20,8 @@
 #include <gsl/gsl_assert>  // Ensures/Expects
 #include <gsl/gsl_util>    // finally()/narrow()/narrow_cast()...
 #include <gsl/multi_span>  // multi_span, strided_span...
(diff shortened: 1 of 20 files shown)


tiagomacarios:
I can try to add a travis config for the build to check for this problems

paweldac:
Hi @tiagomacarios

It looks fine to me.

LegalizeAdulthood:
Looks good to me!

tiagomacarios:
@neilmacintosh bump!

neilmacintosh:
Looks good to me @tiagomacarios, thanks!

PR #593, opened 2017-11-28
Rename final_act in GSL.natvis

bbolli:
Commit 6e2e207c8d renamed final_act to final_action. Do this also in
GSL.natvis.


neilmacintosh:
Thanks for chasing this up @bbolli!

PR #509, opened 2017-05-21
Fix for: issue #471 Adds gsl::joining_thread

galik:
This PR adds <gsl/gsl_thread> and tests/thread_tests.cpp

As per discussion in CppCoreGuidelines https://github.com/isocpp/CppCoreGuidelines/issues/925

This PR provides `gsl::joining_thread` (gsl::detached_thread was removed)

PR #596, opened 2017-12-05
Fix for: issue #471 Adds gsl::joining_thread (replaces original PR #509)

galik:
This PR adds <gsl/gsl_thread> and tests/thread_tests.cpp

As per discussion in CppCoreGuidelines https://github.com/isocpp/CppCoreGuidelines/issues/925

This PR provides `gsl::joining_thread` (gsl::detached_thread was removed)

(This PR is a cleaned up version of PR #509 - deleted)

### DIFF
--- a/include/gsl/gsl_thread
+++ b/include/gsl/gsl_thread
@@ -1,0 +1,104 @@
+///////////////////////////////////////////////////////////////////////////////
+//
+// Copyright (c) 2017 Microsoft Corporation. All rights reserved.
+//
+// This code is licensed under the MIT License (MIT).
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+///////////////////////////////////////////////////////////////////////////////
+
+#pragma once
+
+#ifndef GSL_THREAD_H
+#define GSL_THREAD_H
+
+#include <thread>
+
+#ifdef _MSC_VER
+#pragma warning(push)
+
+// turn off some warnings that are noisy about our Expects statements
+#pragma warning(disable : 4127) // conditional expression is constant
+
+// blanket turn off warnings from CppCoreCheck for now
+// so people aren't annoyed by them when running the tool.
+// more targeted suppressions will be added in a future update to the GSL
+#pragma warning(disable : 26481 26482 26483 26485 26490 26491 26492 26493 26495)
+#endif // _MSC_VER
+
+namespace gsl
+{
+
+class joining_thread
+{
+    friend void swap(joining_thread& t1, joining_thread& t2) noexcept;
+
+public:
+    joining_thread() noexcept = default;
+
+    joining_thread(joining_thread const&) = delete;
+    joining_thread(joining_thread&& other) : t(std::move(other.t)) {}
+
+    joining_thread(std::thread const&) = delete;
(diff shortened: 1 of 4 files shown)


include/gsl/gsl_thread
```diff
@@ -0,0 +1,104 @@
...
+
+class joining_thread
+{
+    friend void swap(joining_thread& t1, joining_thread& t2) noexcept;
+
+public:
+    joining_thread() noexcept = default;
+
+    joining_thread(joining_thread const&) = delete;
+    joining_thread(joining_thread&& other) : t(std::move(other.t)) {}
```
paweldac: In joining_thread d-tor you are using t, but with this c-tor is has already been moved.

Consider following:
```c++
{
gsl::joining_thead th1([](){});
gsl::joining_thread th2 = std::move(th1);
} // here scope of both th1 and th2 ends, so all destructors are called. th1 destructor is using already moved thread t.
```

MikeGitb: Nothing forbids you to call `joinable()` on a moved from `std::thread` object.

MikeGitb: Nothing forbids you to call `joinable()` on a moved from `std::thread` object.

MikeGitb: Nothing forbids you to call `joinable()` on a moved from  `std::thread`

MikeGitb:
I have to say, I'd feel much more comfortable with the idea of joining threads, if there was a way to indicate to the thread that termination was requested (something like a gsl:: `joining_thread::request_stop()/ gsl::this_thread::stop_requested()` pair, but that comes with overhead and is probably out of scope for the gsl.

include/gsl/gsl_thread
paweldac: @MikeGitb my bad. Deleted comment not to introduce confusion.

Quuxplusone: I don't know if this PR is still active, but: this move constructor should be `noexcept`, or even better, defaulted.

include/gsl/gsl_thread
```diff
@@ -0,0 +1,104 @@
...
+
+    joining_thread(std::thread const&) = delete;
+    joining_thread(std::thread&& other) noexcept : t(std::move(other)) {}
+
+    joining_thread& operator=(joining_thread const&) = delete;
+    joining_thread& operator=(joining_thread&& other) noexcept
```
Quuxplusone, line 53: Could be `=default`ed.

include/gsl/gsl_thread
```diff
@@ -0,0 +1,104 @@
...
+    {
+        t = std::move(other.t);
+        return *this;
+    }
+
+    joining_thread& operator=(std::thread const&) = delete;
```
Quuxplusone, line 59: This function signature should not exist (it shouldn't be explicitly `=delete`d).

include/gsl/gsl_thread
```diff
@@ -0,0 +1,104 @@
...
+    friend void swap(joining_thread& t1, joining_thread& t2) noexcept;
+
+public:
+    joining_thread() noexcept = default;
+
+    joining_thread(joining_thread const&) = delete;
+    joining_thread(joining_thread&& other) : t(std::move(other.t)) {}
+
+    joining_thread(std::thread const&) = delete;
+    joining_thread(std::thread&& other) noexcept : t(std::move(other)) {}
```
Quuxplusone, line 50: I don't know if we're implementing a formal spec here, but if the design is still up for grabs, I'd be strongly inclined to make this constructor `explicit` — for the same reason that `explicit unique_ptr<T>(T *p)` is explicit.

If you do make it `explicit`, then you should remove the signature `joining_thread& operator=(std::thread&& other) noexcept` below, because we wouldn't want implicit assignment to work either, in that case.

include/gsl/gsl_thread
```diff
@@ -0,0 +1,104 @@
...
+
+    template <typename Callable, typename... Args>
+    explicit joining_thread(Callable&& f, Args&&... args)
+        : t(std::forward<Callable>(f), std::forward<Args>(args)...)
+    {
+    }
+
+    ~joining_thread() { if (t.joinable()) t.join(); }
+
+    bool joinable() const { return t.joinable(); }
```
Quuxplusone, line 74: `noexcept`?

include/gsl/gsl_thread
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+private:
+    std::thread t;
+};
+
+void swap(joining_thread& t1, joining_thread& t2) noexcept
+{
+    using std::swap;
+    swap(t1.t, t2.t);
```
Quuxplusone, line 95: It would be better to keep this as a hidden friend (defined in-line in the class body). Also, the implementation could be just `t1.swap(t2);` for simplicity.

PR #601, opened 2018-01-10
Use in-class init for final_action members (C.48)

tmr232:
Initialize `invoke_` in-class, according to http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c48-prefer-in-class-initializers-to-member-initializers-in-constructors-for-constant-initializers


neilmacintosh:
Thanks for that clean-up @tmr232, that looks like a good change to me.

PR #602, opened 2018-01-14
Fix wrong check of the CMAKE_VERSION

delaitre:
`CMAKE_VERSION_MAJOR` is just the first component of the CMake version (e.g. 3 for 3.10.1) which compares less than `3.7.9` if it is less than 3. The proper variable to compare is `CMAKE_VERSION` which contains the full version number.


neilmacintosh:
This looks good to me. Thanks for picking up on that bug @delaitre!

PR #608, opened 2018-02-02
Don't use clang's __builtin_assume as an optimizer hint

### LINKED ISSUE #607: Clang build fails with GSL_UNENFORCED_ON_CONTRACT_VIOLATION
In our publish builds, we define GSL_UNENFORCED_ON_CONTRACT_VIOLATION. We've updated to the latest GSL and have since been hitting the issue below. I believe the solution is to declare the `span::length()` member function as `__attribute__ ((const))`, but I am not really sure due to me not being familiar with the clang language extensions. I believe this issue is caused by commit 16a6a416. 

We are using Clang 3.8 with the Android r13b NDK. We also have warnings as errors for reference.

[compilerlog.txt](https://github.com/Microsoft/GSL/files/1690385/compilerlog.txt)

CaseyCarter:
Fixes #607.

Drive-by: use '?:' to contextually convert to bool in the fallback definition of `GSL_ASSUME`

### DIFF
--- a/include/gsl/gsl_assert
+++ b/include/gsl/gsl_assert
@@ -52,12 +52,10 @@
 //
 #ifdef _MSC_VER
 #define GSL_ASSUME(cond) __assume(cond)
-#elif defined(__clang__)
-#define GSL_ASSUME(cond) __builtin_assume(cond)
 #elif defined(__GNUC__)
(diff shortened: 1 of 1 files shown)


neilmacintosh:
Thanks @CaseyCarter, I really like this fix!

PR #615, opened 2018-02-11
Remove move ops from span

neilmacintosh:
`span` is designed to be a simple, copy-only type. As a result, there is no need to define (even using defaults) move operations for it. This change removes the unnecessary move constructor and assignment operator. The change is a result of feedback from the C++ standardization process.


CaseyCarter:
This one seems weird to me, *and* doesn't reflect any discussion I recall. (Admittedly I wasn't in the LEWG small group discussion in ABQ, but I did read the minutes.) `span` is one of the simplest value types we have in the standard library: why should it define *any* copy/move SMFs rather than using the implicitly-defined defaults?

neilmacintosh:
There were some discussions that requested the move SMFs disappear. I could dig up the request, but I trusted my memory there as STL also asked for them to go.

But I think what you are really saying is not that you object to removing the move ones, but why are the copy ones still there? It's a fair point...but then this PR would still be ok, right? You'd just want to see more deleting in the future ;-)

CaseyCarter:
Yes - I don't think this change breaks the class, it's the inconsistency that I find disturbing.

neilmacintosh:
Makes perfect sense! I'll address that inconsistency in a future PR then. For now, I've been trying to just check off addressing the existing feedback. Your new feedback has gone into my queue. It is important to me, so please, continue to hold. :-)

gdr-at-ms:
Looks good to me.
Do you need someone to perform the merge?

neilmacintosh:
Thanks @gdr-at-ms! And the offer of a merge is very kind, but it appears I have merge rights once more (yay).

Issue #617, opened 2018-02-12
Add index typedef as per Core Guidelines


neilmacintosh:
https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#gslutil-utilities recently gained an `index` type definition. We should update the GSL to include this.

PR #620, opened 2018-02-15
add gsl::index typedef

paweldac:
renamed existing gsl::index to gsl::span_index in order to resolve name collision.
added gsl::index as described in ES.107: Don't use unsigned for subscripts, prefer gsl::index
https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Res-subscripts

fixes: #617

### DIFF
--- a/include/gsl/gsl_util
+++ b/include/gsl/gsl_util
@@ -43,6 +43,9 @@ namespace gsl
 // GSL.util: utilities
 //
 
+// index type for all container indexes/subscripts/sizes
+using index = std::ptrdiff_t;
+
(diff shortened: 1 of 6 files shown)


neilmacintosh:
Thanks for this @paweldac, it looks great. I think `multi_span_index` might be a clearer name than `span_index` if we're here changing it anyway. Would you mind making that adjustment, or would you prefer that I do it before accepting the rest of the change?

paweldac:
@neilmacintosh rename is done ;-)

neilmacintosh:
Thanks @paweldac this looks great now. It adds another missing element to the GSL.

Issue #621, opened 2018-02-16
gsl::copy => compiler error C4996


Ziriax:
The following code 

```
	float s[] = { 1,2,3 };
	float t[] = { 4,5,6 };
	gsl::copy(gsl::make_span(s, 3), gsl::make_span(t, 3));
```

gives the compiler error (Visual Studio 2017 15.5.6 toolset v141, GSL code from 14-feb-2018, SHA c87c123d1b3e64ae2cf725584f0c004da4d90f1c)

```
Error C4996 'std::copy_n::_Unchecked_iterators::_Deprecate': Call to 'std::copy_n' with parameters that may be unsafe - this call relies on the caller to check that the passed values are correct. To disable this warning, use -D_SCL_SECURE_NO_WARNINGS. 
```

The above is just dummy code to reproduce my problem.

What is the correct way to call `gsl::copy`?

If I look at the implementation, it already performs the following check:

```
    Expects(dest.size() >= src.size());
```

So why do I get error C4996?

Thanks a lot,
Peter Verswyvelen

beinhaerter:
For more information you can read https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-3-c4996#unsafe-standard-library-functions.
STL does not know that GSL is doing it right, so it emits the warning.
An easy solution for you is already written in the warning: "To disable this warning, use -D_SCL_SECURE_NO_WARNINGS."

Other solutions would require a change in the VS STL or in GSL.

Regards,
Werner

Ziriax:
Thanks for the quick reply.

Yes, I'm aware that defining the preprocessor symbol `_SCL_SECURE_NO_WARNING` suppresses these warnings, but I like have these safety checks in debug builds. 

I was assuming that `gsl::copy` was specifically made for handling this. After all, I first called `std::copy` and obviously also got the error. 

Shouldn't the  `gsl::copy` implementation use `stdext::make_checked_array_iterator` when it is compiled using the MSVC compiler? That would fix it no?

beinhaerter:
Good idea. The following patch fixes it for me. Maybe some MS official can comment if this is the right way to do it and if I should create a pull request.

diff --git a/include/gsl/gsl_algorithm b/include/gsl/gsl_algorithm
index 710792f..b7c704b 100644
--- a/include/gsl/gsl_algorithm
+++ b/include/gsl/gsl_algorithm
[...]

PR #622, opened 2018-02-16
#621 on MSVC use copy_n with stdext::make_checked_array_iterator

beinhaerter:
This solves issue #621


neilmacintosh:
Thanks @beinhaerter but I think that #633 manages an even simpler fix, so I'll close this PR out in favor of that one.

PR #624, opened 2018-02-20
gsl::index bug fix

calebillman:
Fixing a bug in the != operator in the gsl::index class


neilmacintosh:
Nice catch @calebillman, that looks good to me!

PR #625, opened 2018-02-21
Fix return type of templated span.subspan()

neilmacintosh:
This PR modifies the return type of the templated version of `span::subspan()` so that it returns a fixed-sized span wherever possible. This is important for encouraging optimization of code that uses `subspan()`.

Writing the tests for this made me actually use `span::extent` which made me realize that it does not work unless you are using a C++17-compliant compiler, so I also added a workaround to make `span::extent` available on C++11/14 compilers until we can all live in the better world of C++17. ;-)

### DIFF
--- a/include/gsl/span
+++ b/include/gsl/span
@@ -345,7 +345,7 @@ public:
 
     using size_type = index_type;
 
-#if defined(GSL_USE_STATIC_CONSTEXPR_WORKAROUND>)
+#if defined(GSL_USE_STATIC_CONSTEXPR_WORKAROUND)
     static constexpr const index_type extent { Extent };
 #else
     static constexpr index_type extent { Extent };
--- a/include/gsl/span
+++ b/include/gsl/span
@@ -345,11 +345,8 @@ public:
 
     using size_type = index_type;
 
-#if defined(GSL_USE_STATIC_CONSTEXPR_WORKAROUND)
-    enum : index_type { extent = Extent };
-#else
     static constexpr index_type extent { Extent };
-#endif
+
     // [span.cons], span constructors, copy, assignment, and destructor
     template <bool Dependent = false,
               // "Dependent" is needed to make "std::enable_if_t<Dependent || Extent <= 0>" SFINAE,
@@ -549,6 +546,11 @@ private:
         return { data() + offset,  count, true };
     }
 };
+
+#if defined(GSL_USE_STATIC_CONSTEXPR_WORKAROUND)
+template <class ElementType, std::ptrdiff_t Extent>
+constexpr typename span<ElementType, Extent>::index_type span<ElementType, Extent>::extent;
+#endif
 
 
 // [span.comparison], span comparison operators
--- a/include/gsl/span
+++ b/include/gsl/span
@@ -45,8 +45,17 @@
 #if _MSC_VER < 1910
 #pragma push_macro("constexpr")
 #define constexpr /*constexpr*/
+#define USE_STATIC_CONSTEXPR_WORKAROUND
 
 #endif                          // _MSC_VER < 1910
+#else                           // _MSC_VER
+
+// See if we have enough C++17 power to use a static constexpr data member
+// without needing an out-of-line definition
+#if !(defined(__cplusplus) && (__cplusplus >= 201703L))
+#define GSL_USE_STATIC_CONSTEXPR_WORKAROUND
+#endif // !(defined(__cplusplus) && (__cplusplus >= 201703L))
+
 #endif                          // _MSC_VER
 
 #ifdef GSL_THROW_ON_CONTRACT_VIOLATION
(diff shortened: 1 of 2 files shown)


include/gsl/span
```diff
@@ -45,8 +45,17 @@
 #if _MSC_VER < 1910
 #pragma push_macro("constexpr")
 #define constexpr /*constexpr*/
+#define USE_STATIC_CONSTEXPR_WORKAROUND
```
CaseyCarter: `GSL_USE_STATIC_CONSTEXPR_WORKAROUND`

include/gsl/span
```diff
@@ -330,8 +345,11 @@ public:
 
     using size_type = index_type;
 
-    constexpr static const index_type extent = Extent;
-
+#if defined(GSL_USE_STATIC_CONSTEXPR_WORKAROUND)
```
CaseyCarter: This is an odd choice. Why not simply provide the out-of-class definition for C++ < 17?

include/gsl/span
```diff
@@ -309,6 +318,12 @@ namespace details
     private:
         index_type size_;
     };
+
+    template <class ElementType, std::ptrdiff_t Extent, std::ptrdiff_t Offset, std::ptrdiff_t Count>
+    struct calculate_subspan_type
+    {
+      using type = span<ElementType, Count != dynamic_extent ? Count : (Extent != dynamic_extent ? Extent - Offset - 1 : Extent)>;
```
CaseyCarter: `Extent - Offset - 1` means that given:
```c++
int data[5]{};
span<int, 5> s = data;
```
`decltype(s.subspan<5>())` is `span<int, -1>` instead of `span<int, 0>`, and `decltype(s.subspan<0>())` is `span<int, 4>` instead of `span<int, 5>`. Surely it should be an invariant that `s.subspan<0>() == s` for a span `s`. This expression should be `Extent - Offset`.

tests/span_tests.cpp
```diff
@@ -776,35 +776,47 @@ TEST_CASE("subspan")
...
+        CHECK(decltype(av.subspan<4, 0>())::extent == 0);
         CHECK(av.subspan(4, 0).size() == 0);
         CHECK(av.subspan(5, 0).size() == 0);
         CHECK_THROWS_AS(av.subspan(6, 0).size(), fail_fast);
     }
 
+    {
+        span<int, 5> av = arr;
+        CHECK((av.subspan<1>().size() == 3));
+        CHECK(decltype(av.subspan<1>())::extent == 3);
```
CaseyCarter: Ditto for these two lines: `decltype(av.subspan<1>())::extent` should be `4`.

include/gsl/span
CaseyCarter: See https://github.com/ericniebler/range-v3/blob/master/include/range/v3/span.hpp#L213 =)

neilmacintosh: You are completely right. Nice catch. That will teach me for working at night.

tests/span_tests.cpp
neilmacintosh: Good point! My brain clearly had extent and size confused.

include/gsl/span
neilmacintosh: From memory: one of the compilers gave me a complaint about it. Plus, this seemed like the very well-known-but-ugly moral equivalent of the C++17 constexpr (i.e. it has a very good chance of actually getting optimized away). It's only a temporary band-aid, so I'm not overly attached, but nor did I want to spend too much time perfecting it.

include/gsl/span
neilmacintosh: Oops! Thanks.

include/gsl/span
CaseyCarter: Fair enough. I'd suggest fixing the underlying type of the enum so `extent` always has the same type:
```c++
enum : index_type { extent = Extent };
```

neilmacintosh: Yep, sounds good!

include/gsl/span
neilmacintosh: In passing...I took a look at the ranges implementation and it doesn't seem to match the intended design for the templated `subspan`: that people can pass a `dynamic_extent` value for the second template parameter to mean "consume everything that remains". (symmetrical to the way the non-template `subspan` works). Is that something you guys chose to do explicitly, or just an outcome of the implementation choice?

CaseyCarter: Oh, and thanks for the bug report!

neilmacintosh: Hahahaha. I thought about the lack of test coverage just after posting the comment....I had a distinct memory you'd run these tests against the implementation...so it's all really my fault (spoiler alert: most things are).

neilmacintosh: And thanks for all the help! You came for the review, stayed for the bug report!

PR #627, opened 2018-02-23
constexpr for make_span

beinhaerter:
VS 2017 warns me that make_span could be made constexpr.
This PR gets rid of the warning.


neilmacintosh:
Thanks @beinhaerter, this looks good to me! And great to hear that CppCoreCheck in VS 2017 is helping to promote "`constexpr` all the things!".

PR #628, opened 2018-02-25
=default

beinhaerter:
clang tidy suggest making the two ctors with an empty body ("{}") "=default;".
I am not sure if there is any more benefit than being explicit.
And I hope it does not break anything (maybe only on some compilers or because of the combination with constexpr and noexcept).

### DIFF
--- a/include/gsl/multi_span
+++ b/include/gsl/multi_span
@@ -99,7 +99,7 @@ public:
     using reference = std::add_lvalue_reference_t<value_type>;
     using const_reference = std::add_lvalue_reference_t<std::add_const_t<value_type>>;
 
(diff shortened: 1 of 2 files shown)


beinhaerter:
I am closing this pull request because it crashes VS2015:
Internal Compiler Error in C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\bin\CL.exe.  You will be prompted to send an error report to Microsoft later.

neilmacintosh:
Thanks beinhaerter. @AndrewPardoe you might be interested in this ICE report.

AndrewPardoe:
Thanks for the heads-up!

PR #630, opened 2018-02-27
make explicit not_null move constructor so it is noexcept

xaxxon:


xaxxon:
Makes move constructor noexcept as well as skipping the null check when you're moving from an object you already know has been checked.

xaxxon:
Original impetus for this change was that I had a class I wanted to put in tsl::hopscotch_map and that type's move constructor wasn't noexcept because not_null wasn't noexcept.

neilmacintosh:
Seems very reasonable to me. Thanks for the PR @xaxxon!

PR #631, opened 2018-03-02
Compare span_iterators to the same data

suncho:
For example, when span1 and span2 refer to the same data, this allows you to use begin(span1) and end(span2) to form a valid range.  It's useful when you're creating identical temporary spans just for the purpose of getting the iterators out:

`copy(begin(make_span(weird_container), end(make_span(weird_container), begin(some_vector));`

### DIFF
--- a/include/gsl/span
+++ b/include/gsl/span
@@ -136,13 +136,13 @@ namespace details
 
     private:
         constexpr span_iterator(const Span* span, typename Span::index_type idx) noexcept
-            : span_(span), index_(idx)
+            : base_(span->data()), index_(idx), size_(span->size())
         {}
     public:
 
         template<bool B, std::enable_if_t<!B && IsConst>* = nullptr>
         constexpr span_iterator(const span_iterator<Span, B>& other) noexcept
-            : span_iterator(other.span_, other.index_)
+            : base_(other.base_), index_(other.index_), size_(other.size_)
         {
         }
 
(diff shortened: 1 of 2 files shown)


suncho:
span_iterators that refer to the same memory compare neither greater than nor less than each other.  If they correspond to spans of different sizes, they do not compare equal to each other either.

A consequence is that the set of span_iterators that compare neither less than nor greater than each other form an equivalence class.  This equivalence class is a superset of the equivalence class defined by the equality operator.

This is how how it should be.  It ensures that span_iterators are regular.  Imagine two span_iterators that are equivalent according to the less-than operator, but that are associated with spans of different lengths.    The two iterators will be bounds-checked differently.  If the iterators compared equal, then the bounds-checking procedures would not be regular on the span_iterator type.

neilmacintosh:
@suncho This PR doesn't look desirable to me. We do not want to encourage people to compare and mix iterators from different spans, regardless of whether or not they refer to the underlying data. This issue has been discussed before, if I recall correctly. It encourages practices that might just happen to work because of implementation details.

suncho:
@neilmacintosh Hmm.  What kinds of undesirable practices does it encourage people to do?  Can you point me to a previous discussion?

I've found this functionality to be incredibly useful in my own work as I attempt to clean up legacy code and make safer and easier to read.

See this gist:
https://gist.github.com/suncho/4ccaeb9c69dfef9ffdc0b712f08d85fa

In particular, the `wipe` function template will fail to compile if I replace:

```
template <class T>
void wipe(T& object)
{
  fill(begin(rep(object)), end(rep(object)), byte(0));
}
```

with this:

```
template <class T>
void wipe(T& object)
{
  fill(begin(rep_span(object)), end(rep_span(object)), byte(0));
}
```

This is because the two calls to `rep_span` produce different (but equivalent) spans.  Under the current Microsoft GSL implementation, the `begin` iterator is incompatible with the `end` iterator.

I use this pattern throughout my code when working with byte representations of objects.  This way, I can ensure that my byte iterators are range checked.  But it only works with spans if we allow iterators from equivalent spans to be compared against one another.

annagrin:
Closing this as the Guidelines discussion desided to go with range-based API.

PR #632, opened 2018-03-04
fixed noexept warnings

annagrin:
- Removed conditional compilation for throwing version of GSL
  vs fail_fast because we don't want users of the code to see
  differences in the span interface dependent on error mechanism
  chosen
- Removed noexcept from functions that may fail at runtime
- Fixed CppCoreCheck warnings related to missing and incorrect
  noexcept


neilmacintosh:
Looks good to me @annagrin, thanks!

PR #636, opened 2018-03-08
Added test case to catch known GCC bug.

neilmacintosh:
This PR just adds a small test case for `span`, helpfully provided by @ggajoch in his report of certain `span` operations being affected by a GCC bug. Having the test should help anyone running an affected compiler identify the problem.


annagrin:
Thanks @neilmacintosh! Test was added #704 - closing.

PR #637, opened 2018-03-09
Improve GSL postconditions and add invariants

suncho:
This update implements `Ensures` so that the check fires when the scope exits.  Now you can put your `Ensures` statements at the top of your function definition and it will apply to all possible return paths.

It also adds `AlwaysEnsures`, which does the check even when exiting via exception, in which case the program will terminate on a double exception.

Similarly, I've added `Maintains` and `AlwaysMaintains` for asserting invariants that hold both at the beginning and the end of the function.

### DIFF
--- a/include/gsl/gsl_assert
+++ b/include/gsl/gsl_assert
@@ -88,6 +88,58 @@ struct fail_fast : public std::logic_error
 #endif
 
 #define Expects(cond) GSL_CONTRACT_CHECK("Precondition", cond)
-#define Ensures(cond) GSL_CONTRACT_CHECK("Postcondition", cond)
+
+#if defined(GSL_UNENFORCED_ON_CONTRACT_VIOLATION)
+
+#define Ensures(cond)         static_cast<void>(0)
+#define AlwaysEnsures(cond)   static_cast<void>(0)
+
+#else // GSL_UNENFORCED_ON_CONTRACT_VIOLATION
+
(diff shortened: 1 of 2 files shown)


suncho:
Of course all the tests are failing.  I'll have a look at it tomorrow.

neilmacintosh:
Thanks for the PR @suncho. However, I'm afraid I'm not particularly inclined to accept this PR. The GSL is designed as part of the C++ Core Guidelines Effort. Although we do have some helper functions that aren't described in the Guidelines, we generally don't just redesign a core component described in the Guidelines (like `Expects`/`Ensures`) or add a new one like `AlwaysEnsures`\`Maintains`\`AlwaysMaintains`.

I don't mean to reduce your enthusiasm, though. Can I suggest the right approach here is to open an issue over at the Core Guidelines repository describing why you think the Ensures/Expects described there is lacking and your proposed interface would be a good improvement. If the editors review the issue and decide to make a change, then there'll be an opportunity to flow the changes back into the library. 

So for now, I'll close out this PR as it does not match up with the description in the Core Guidelines.

suncho:
Will do.  Thanks for the suggestion.

dorbeus:
Det e sa på plats och jag är. 2 det X på å på ååpååååpp på ååpååååpp KGB @:8

Den 9 mars 2018 02:23 skrev "Alex Howlett" <notifications@github.com>:

PR #639, opened 2018-03-11
simplified constructors and removed GSL_NOEXCEPT

annagrin:
- simplified storage constructors to reduce branching
- eliminated null check for constructor from array
- removed GSL_NOEXCEPT from span

### DIFF
--- a/include/gsl/span
+++ b/include/gsl/span
@@ -58,8 +58,6 @@
 
 #endif                          // _MSC_VER
 
-#define GSL_NOEXCEPT noexcept
-
 namespace gsl
 {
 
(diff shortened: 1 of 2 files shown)


beinhaerter:
As far as I understand you are using KnownNotNull to get rid of some nullptr checks.
For the case where span gets a reference to a C style array, wouldn't it be better to use gsl::not_null and to add a gsl::not_null ctor from "element_type (&arr)[N]"? That would enable your check elimination for all users of gsl::not_null, not only for gsl::span.
I am not quite sure what to do about the ctors taking a std::array. Maybe gsl::span could retrieve the data pointer with std::array:data() and convert it to a C style array reference that could be passed to my proposed gsl::not_null ctor?

annagrin:
You are right, I'd like to optimize away a few unnecessary nullptr checks that cause particularly branch-heavy code if you use subspans. Array reference is a nice side benefit of the same strategy. 
Using the same strategy in case of std::array is my mistake - zero length std::arrays are legal and getting their data is unspecified behavior. Will remove.

Will think about a more generic solution for not_null, thanks for pointing that out!

neilmacintosh:
LGTM @annagrin!

Issue #635, opened 2018-03-07
GSL Fails to Compile with Exceptions Disabled


aheysse:
gsl::narrow will fail to compile on clang with exceptions disabled. It doesn't seem to honor 

GSL_UNENFORCED_ON_CONTRACT_VIOLATION or
GSL_THROW_ON_CONTRACT_VIOLATION or
GSL_UNENFORCED_ON_CONTRACT_VIOLATION,

but rather just blindly throws an exception. I assume GSL_CONTRACT_CHECK is meant to be used.

For good measure, here's the compiler error:
1>D:\Github\handheld\src-external\gsl\include\gsl/gsl_util(109,33): error : cannot use 'throw' with exceptions disabled
1>    if (static_cast<U>(t) != u) throw narrowing_error();
1>                                ^
1>D:\Github\handheld\src-external\gsl\include\gsl/gsl_util(111,9): error : cannot use 'throw' with exceptions disabled
1>        throw narrowing_error();
1>        ^

aheysse:
@neilmacintosh Thanks for the quick update. Our team is in the process of taking GSL as a submodule so I look forward to the fix. Thx!

annagrin:
Note that _HAS_EXCEPTIONS=0 in MS STL is not an official story for no exception mode. Current resolution of the issue is experimental only, until we have an official support for no-exception mode in the language and STL.

JordanMaples:
Maintainer's call: It looks like this issue has already been handled, however I'm closing the issue as the contract violation behavior has further changed to only result in termination.

phillipjohnston:
Hi @JordanMaples,

Is this actually closed? I still get this error with GSL 3.1.0:

```
In file included from ../subprojects/GSL-3.1.0/include/gsl/gsl:23:
../subprojects/GSL-3.1.0/include/gsl/gsl_util:110:33: error: cannot use 'throw' with exceptions disabled
    if (static_cast<U>(t) != u) throw narrowing_error{};
                                ^
../subprojects/GSL-3.1.0/include/gsl/gsl_util:112:9: error: cannot use 'throw' with exceptions disabled
        throw narrowing_error{};
        ^
```

Looking at the narrow function, it doesn't seem like there's any condition that causes the `throw` to be avoided.

```
// narrow() : a checked version of narrow_cast() that throws if the cast changed the value
template <class T, class U>
GSL_SUPPRESS(type.1) // NO-FORMAT: attribute
[...]

PR #640, opened 2018-03-12
Enable usage of gsl::narrow with exceptions disabled

annagrin:
This change is based on the previous PR by @djarek:

This solution uses the approach of boost::asio to enabling usage of the
library in environments where exception usage is either prohibited
or not feasible (due to code size constraints). A function template
gsl::throw_exception has been added, which in a normal environment just
throws the exception. However, when GSL_TERMINATE_ON_CONTRACT_VIOLATION
is defined the function is only declared by gsl and the definition of
this function template must be supplied by the library's user.

Closes: #468

Signed-off-by: Damian Jarek <damian.jarek93@gmail.com>

Addition:
- made sure narrow() terminates only in GSL_TERMINATE_ON_CONTRACT_VIOLATION 
  mode, and throws in all others
- added work around for MS STL to ensure termination in no-exception mode
       - understand MSVC STL no exception macro
      -  use function static variable to set termination handler in no-exception mode (msvc)
- add compile-only tests for no-exception mode


include/gsl/gsl_assert
```diff
@@ -20,6 +20,13 @@
 #include <exception>
 #include <stdexcept> // for logic_error
 
+//
+// Temporary until MSVC STL supports no-exceptions mode.
+// Currently terminate is a no-op in this mode, so we add termination behavior back
+//
+#if defined(_MSC_VER) && !_HAS_EXCEPTIONS
```
neilmacintosh: I think you want to test if _HAS_EXCEPTIONS is `#define`d before testing its value. People catch me making that mistake all the time.

tests/no_exception_tests.cpp
```diff
@@ -0,0 +1,42 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+///////////////////////////////////////////////////////////////////////////////
+
+#include <gsl/gsl_util> // for narrow_cast, at
+#include <gsl/span>     // for span, span_iterator, operator==, operator!=
+
+#include "test_noexcept.hpp"
```
neilmacintosh: Everywhere else we use `.h` I think. Any reason to not follow suit here and have `test_noexcept.h`?

tests/test_noexcept.cpp
```diff
@@ -0,0 +1,21 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+///////////////////////////////////////////////////////////////////////////////
+
+#include "test_noexcept.hpp"
+int main()
```
neilmacintosh: Wouldn't you want to install a replacement to the usual "terminate" behavior in this test so you can ensure each case  ends up at the terminate-handler as expected without having to actually kill the process?

neilmacintosh:
I really like the change overall, thanks for picking up @djarek's work here @annagrin and finishing it up. I left a couple of comments to consider but (apart from the `#if`-related one) you should feel free to merge and get to them at your leisure.

tests/test_noexcept.cpp
annagrin: yes!

include/gsl/gsl_assert
annagrin: sure

tests/no_exception_tests.cpp
annagrin: copied from catch.hpp, but no reason to do it here

tests/test_noexcept.cpp
annagrin: I am not using the catch infrastructure, need to think about the best way to implement this, so will resolve it in a later PR

annagrin:
Addressed #if-related and .hpp comments. Will implement testing for no-exception mode separately.

neilmacintosh:
LGTM, thanks @annagrin!

annagrin:
this closes #468 and  #635

PR #641, opened 2018-03-15
Revert CMake Minimum

rianquinn:
This patch reverts the CMake minimum back to 3.1.3, re-enabling support for Cygwin, without effecting the original intention. 

Signed-off-by: Rian Quinn <rianquinn@gmail.com>


CaseyCarter:
Question: do you think it makes sense to update the Travis configuration so as to install the same cmake version which we are at minimum requiring?

rianquinn:
Probably. We do that with our testing as well. I will make that change quick.

annagrin:
@rianquinn let's try:)

PR #645, opened 2018-03-22
Make not_null work with move-only types like unique_ptr

xaxxon:
Changed so you can have a not_null<unique_ptr<T>>.   All existing unit tests pass except for CustomPtr which didn't allow dereferencing a const CustomPtr, which seemed incorrect, so I added that behavior to CustomPtr and now all tests pass.

Checked that returning references for primitive types doesn't impact performance:  

https://godbolt.org/g/32QkEX

exact same ASM generated between modified version and current version.

### DIFF
--- a/include/gsl/pointers
+++ b/include/gsl/pointers
@@ -82,6 +82,7 @@ public:
     {
     }
 
+    not_null(not_null&& other) = default;
     not_null(const not_null& other) = default;
     not_null& operator=(const not_null& other) = default;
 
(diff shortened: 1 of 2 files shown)


xaxxon:
The reason I made this change is I was refactoring a large number of classes with a PIMPL member and I was adding new constructors to each and wanted to make sure I didn't miss allocating for the pimpl member unique_ptr data.   While the existing not_null implementation technically worked for that, the resulting object was unusable because all accesses required returning a copy.

xaxxon:
Sorry about the merge conflict.. I don't know how to resolve a conflict between my change and.. nothing.

Issue #604, opened 2018-01-23
Checks in not_null.


john-preston:
Current implementation of not_null does checks (by Expects and Ensures):

- in every constructor (even if constructed from other not_null)
- in every dereference using any method (all go through get())

This could be an overhead in case you use not_null quite a lot and pass it everywhere by value and leave the Expects()/Ensures() checks even in the Release build (my case). Is this intended or should the check really be only in raw-pointer constructor?

I don't see any sense in those additional checks for not_null<raw_pointer> types, because they can't suddenly become null in the middle of their lifetime. If this is not intended I can prepare a PR with single check. It also should fix #535 because currently copy constructor is not used for non-const_ref not_null copies.

xaxxon:
This is addressed in this PR https://github.com/Microsoft/GSL/pull/650

beinhaerter:
Is there a reason to keep this open? As far as I can see it is solved.

```
    not_null(const not_null& other) = default;
    constexpr details::value_or_reference_return_t<T> get() const { return ptr_; }
```

Issue #550, opened 2017-08-15
not_null<shared_ptr> overhead


cconverse711:
What is the intended usage of not_null with std::shared_ptr? There is a significant overhead to wrapping std::shared_ptr in not_null because every access results in a costly copy of the shared_ptr.  As not_null cannot be used with unique_ptr, it seems that not_null is not really usable with smart pointers in general.

janwilmans:
@neilmacintosh can this issue #550, #415 and #89 be closed now that not_null<smartptr> works?

asfernandes:
> @neilmacintosh can this issue #550, #415 and #89 be closed now that `not_null<smartptr>` works? or is the support still debated?

Is the problem resolved or not?

jwakely:
Using `shared_ptr` still results in lots of unnecessary copies, so I don't think this is resolved at all.

C++17 guaranteed elision helps with some of them, e.g. `not_null<shared_ptr<T>>::operator->` will be less expensive, but `operator*` still makes an unnecessary copy by calling `get()` (apparently just to avoid repeating the `Expects` check in `operator*`).

hsutter:
Thanks for pinging this old issue, sorry for the lag.

I agree that’s reasonable, though I thought “POD” was deprecated these days in favor of “trivially copyable”? FWIW, personally, if I were writing the code I’d write this general type alias (name can be bikeshedded; this is copied from cppfront where I use this for "in" parameters and it's called [`in<T>`](https://github.com/hsutter/cppfront/blob/main/include/cpp2util.h#L365)):

```
template<typename T>
using value_access_t =
    std::conditional_t <
        sizeof(T) < 2*sizeof(void*) && std::is_trivially_copy_constructible_v<T>,
        T const,
        T const&
    >;
```

which is generally useful, and change `get()`’s return from

```
std::conditional_t<std::is_copy_constructible<T>::value, T, const T&>
```

to 

```
value_access_t<T>
```

Note this also solves the slight incongruence that right now the return value is sometimes `const` and sometimes not... here it's always `const`.

Caveat: I haven't tried it on `not_null`'s `.get()`.

What do you think?

daschuer:
What is your reason for `sizeof(T) < 2*sizeof(void*)` ? 
Or will just `std::is_pointer` work?

hsutter:
@daschuer Good question!

[...]

Issue #325, opened 2016-08-14
not_null and std::function


davidhunter22:
Has there been any discussion of having a specialization of not_null for std::function. So 
`not_null<std::function<blah>>`
would imply the std::function contains some non null function pointer.

Also has not_null been put for standardization?

neilmacintosh:
Hi @davidhunter22. There has been no previous talk of specializing for `std::function`, though it seems like a pretty reasonable idea at first glance....so consider it on the table for discussion!

We have not put together a proposal to standardize `not_null` yet. I think there are still some parts of the design that need refinement (see the issues on this repo for a quick summary of some) before it could be considered ready to take to WG21. Plus, my personal queue has been full with `span` and `byte` (GSL and guidelines are only supposed to be a small proportion of my work time).

With that said, I think once we get it a little more polished it should be one of the next types we try to standardize.

xaxxon:
it seems that I have a PR already in that makes not_null work with std::function (if it didn't already).   I added unit tests to check for std::function compatibility to my existing PR to make sure it continues to work.   The syntax isn't the nicest not_null_std_function.get()(parameters, go, here); but it does work.

JordanMaples:
Maintainers' call: Sorry for the delay in replying to this issue. The design of `not_null` is for pointers to objects and for standard smart pointers, which do not support pointers to functions. For any changes to be made to not_null, an extension request for not_null needs to be made in the Core Guidelines. Thanks.

PR #650, opened 2018-03-30
remove null check during access and allow non-copyable pointer types like unique_ptr

xaxxon:
as discussed here: https://github.com/Microsoft/GSL/issues/649

the constructor/assignment operators already check that ptr_ is never set to null, so a runtime check during reads isn't necessary.

fixes https://github.com/Microsoft/GSL/issues/604  fixes https://github.com/Microsoft/GSL/issues/550 fixes https://github.com/Microsoft/GSL/issues/325 fixes https://github.com/Microsoft/GSL/issues/89 fixes https://github.com/Microsoft/GSL/issues/651

### DIFF
--- a/include/gsl/pointers
+++ b/include/gsl/pointers
@@ -68,10 +68,9 @@ using owner = T;
 template <class T>
 class not_null
 {
-public:
-    static_assert(std::is_assignable<T&, std::nullptr_t>::value, "T cannot be assigned nullptr.");
 
-    template <typename U, typename = std::enable_if_t<std::is_convertible<U, T>::value>>
+public:
+    template <typename U, typename = typename std::enable_if<std::is_constructible<T, U>::value>::type>
     constexpr not_null(U&& u) : ptr_(std::forward<U>(u))
     {
         Expects(ptr_ != nullptr);
@@ -86,7 +85,8 @@ public:
     not_null(const not_null& other) = default;
     not_null& operator=(const not_null& other) = default;
 
-    constexpr T get() const
+    
+    constexpr T const & get() const
     {
 	/*
 	  not_null constructors and assignment operators always verify ptr_ is
@@ -97,8 +97,8 @@ public:
         return ptr_;
     }
 
-    constexpr operator T() const { return get(); }
-    constexpr T operator->() const { return get(); }
+    constexpr operator T const & () const { return get(); }
+    constexpr T const & operator->() const { return get(); }
     constexpr decltype(auto) operator*() const { return *get(); } 
 
     // prevents compilation when someone attempts to assign a null pointer constant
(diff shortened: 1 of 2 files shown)


xaxxon:
Updated pull request to include changes for supporting unique_ptr as discussed here:

https://github.com/Microsoft/GSL/issues/651

xaxxon:
I'm not sure what the point of the static_assert that you could assign nullptr_t but it was blocking what seemed to be valid test cases, so I removed it.

Also, the is_convertible was also stopping a better construction path so I updated that, as well.

include/gsl/pointers
```diff
@@ -68,10 +68,9 @@ using owner = T;
 template <class T>
 class not_null
 {
-public:
-    static_assert(std::is_assignable<T&, std::nullptr_t>::value, "T cannot be assigned nullptr.");
```
ericLemanissier: cppcoreguidelines says in http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#gslview-views the following about not_null<T : "T can be any type for which ==nullptr is meaningful.", so I guess it would make sense to static_assert on this kind of type trait instead : https://ideone.com/nuYfaB

xaxxon:
Yes, that test would be correct.   However, I think there's a certain
elegance to how small the code for not_null is and you still get a
compilation error without the static_assert, just not as nice of a
message.   I'd rather not deal with deciding if introducting type traits in
<pointer> is the right thing to do during this PR.  Removing the incorrect
static assert clearly makes not_null better, so I'd like to leave it like
that for now.

--Zac

neilmacintosh:
@xaxxon this PR makes two separate changes. Can you separate them out, so that we are not forced to take one to get the other? The part of the change that is easy to accept is removing the `Ensures` (as discussed in the relevant issue). There is still some discussion going on support for smart pointers.

xaxxon:
Do you know how to do that in github?  I didn’t merge them intentionally it
just happened and you can’t have two forks of the same repo on GitHub from
what I can tell.

xaxxon:
Also, my interest in putting more effort into this PR is directly tied to the interest this project puts into not breaking my existing code.  I've spent a lot of time putting together benchmarks and such and am concerned that https://github.com/Microsoft/GSL/pull/659 is going to make me put a lot of time into adjusting my API to the point where I will be (again) deciding if I should just use my own fork.

xaxxon:
wrong button.

beinhaerter:
> Do you know how to do that in github?

@xaxxon You do one fork of the GSL repository, then create two (or more) branches in your fork. For each branch you can create a pull request. If you want so see this in action you can look at my fork https://github.com/beinhaerter/GSL with three branches (https://github.com/beinhaerter/GSL/branches), each for a different PR (https://github.com/Microsoft/GSL/pulls/beinhaerter).

xaxxon:
closing all my issues/PRs because GSL isn't at the maturity level of a library I can depend on.

PR #652, opened 2018-04-05
Fix gcc build

AntonBikineev:
This fixes build with gcc (at least 7.2.0)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -160,17 +160,16 @@ matrix:
     - env: COMPILER=g++-6 BUILD_TYPE=Release
       addons: *gcc6
 
-    # Currently fails due to #525
-    ## GCC 7
-    #- env: COMPILER=g++-7 BUILD_TYPE=Debug
-    #  addons: &gcc7
-    #    apt:
-    #      packages: g++-7
-    #      sources:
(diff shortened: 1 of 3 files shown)


menete:
Refering to #525 , #526 Is there any consensus on how to proceed with this issue ?

gdr-at-ms:
The template parameters should be kept as `ptrdiff_t`; the problem is the compiler; aren't there GCC-specific ways of telling it to stop complaining here?

menete:
Sure if the goal is just to silence the compiler there is
```
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wsign-conversion"
```
as well as
```
#pragma GCC diagnostic pop
```
If those are applied on top/bottom of the files `span` and `multi_span` gcc compiles the tests withou a warning.
Of course the pragmas can also be applied a bit more selectively.

menete:
@gdr-at-ms #690 implements a solution based on the pragmas above, is this the kind of approach you where proposing?

gdr-at-ms:
@menete Yes, that looks good to me.

annagrin:
fixed in #690 . Thanks all!

Issue #395, opened 2016-10-12
not_null has converting constructors?


akrzemi1:
It seems to me that offering a conversion from `T` to `not_null<T>` looses the potential to detect certain bugs at compile-time.

if I have a function that returns a raw (potentially null) pointer, and I carelessly pass it to a function taking not_null it will compile fine, and will try report a bug at run-time when it is likely too late.

Instead, if the constructor from `T` were explicit, an inadvertent assignment:

``` c++
use_ptr(make_ptr());
```

would be impossible, and I would be forced to explicitly require a potentially unsafe conversion:

``` c++
use_ptr(not_null<T>{make_ptr()});
```

This would be a kind of the signature: by writing this cast, I am taking the responsibility for guaranteeing that the raw pointer will not be null. If it is not the case, you will know that I did it consciously.

See also CppCoreGuidelines issue: https://github.com/isocpp/CppCoreGuidelines/issues/767

Amaroker:
I agree with Andrzej.

It could at least be optional to force explicit conversions with something like:

#if defined(GSL_NOT_NULL_EXPLICIT_CONVERSION)

viboes:
+1
It is up to the user to ensure that the pointer is not null.

In other words if had both not_null and not_null_explicit, I will use not_null_explicit.

MikeGitb:
Whatever the solution: Please don't make its behavior dependent on some macro. I'm dreading the day when I include two external libraries, where each of them is relying on a different behavior.

gdr-at-ms:
@MikeGitb Full agreed.

MikeGitb:
There is a related PR from last year: #83

neilmacintosh:
@akrzemi1 you make a good point. I think the principle to apply here is not whether or not we can catch more errors at compile-time. After all, the sorts of errors the compiler could catch this way are going to be relatively obvious, and the change would not substantively affect what other compile-time tooling could catch.

What convinces me here is a broader issue. If someone applies `not_null` to a parameter on an existing function, then they have effectively strengthened the requirements for calling that function. That should be a "breaking change" for callers - forcing them to ensure that their call sites meet the new contract. The best way to do that is to force the conversions to be explicit.

As @MikeGitb observes, there was a previous PR last year that got lost amidst conflicts with other changes to the GSL. I've closed that out as it is probably painful to revive at this point with all the file renames and so on that have occurred. Would you be interested in providing a new one?

neilmacintosh:
[...]

PR #659, opened 2018-04-13
not_null constructor is now explicit

ericLemanissier:
solves #395

### DIFF
--- a/include/gsl/pointers
+++ b/include/gsl/pointers
@@ -72,7 +72,7 @@ public:
     static_assert(std::is_assignable<T&, std::nullptr_t>::value, "T cannot be assigned nullptr.");
 
     template <typename U, typename = std::enable_if_t<std::is_convertible<U, T>::value>>
-    constexpr not_null(U&& u) : ptr_(std::forward<U>(u))
+    constexpr explicit not_null(U&& u) : ptr_(std::forward<U>(u))
     {
         Expects(ptr_ != nullptr);
     }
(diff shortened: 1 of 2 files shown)


beinhaerter:
Just curious: what's the reason to change the initialization in the tests and why have different initialization ("{}" vs. "()") in the same test?

ericLemanissier:
it was just a way to check they all work. I can harmonize them if it is a blocking point

beinhaerter:
OK, thanks, that was my guess.

xaxxon:
I question the usefulness and desirability of this change.

xaxxon:
an alternative would be to make a new type explicit_not_null or something if this behavior is desired.

Amaroker:
> I question the usefulness and desirability of this change as stated in the associated issue - especially as a breaking change.

Changing a parameter to not_null can be a breaking change, regardless if the constructor is explicit or not.

However, the explicit constructor is very useful because we don't want such changes to silently (until run-time) break something.

The sooner we have the explicit constructor, the better.

The core guidelines say:

- Prefer compile-time checking to run-time checking.

- Single-argument constructors should be declared explicit. 

- Do not introduce implicit conversions (through conversion operators or non-explicit constructors) just to gain a minor convenience.

xaxxon:
This breaks code which already uses notnull.

ericLemanissier:
@neilmacintosh What do you think about this pull request ?

neilmacintosh:
@ericLemanissier I personally like this change, for the reasons observed on the issue threads: `not_null` will frequently be used to make a function have a strong contract about null values, and it seems sensible to ask callers to explicitly acknowledge that contract.

I'll leave final approval to @annagrin though.

annagrin:
@neilmacintosh Agreed. We do want to follow the Guidelines in GSL:)

PR #664, opened 2018-04-17
[span] Tell MSVC how to avoid range-checking 

CaseyCarter:
and cleanup `span_iterator` non-member operators.

### DIFF
--- a/include/gsl/span
+++ b/include/gsl/span
@@ -191,6 +191,11 @@ namespace details
             return ret += n;
         }
 
+        friend constexpr span_iterator operator+(difference_type n, span_iterator const& rhs)
+        {
+            return rhs + n;
+        }
+
         constexpr span_iterator& operator+=(difference_type n)
         {
             Expects((index_ + n) >= 0 && (index_ + n) <= span_->size());
@@ -257,22 +262,6 @@ namespace details
         const Span* span_ = nullptr;
         std::ptrdiff_t index_ = 0;
     };
-
-    template <class Span, bool IsConst>
-    constexpr span_iterator<Span, IsConst>
-    operator+(typename span_iterator<Span, IsConst>::difference_type n,
-              span_iterator<Span, IsConst> rhs)
-    {
-        return rhs + n;
-    }
-
-    template <class Span, bool IsConst>
-    constexpr span_iterator<Span, IsConst>
-    operator-(typename span_iterator<Span, IsConst>::difference_type n,
-              span_iterator<Span, IsConst> rhs)
-    {
-        return rhs - n;
-    }
 
     template <std::ptrdiff_t Ext>
     class extent_type
@@ -418,7 +407,7 @@ public:
 
     ~span() noexcept = default;
     constexpr span& operator=(const span& other) noexcept = default;
-    
+
     // [span.sub], span subviews
     template <std::ptrdiff_t Count>
     constexpr span<element_type, Count> first() const
@@ -496,7 +485,7 @@ public:
 private:
 
     // Needed to remove unnecessary null check in subspans
-    struct KnownNotNull 
+    struct KnownNotNull
     {
         pointer p;
     };
@@ -508,7 +497,7 @@ private:
     class storage_type : public ExtentType
     {
     public:
-        // KnownNotNull parameter is needed to remove unnecessary null check 
+        // KnownNotNull parameter is needed to remove unnecessary null check
         // in subspans and constructors from arrays
         template <class OtherExtentType>
         constexpr storage_type(KnownNotNull data, OtherExtentType ext) : ExtentType(ext), data_(data.p)
--- a/include/gsl/span
+++ b/include/gsl/span
@@ -482,6 +482,12 @@ public:
     constexpr const_reverse_iterator crbegin() const noexcept { return const_reverse_iterator{cend()}; }
     constexpr const_reverse_iterator crend() const noexcept { return const_reverse_iterator{cbegin()}; }
 
+#ifdef _MSC_VER
+    // Tell MSVC how to unwrap spans in range-based-for
+    constexpr pointer _Unchecked_begin() const noexcept { return data(); }
+    constexpr pointer _Unchecked_end() const noexcept { return data() + size(); }
+#endif // _MSC_VER
+
 private:
 
     // Needed to remove unnecessary null check in subspans


neilmacintosh:
Thanks @CaseyCarter. Can you clarify exactly what this fixes for people who might not be super-familiar with MSVC? Does this allow `span` to work better with MSVC standard library (esp. in its "checked" mode) or is this for better compiler optimization?

CaseyCarter:
> Can you clarify exactly what this fixes for people who might not be super-familiar with MSVC? Does this allow `span` to work better with MSVC standard library (esp. in its "checked" mode) or is this for better compiler optimization?

* `operator-(difference_type, span_iterator)` - unlike `operator-(span_iterator, difference_type)` - was completely bogus and not an iterator requirement. I've eliminated it with prejudice.

* Changing `operator+(difference_type, span_iterator)` from a namespace-scope non-member function template to a "hidden friend" non-member function is simply a tiny throughput improvement in that there are fewer candidates for overload resolution of operator `+` in namespace `GSL`.

* `_Unchecked`/`_Rechecked` implement an undocumented protocol (please eat this message after reading so it doesn't fall into the wrong hands) for unwrapping checked iterators into unchecked iterators and vice versa. The MSFT STL uses them internally with our "debug iterators" so that algorithms can operate on "unchecked" iterators after range checking. This can be faster both by avoiding checking that the optimizer fails to eliminate, and by engaging some pointer optimizations for `span` iterators, e.g., turning `std::copy` into `memcpy`.

* `_Unchecked_begin`/`_Unchecked_end` are the same thing for range-based-for loops (again undocumented and super secret). They allow `for (auto&& x : span{/*...*/})` to avoid iterator range-checking, which I suspect has much less practical effect when optimizing than for debug builds.

neilmacintosh:
Thanks @CaseyCarter! That helps people understand exactly what's going on. 
Would this PR allow us to remove the use of ` _SCL_SECURE_NO_WARNINGS` in CMakeLists.txt? Would doing so mean that the Debug build of unit tests with MSVC could validate this change (or does it already do that...I don't have a Windows machine lurking around anymore to check ;-)).

CaseyCarter:
> That helps people understand exactly what's going on.

Sorry for the complete lack of explanation in the OP. @annagrin knows what's going on here, but I needed a reminder that the rest of the world might like to know as well ;)

> Would this PR allow us to remove the use of `_SCL_SECURE_NO_WARNINGS` in `CMakeLists.txt`? 

As a MSFT employee, I can neither confirm nor deny my extreme hatred for the "deprecation" warnings that `_SCL_SECURE_NO_WARNINGS` suppresses. I don't think this PR improves the situation at all. Passing checked span iterators instead of unchecked pointers to `std` algorithms would fix this problem for `span`, FWIW, at the cost of conditional code or poorer code quality on not-MSVC.

> Would doing so mean that the Debug build of unit tests with MSVC could validate this change (or does it already do that...I don't have a Windows machine lurking around anymore to check ;-)).

Yes, the AppVeyor tests are validating the change.

CaseyCarter:
Replacing the body of `gsl::copy` with:

```c++
 template <class SrcElementType, std::ptrdiff_t SrcExtent, class DestElementType,
           std::ptrdiff_t DestExtent>
 void copy(span<SrcElementType, SrcExtent> src, span<DestElementType, DestExtent> dest)
 {
     static_assert(std::is_assignable<DestElementType&, SrcElementType&>::value,
                   "Elements of source span can not be assigned to elements of destination span");
     static_assert(DestExtent == dynamic_extent || SrcExtent <= DestExtent,
                   "Source range is longer than target range");
 
     Expects(dest.size() >= src.size());
 #if _MSC_VER
     std::copy_n(src.begin(), src.size(), dest.begin());
 #else
     std::copy_n(src.data(), src.size(), dest.data());
 #endif
 }
```
eliminates all `_SCL_SECURE_NO_WARNINGS` warnings except for those in `multi_span_tests.cpp`.

neilmacintosh:
Thanks for reminding me of those "deprecation" warnings that happen without `_SCL_SECURE_NO_WARNINGS `. Bleah.

It seems like keeping `_SCL_SECURE_NO_WARNINGS` may not be the worst thing in the world.

This all makes perfect sense to me now, thanks for satisfying my curiosity! I'll leave this in the very capable hands of @annagrin from here.

include/gsl/span
```diff
@@ -253,27 +258,27 @@ namespace details
             return !(rhs > lhs);
         }
 
+#ifdef _MSC_VER
+        // Tell Microsoft standard library how to unwrap span iterators
```
CaseyCarter: @annagrin, given our discussion of the impedance mismatch - `span` always wants range checking, STLSTL only range checks when compiling `_DEBUG` - would you prefer that I make this part of the change conditional on `_DEBUG`, or just rip it out for now?

CaseyCarter: To be clear: the rest of this change doesn't interfere with `span`'s range-checking behavior, assuming that you trust the compiler no to miscompile range-based for loops.

annagrin:
@CaseyCarter Let's wait until the missing check issue in STL is resolved. We can merge this PR in then. Looking forward to it, the code gen looks awesome!

annagrin:
@CaseyCarter If you would like to submit just the range-for change separately, that would be great.

include/gsl/span
annagrin: @CaseyCarter Let's wait with the range checking until the impedance mismatch is resolved. If you would like to submit just the range-based loops change, that would be great! Thanks, the code gen looks awesome with the change.

include/gsl/span
```diff
@@ -191,6 +191,11 @@ namespace details
             return ret += n;
         }
 
+        friend constexpr span_iterator operator+(difference_type n, span_iterator const& rhs)
+        {
+            return rhs + n;
+        }
+
         constexpr span_iterator& operator+=(difference_type n)
```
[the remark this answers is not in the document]
CaseyCarter: `<iterator> + <difference>` and `<difference> + <iterator>` are equivalent expressions since `+` is commutative. Since `-` is *not* commutative, `<difference> - <iterator>` has no defined meaning and is therefore not required.

CaseyCarter:
I've yanked the iterator unwrapping code, this should be fine to land once CI completes.

annagrin:
Thanks @CaseyCarter! I am merging this in, but will keep this issue open for now until we have an STL update allowing to check before unwrap.

PR #667, opened 2018-04-20
Make gsl::details::throw_exception not copy exception object

StephanDollberg:
This fixes the build with clang 7 which introduces a new warning
`-Wreturn-std-move` which warns about needless copies when a move
operation is available (see https://reviews.llvm.org/rL329914).

We take the exception object by uref so we should throw the forwarded
version.


annagrin:
thanks @StephanDollberg!

PR #668, opened 2018-04-21
Add __may_alias__ attribute to gsl::byte

### LINKED ISSUE #663: Your byte implementation is unsafe under optimisation on GCC and clang
The C++ 20 standard defines `std::byte*` to have the same may-alias properties as `char*`.

Your byte implementation therefore needs to be marked up with `__attribute__((__may_alias__))` on GCC and clang compilers to tell the compiler that it must not assume that `byte*` cannot alias a pointer to a different type.

StephanDollberg:
C++20 defines `std::byte*` to have the same aliasing properties as
`char*`. Hence, we mark it with the `__may_alias__` attribute under gcc
and clang.

Fixes #663 

(Actually couldn't find whether this requirement is already in C++17 or not but gcc and clang already implement this in the compiler, trusting Niall on this part)

### DIFF
--- a/include/gsl/gsl_byte
+++ b/include/gsl/gsl_byte
@@ -57,6 +57,13 @@
 
 #endif           // _MSC_VER
 
+// Use __may_alias__ attribute on gcc and clang
+#if defined __clang__ || (__GNUC__ > 5)
+#define byte_may_alias __attribute__((__may_alias__))
+#else // defined __clang__ || defined __GNUC__
+#define byte_may_alias
+#endif // defined __clang__ || defined __GNUC__
+
 namespace gsl
 {
 #if GSL_USE_STD_BYTE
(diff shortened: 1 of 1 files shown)


annagrin:
@StephanDollberg Are there any tests possible to show correct code gen?

martinmoene:
@StephanDollberg 
>(Actually couldn't find whether this requirement is already in C++17 or not but gcc and clang already implement this in the compiler, trusting Niall on this part)

See [this remark at issue #663 ](https://github.com/Microsoft/GSL/issues/663#issuecomment-381893893)

Related: PR #390

StephanDollberg:
Ah great, somehow missed that comment. Will amend the commit message.

StephanDollberg:
@annagrin Checked with the tests. So the other day I was compiling with clang which apparently doesn't warn. GCC requires `-fstrict-aliasing` or `-O2` (which turns strict aliasing on) to warn about the aliasing test. 

Given that gcc 5 doesn't support the attribute I'd recommend against turning those flags on for now.

Rebased as well.

annagrin:
Thanks @StephanDollberg!

PR #675, opened 2018-05-02
not_null no longer requires a copyable pointer type

### LINKED ISSUE #89: not_null<unique_ptr<>> doesn't work

### LINKED ISSUE #550: not_null<shared_ptr> overhead

### LINKED ISSUE #651: support for unique_ptr and significant speedups for shared_ptr vs returning const& to internal ptr_ field
If get(), operator->, and operator T(), are changed from returning  a copy to returning a T const &, not_null gets support for unique_ptr (and any other move-only pointer-like object) as well as drastic speedups with pointer-like types with expensive copy constructors, like shared_ptr.   

However, this means that a not-aware but potentially not malicious user could const cast the const away from the reference and change the internal value.   With the recent discussion to remove the per-access null check, this would allow nulls to be returned from not_null without jumping through massive hoops.

The speed benefit to shared_ptr is drastic.   In the microbenchmark code (listed below), having a not_null<shared_ptr<T>> doubles the time to access data through it in a "getter"-type function, from 20 microseconds to 40 microseconds even when fully inline-able by the optimizer, since the reference count changes cannot be optimized out.   This type of cost is associated with any non-trivial copy constructor, not just shared_ptr.   With the proposed change, there is no additional cost for shared_ptr since no copy is made.

As for unique_ptr, I find the use case somewhat compelling for making sure you don't forget to allocate a pointer to it, when you don't intend for it to ever be empty.  Of potentially limited use, it also stops it from being moved out of, since only a const version of the unique_ptr can ever be obtained.   Without the concept of a "destructive move" in c++, there's no way to say that it's ok for the contained pointer to become null because it will never be accessed again.   However, this doesn't mean there is no use case for this, just that it may be smaller than one might anticipate.

There also may be ways to achieve these goals with a more complex implementation of not_null involving much more TMP and making all the comparison operators friend functions.   So I guess I'm curious what other people's thoughts are on speed versus implementation simplicity versus more heavily guaranteed safety against well-intentioned but clearly incorrect use of not_null.   

My vote, as evidenced by the hacked up version of not_null I actually use is to return a const & (and I intentionally violate the contract sometimes to move out of it).   

```
#include "GSL/include/gsl/pointers"
#include <benchmark/benchmark.h>
#include <memory>

using namespace std;

struct SomeStruct {
    int i;
};

int f(gsl::not_null<shared_ptr<SomeStruct>> nnsp) {
    return nnsp->i;
}

int f2(shared_ptr<SomeStruct> sp) {
    return sp->i;
}


static void null_test(benchmark::State& state) {
    auto nnsp = make_shared<SomeStruct>();
  for (auto _ : state) {
      for (int i = 0; i < 1000; i++)
      benchmark::DoNotOptimize(f(nnsp));
  }
}
// Register the function as a benchmark
BENCHMARK(null_test);


static void no_null_test(benchmark::State& state) {

    auto sp = make_shared<SomeStruct>();
    for (auto _ : state) {
	for (int i = 0; i < 1000; i++)
	benchmark::DoNotOptimize(f2(sp));
    }
}
BENCHMARK(no_null_test);

BENCHMARK_MAIN();

```

results:

```
2018-03-31 00:17:15
Run on (4 X 1700 MHz CPU s)
CPU Caches:
  L1 Data 32K (x2)
  L1 Instruction 32K (x2)
  L2 Unified 262K (x2)
  L3 Unified 3145K (x1)
----------------------------------------------------
Benchmark             Time           CPU Iterations
----------------------------------------------------
null_test         20129 ns      20055 ns      35072
no_null_test      19844 ns      19775 ns      34906

```

ericLemanissier:
allows support of supporting std::unique_ptr
based on @xaxxon's work

fixes #89
fixes #550
fixes #651

From the Core Guidelines:

http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#note-61

> Note not_null is not just for built-in pointers. It works for unique_ptr, shared_ptr, and other pointer-like types.

### DIFF
--- a/tests/notnull_tests.cpp
+++ b/tests/notnull_tests.cpp
@@ -440,7 +440,7 @@ TEST_CASE("TestNotNullUniquePtrComparison") {
         CHECK(p->i == NotNull1(std::make_unique<UniquePointerTestStruct>())->i);
     }
 }
-static_assert(!std::is_assignable<not_null<std::unique_ptr<int>>, std::unique_ptr<int>>::value, "not_null<std::unique_ptr> must be non assignable from std:unique_ptr");
+static_assert(!std::is_assignable<not_null<std::unique_ptr<int>>, std::unique_ptr<int>&>::value, "not_null<std::unique_ptr> must be non assignable from std:unique_ptr lvalues");
 static_assert(!std::is_copy_constructible<not_null<std::unique_ptr<int>>>::value, "not_null<std::unique_ptr> must be non copy constructible");
 static_assert(!std::is_copy_assignable<not_null<std::unique_ptr<int>>>::value, "not_null<std::unique_ptr> must be non copy assignable");
 static_assert(std::is_nothrow_move_constructible<not_null<std::unique_ptr<int>>>::value, "not_null<std::unique_ptr> must be no-throw move constructible");
@@ -492,7 +492,7 @@ TEST_CASE("TestNotNullSharedPtrValueComparison") {
         CHECK(p->i == NotNull1(std::make_shared<UniquePointerTestStruct>())->i);
     }
 }
-static_assert(!std::is_assignable<not_null<std::shared_ptr<int>>, std::shared_ptr<int>>::value, "not_null<std::shared_ptr> must be non assignable from std::shared_ptr");
+static_assert(std::is_assignable<not_null<std::shared_ptr<int>>, std::shared_ptr<int>>::value, "not_null<std::shared_ptr> must be assignable from std::shared_ptr");
 static_assert(std::is_copy_constructible<not_null<std::shared_ptr<int>>>::value, "not_null<std::shared_ptr> must be copy constructible");
 static_assert(std::is_copy_assignable<not_null<std::shared_ptr<int>>>::value, "not_null<std::shared_ptr> must be copy assignable");
 static_assert(std::is_nothrow_move_constructible<not_null<std::shared_ptr<int>>>::value, "not_null<std::shared_ptr> must be no-throw move constructible");
--- a/tests/notnull_tests.cpp
+++ b/tests/notnull_tests.cpp
@@ -332,7 +332,7 @@ TEST_CASE("TestNotNullCustomPtrComparison")
 static_assert(std::is_nothrow_move_constructible<not_null<void *>>::value, "not_null must be no-throw move constructible");
 
 struct UniquePointerTestStruct {
-    int i = 0;
+    int i = 42;
 };
 
 TEST_CASE("TestNotNullUniquePtrComparison") {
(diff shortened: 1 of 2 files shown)


annagrin:
@ericLemanissier thanks! That looks clean and separate from the get() function change. How does 
`not_null<unique_ptr<T>>` behave when copied, moved, assigned, reset? If would be great to add tests to ensure right things compile and wrong do not (you can #define the non-compiling tests away). Also, adding tests to get the value out of the unique_ptr would be useful as well.

annagrin:
@ericLemanissier tried it out, and it looks like something is not yet working as expected - the following causes access violation. Looks like unique_ptr containting nullptr is not equal to nullptr, so Ensure() in get() does not catch it. Not sure how to fix it - any thoughts? 
```
int main()
{
    not_null<std::unique_ptr<int*>> y{ std::make_unique<int*>() };
    *y = nullptr;
    return **y; // AV here
}
```

tests/notnull_tests.cpp
```diff
@@ -329,3 +330,57 @@ TEST_CASE("TestNotNullCustomPtrComparison")
...
+    int i = 0;
+};
+
+TEST_CASE("TestNotNullUniquePtrComparison") {
+
+    {
+        using NotNull1 = not_null<std::unique_ptr<int>>;
+
+        // values are the same
+        CHECK((*NotNull1(std::make_unique<int>(42)) == *NotNull1(std::make_unique<int>(42))));
```
annagrin: can we add a tests to see that the value is actually 42?

tests/notnull_tests.cpp
```diff
@@ -329,3 +330,57 @@ TEST_CASE("TestNotNullCustomPtrComparison")
...
+    }
+    {
+        using NotNull1 = not_null<std::unique_ptr<UniquePointerTestStruct>>;
+
+        // values are the same
+        CHECK((NotNull1(std::make_unique<UniquePointerTestStruct>())->i == NotNull1(std::make_unique<UniquePointerTestStruct>())->i));
```
annagrin: Would be nice also to check the value of field i

tests/notnull_tests.cpp
annagrin: I see it below, please ignore this comment

tests/notnull_tests.cpp
```diff
@@ -329,3 +330,57 @@ TEST_CASE("TestNotNullCustomPtrComparison")
...
+        NotNull1 src(std::make_unique<int>(42));
+        NotNull1 dst(std::move(src));
+        #ifdef GSL_THROW_ON_CONTRACT_VIOLATION
+        CHECK_THROWS_AS(!src.get(), fail_fast);
+        #endif
+        CHECK(dst.get());
+        CHECK(*dst.get() == 42);
+        src = std::move(dst);
+        #ifdef GSL_THROW_ON_CONTRACT_VIOLATION
+        CHECK_THROWS_AS(!dst.get(), fail_fast);
```
annagrin: can we show that reset() and setting to nullptr do not compile?

tests/notnull_tests.cpp
```diff
@@ -329,3 +330,57 @@ TEST_CASE("TestNotNullCustomPtrComparison")
...
+        CHECK(src.get());
+        CHECK(*src.get() == 42);
+    }
+    {
+        using NotNull1 = not_null<std::unique_ptr<UniquePointerTestStruct>>;
+
+        // values are the same
+        CHECK((NotNull1(std::make_unique<UniquePointerTestStruct>())->i == NotNull1(std::make_unique<UniquePointerTestStruct>())->i));
+    }
+}
```
annagrin: would be great to see how does this change behaves with other smart pointers

annagrin:
@ericLemanissier please ignore deleted comment, I misused unique_ptr.
Would love to see more tests for getting fields out of `not_null<unique_ptr<CustomPointer>>` , tests showing that reset() does not compile, and tests for other smart pointers.

BTW, looks like your other change - removing Ensures() from get() - is incompatible with this one, as it will remove the null checking on moved `not_null<unique_ptr<X>>` 

```
int main()
{
    not_null<std::unique_ptr<int>> y{ std::make_unique<int>() };
    auto z = std::move(y);
    return *y; // AV here
}
```

ericLemanissier:
@annagrin Are these the tests you had in mind ?
Regarding the interaction with #674, it is limited to changing the blocks
```c++
#ifdef GSL_THROW_ON_CONTRACT_VIOLATION
    CHECK_THROWS_AS(!src.get(), fail_fast);
#endif
```
to
```c++
    CHECK(!src.get());
```

ericLemanissier:
@annagrin what do you think of this PR ?

annagrin:
@ericLemanissier I would like to clear some points with this PR and #674.

1. A conceptual problem - before this change you could not have a variable of `not_null<T>` containing a null pointer (and that would allow for #674 to be possible). Now `notnull<unique_ptr<T>>` might contain a null pointer if unique pointer was reset. We could argue that the relaxed semantics makes more sense, but that should be a conscious choice, and see that argument actually happening...  I would love to gather more opinions on that.

2. Corollary of 1 - problem with this change being incompatible with #674 - together they remove termination/exception in the example I provided before, and allow an actual null dereference, which is going against the purpose of not_null type.

ericLemanissier:
A mitigation of the problem is that get returns a reference to const, so the unique_ptr cannot be reset, unless explicitely casting const away, which is not a good pratice.

ericLemanissier:
The moved from not_null contains an empty unique_ptr, so it is actually null. This is not a problem if the next operation is destruction, or assignment, which is a usual behaviour for moved from objects, and not surprising because it has the same behaviour as the class template argument type.
One might argue that all other member functions(get and all operators calling get) documentation should be explicit about the requirement for this to be in a non moved from state. Any way, such missuses are already caught by the ensures clause. By the way, shouldn't it be an expects clause instead?

annagrin:
@gdr-at-ms I could use your opinion on the semantic change - see my "1. Conceptual problem" above,

@ericLemanissier If we accept this change, we would have access violations after move if #674 is checked in as well. So my inclination is to take this change but not the #674.

KVic:
@gdr-at-ms, @annagrin, what  happens?
```
#include <gsl/pointers>

class SomeClass {};

int main()
{
    using namespace std;

    gsl::not_null<unique_ptr<SomeClass>> some_unique{make_unique<SomeClass>()};

    SomeClass some_class = move(*some_unique);

    return 0;
}
```

So I agree with @ericLemanissier that "This is not a problem if the next operation is destruction, or assignment, which is a usual behaviour for moved from objects". And there is no need in the Ensures() check in the get() in the Release build.

annagrin:
Thanks @ericLemanissier , I am merging this change, but would need to reject the #674 - we cannot remove the null check now, because moved-from not-null smart pointer can now contain a null pointer.

JordanMaples:
Maintainer's call: I've opened issue isocpp/CppCoreGuidelines#1499 to continue the discussion.

PR #676, opened 2018-05-03
fix requirement on not_null template

ericLemanissier:
it should be comparable to nullptr, it does not have to be assignable

http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#gslview-views says

> not_null<T> // T is usually a pointer type (e.g., not_null<int*> and not_null<owner<Foo*>>) that may not be nullptr. T can be any type for which ==nullptr is meaningful

### DIFF
--- a/include/gsl/pointers
+++ b/include/gsl/pointers
@@ -69,7 +69,7 @@ template <class T>
 class not_null
 {
 public:
-    static_assert(std::is_assignable<T&, std::nullptr_t>::value, "T cannot be assigned nullptr.");
+    static_assert(std::is_convertible<decltype(std::declval<T>() != nullptr), bool>::value, "T cannot be compared to nullptr.");
 
     template <typename U, typename = std::enable_if_t<std::is_convertible<U, T>::value>>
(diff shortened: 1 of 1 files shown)


annagrin:
@ericLemanissier, I am not sure what differences this introduces - could you add tests for anything new that compiles and anything that does not compile anymore?

ericLemanissier:
The current static_assert checks that T can be assigned a nullptr, but nowhere in the implementation of not_null is T assigned nullptr. This forbids the use of not_null with types without assignment operator like this : https://godbolt.org/g/WyBWhd
Also, being assignable a nullptr does not imply that T can be compared to nullptr. Of course, the compilation still fails, but the error message is not as clear as a static assert : https://godbolt.org/g/C7Mv8D

ericLemanissier:
@JordanMaples is there some interest in this PR ?

JordanMaples:
Maintainers' call: Thanks for all of your patience. The change looks good and we'll accept it. Thanks!

Issue #685, opened 2018-05-21
Constructing gsl::span with empty std::array may cause crash


dave-hill:
The std::array constructor for span sets the storage equal to `&arr[0]`. This triggers runtime faults in VC++ and clang when called with an empty array (when unoptimized).

This can easily be fixed by using `arr.data()` instead of `&arr[0]`.

Suggested fix in PR #686.

PR #686, opened 2018-05-21
Span can be constructed from empty std::array safely

dave-hill:
- Fixes runtime issues when constructing from an empty std::array (Issue #685)

### DIFF
--- a/include/gsl/span
+++ b/include/gsl/span
@@ -409,13 +409,21 @@ public:
 
     template <std::size_t N, class ArrayElementType = std::remove_const_t<element_type>>
     constexpr span(std::array<ArrayElementType, N>& arr) noexcept
+#if (defined(__cplusplus) && (__cplusplus >= 201703L))
+        : storage_(std::data(arr), details::extent_type<N>())
+#else
         : storage_(arr.data(), details::extent_type<N>())
+#endif
     {
     }
 
     template <std::size_t N>
     constexpr span(const std::array<std::remove_const_t<element_type>, N>& arr) noexcept
+#if (defined(__cplusplus) && (__cplusplus >= 201703L))
+        : storage_(std::data(arr), details::extent_type<N>())
+#else
         : storage_(arr.data(), details::extent_type<N>())
+#endif
     {
     }
 
--- a/include/gsl/span
+++ b/include/gsl/span
@@ -394,17 +394,27 @@ public:
         : storage_(KnownNotNull{std::addressof(arr[0])}, details::extent_type<N>())
     {}
 
-    template <std::size_t N, class ArrayElementType = std::remove_const_t<element_type>>
-    // GSL_SUPPRESS(bounds.4) // NO-FORMAT: attribute // TODO: parser bug
-    constexpr span(std::array<ArrayElementType, N>& arr) noexcept
-        : storage_(arr.data(), details::extent_type<N>())
-    {}
+    template <std::size_t N, class = std::enable_if_t<(N > 0)>>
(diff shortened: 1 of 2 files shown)


include/gsl/span
```diff
@@ -361,13 +361,13 @@ public:
 
     template <std::size_t N, class ArrayElementType = std::remove_const_t<element_type>>
     constexpr span(std::array<ArrayElementType, N>& arr) noexcept
-        : storage_(&arr[0], details::extent_type<N>())
+        : storage_(arr.data(), details::extent_type<N>())
```
neilmacintosh: Would std::data(arr) work as well? From memory, that is closer to the WD description of what happens.

dave-hill: std::data would work too, but it was introduced in C++17. GSL generally assumes C++14 right?

include/gsl/span
```diff
@@ -409,13 +409,21 @@ public:
 
     template <std::size_t N, class ArrayElementType = std::remove_const_t<element_type>>
     constexpr span(std::array<ArrayElementType, N>& arr) noexcept
-        : storage_(&arr[0], details::extent_type<N>())
+#if (defined(__cplusplus) && (__cplusplus >= 201703L))
```
dave-hill: Apologies @annagrin, I lost your comment whilst addressing it by rebasing instead of merging (I'm used to VSTS).

@neilmacintosh feedback from 5/21
> Would std::data(arr) work as well? From memory, that is closer to the WD description of what happens.

@annagrin feedback from 6/15
> I just added tests for c++17. We can have conditional compilation dependent on __cplusplus value (with the idea that we eventually progress out of conditional compilation)

annagrin:
@dave-hill Still crashing intermittently due to `array.data()` sometimes implemented as `&arr[0]`.  Thanks to @neilmacintosh, he figured it out - http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/n4727.pdf says that calling `.data()` on empty arrays is unspecified, so we cannot rely on libraries here even if they work. We need a different approach - maybe use `nullptr` instead of `arr.data()` when `N==0`?

> [array.zero]
> 26.3.7.5 Zero sized arrays [array.zero] 1 array shall provide support for the special case N == 0. 2 In the case that N == 0, begin() == end() == unique value. The return value of data() is unspeciﬁed. 3 The eﬀect of calling front() or back() for a zero-sized array is undeﬁned. 4 Member function swap() shall have a non-throwing exception speciﬁcation.

annagrin:
@dave-hill I recommend specializing for arrays of size 0 - something like the following

```
template <std::size_t N, class = std::enable_if_t<(N > 0)>>
    constexpr span(std::array<std::remove_const_t<element_type>, N>& arr) noexcept
        : storage_(KnownNotNull{arr.data()}, details::extent_type<N>())
    {
    }
 
    constexpr span(std::array<std::remove_const_t<element_type>, 0>& arr) noexcept
        : storage_(static_cast<pointer>(nullptr), details::extent_type<0>())
    {
    }
 
    template <std::size_t N, class = std::enable_if_t<(N > 0)>>
    constexpr span(const std::array<std::remove_const_t<element_type>, N>& arr) noexcept
        : storage_(KnownNotNull{arr.data()}, details::extent_type<N>())
    {
    }
 
    constexpr span(const std::array<std::remove_const_t<element_type>, 0>& arr) noexcept
        : storage_(static_cast<pointer>(nullptr), details::extent_type<0>())
    {
    }

``` 
 
Instead of the couple constructors from std::array.

annagrin:
Hey @dave-hill, would you like to continue with this change?

annagrin:
@dave-hill many thanks! Merging in.

PR #689, opened 2018-05-31
Added template deduction guides for not_null

annagrin:
This allows compilers with c++17 support to infer template
instantiation types when calling not_null constructor:

    int foo(not_null<const int*> x);

    int main()
    {
        int t = 0;
        not_null x{ &t };
        return foo(not_null{ &t });
    }


include/gsl/pointers
```diff
@@ -113,6 +117,11 @@ private:
     T ptr_;
 };
 
+#if defined GSL_DEDUCTION_GUIDES_SUPPORTED
```
neilmacintosh: Trivial nit: parens around the definition being tested?

include/gsl/pointers
```diff
@@ -113,6 +117,11 @@ private:
...
+template <typename T>
```
neilmacintosh: I like that this elegant little guide that makes deduction work nicely. However, does it point to a failing in our existing set of constructors for `not_null`? Do we just need to have a constructor from `T&&` and then the guide would be unnecessary?

CaseyCarter: > Do we just need to have a constructor from `T&&` and then the guide would be unnecessary?

If there were a `T&&` constructor, the implicit deduction guides would deduce from *rvalue* arguments  - but not *lvalue* arguments. A plain `T` constructor would be a better choice.

neilmacintosh: Makes sense. I was just parrotting the `U&&` one without thinking, `T` makes much more sense. And it shouldn't compete with the `U&&` constructor?

annagrin: Maybe. Will check!

CaseyCarter: > And it shouldn't compete with the `U&&` constructor?

It will compete, and it will win by virtue of being a non-template.

CMakeLists.txt
```diff
@@ -14,30 +14,65 @@ if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
...
 # when minimum version required is 3.8.0 remove if below
 # both branches do exactly the same thing
-if (CMAKE_VERSION VERSION_LESS 3.7.9)
+#if (CMAKE_VERSION VERSION_LESS 3.7.9)
     if (NOT MSVC)
-        include(CheckCXXCompilerFlag)
-        CHECK_CXX_COMPILER_FLAG("-std=c++14" COMPILER_SUPPORTS_CXX14)
-        if(COMPILER_SUPPORTS_CXX14)
-            target_compile_options(GSL INTERFACE "-std=c++14")
+        if(COMPILER_SUPPORTS_CXX17)
```
MikeGitb: Just because the compiler supports c++17 doesn't necessarily mean I want to compile my project in that mode. This should be left to the use to decide.

In fact, this library should imho set no flags at all (at most it should `target_compile_features(GSL INTERFACE cxx_std_14)`

The machinery here seems more suited for the tests, but even then, one might want to be able to run multiple tests with the same compiler, but in different language modes.

johelegp: @MikeGitb suggests the right way to do it in modern CMake. CMake 3.8.0 added support for `target_compile_features(GSL INTERFACE cxx_std_`*`)`, which a library should use to suggest the oldest standard they support. From all its calls, CMake calculates which flags to add. Before 3.8.0, I'd guess the correct thing to do is set no `-std=` flag at all and let the top project deal with that.

annagrin: Thanks @MikeGitb, those settings should not affect your project, we only compile tests here. That said, definitely agree with changing the test matrix to support runs with c++14 and c++17 separately, so I am going to revert my changes and explore that other option.
@johelegp consider this an experiment with setting the most c++ standard the compiler supports, for testing the newer features.

CMakeLists.txt
```diff
@@ -14,21 +14,36 @@ if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
...
+endif()
+
 # when minimum version required is 3.8.0 remove if below
 # both branches do exactly the same thing
 if (CMAKE_VERSION VERSION_LESS 3.7.9)
     if (NOT MSVC)
-        include(CheckCXXCompilerFlag)
-        CHECK_CXX_COMPILER_FLAG("-std=c++14" COMPILER_SUPPORTS_CXX14)
         if(COMPILER_SUPPORTS_CXX14)
             target_compile_options(GSL INTERFACE "-std=c++14")
```
johelegp: I think these branches, that only differ in `=`/`:`, can be merged by changing the string to the following generator expression: `$<$<NOT:$<BOOL:${MSVC}>>:"-std=c++14"> $<$<BOOL:${MSVC}>:"-std:c++14">`.

annagrin: @johelegp yes you are right... will keep as is for now just because I want to separate adding c++17 tests in the next PR and will need to change this code anyway. Will address your comment there.

CMakeLists.txt
MikeGitb: @annagrin: I admit I haven't tested it, but how does `target_compile_options(GSL INTERFACE "-std=c++17")` not affect my project if I directly or indirectly do `target_link_libraries(my_app GSL)` ?

CMakeLists.txt
MikeGitb: @johelegp: I'd argue that the branches are actually more readable than the equivalent generator expression

PR #690, opened 2018-06-04
another trial to fix gcc build (gcc 7)

menete:
this is a quite pragmatic solution - it simply disables the specific warning

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -160,17 +160,16 @@ matrix:
     - env: COMPILER=g++-6 BUILD_TYPE=Release
       addons: *gcc6
 
-    # Currently fails due to #525
-    ## GCC 7
-    #- env: COMPILER=g++-7 BUILD_TYPE=Debug
-    #  addons: &gcc7
-    #    apt:
-    #      packages: g++-7
-    #      sources:
-    #        - ubuntu-toolchain-r-test
+    # GCC 7
+    - env: COMPILER=g++-7 BUILD_TYPE=Debug
+      addons: &gcc7
+        apt:
+          packages: g++-7
+          sources:
+            - ubuntu-toolchain-r-test
 
-    #- env: COMPILER=g++-7 BUILD_TYPE=Release
-    #  addons: *gcc7
+    - env: COMPILER=g++-7 BUILD_TYPE=Release
+      addons: *gcc7
 
 
 install:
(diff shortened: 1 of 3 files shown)


dcleblanc:
I think narrow should fix that without quenching the warning.

gdr-at-ms:
This looks good to me. Neil, Anna?

johelegp:
Should we not wait and see how [Sizes Should Only `span` Unsigned ](https://wg21.link/P1089) is received by the committee? It's acceptance would make this unnecessary.

dcleblanc:
@johelegp - I'm in agreement with the analysis in the link you posted. I am especially sensitive to signed-unsigned warnings, because these are often associated with unsafe code and security vulnerabilities. We should never be suppressing these warnings in a library. I don't like the general approach in any case, because a ptrdiff_t is limited to 2^31-1 on 32-bit systems, but some 32-bit systems can actually allocate 3gb, which this cannot hold. If you want to have a magic number to indicate a special case, then make it std::numeric_limits (lt) size_t (gt)::max().

In the meantime, signed to unsigned conversions should be handled with appropriate checking for invalid values. I do note that there are other suppressions of this warning in the library, and those should be addressed as well. Some of it is because the Microsoft headers do not compile cleanly at -Wall, but we can suppress warnings within the headers without also affecting code within this library.

neilmacintosh:
Thanks @gdr-at-ms, I like this fix! @johelegp I don't think it's worth waiting for the paper you referenced to be accepted/rejected by the committee. The C++ Core Guidelines are the origin point of GSL and remain so, and they are clear in preferring signed to unsigned integers for cases such as  this (see ES.106 and ES.107).

neilmacintosh:
Oh and my apologies...thanks to @menete for the original fix! I'm glad @gdr-at-ms pinged us on it.

johelegp:
I agree that it's not worth the wait. I felt like pointing it out because it's been broken for so long, and yet this finally took traction when the committee is meeting and might accept the paper's direction.

dcleblanc:
Suppressing the warning isn't the best fix. Using a template specialization so that it is always correct code is better.

Sent from my Windows 10 phone

________________________________

neilmacintosh:
@dcleblanc I appreciate the feedback. For an open source project like this, we prefer to see PRs that actually offer a better fix. If you think you have one, please feel free to submit it for consideration.

dcleblanc:
Once I sort out the mechanics of creating a pull request, warning cleanup is high on my list. From my experience cleaning up Office warnings, signed unsigned mismatches are often associated with flaws.

On vacation until 18th, will address this upon return.

Sent from my Windows 10 phone

________________________________

PR #698, opened 2018-06-16
Make narrow constexpr

johelegp:
I hope Visual Studio 2015 works.

### DIFF
--- a/include/gsl/gsl_util
+++ b/include/gsl/gsl_util
@@ -37,6 +37,12 @@
 #endif            // _MSC_VER < 1910
 #endif            // _MSC_VER
 
+#if (defined(_MSC_VER) && _MSC_VER < 1910) || (!defined(__clang__) && defined(__GNUC__) && __GUNC__ < 6)
+#define GSL_CONSTEXPR_NARROW 0
+#else
+#define GSL_CONSTEXPR_NARROW 1
+#endif
+
 namespace gsl
 {
 //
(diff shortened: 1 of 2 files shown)


johelegp:
Doesn't work for Visual Studio 2015 nor GCC 5. When you drop Visual Studio 2015 support, consider also dropping support for GCC 5. I'll reopen when those two compilers are no longer supported.

annagrin:
@johelegp  If you want to ifdef around its ok - it will go away when the support for an old version goes away.

johelegp:
Thanks for the suggestion, I'll give it a go.

I just noticed the `#define constexpr /*constexpr*/` when `_MSC_VER < 1910`. If I guard the test with that, then Visual Studio 2015 tests should pass. However, GCC 5 will continue to fail. But this is the first `constexpr` test, so perhaps I shouldn't test and call it a day?

annagrin:
@johelegp I guess the test should be guarded with the same logic as defining constexpr away.

johelegp:
Thanks. I've done that, with the addition of detecting GCC 5 to not test when the macro `GSL_DONT_TEST_CONSTEXPR` is defined.

johelegp:
If this PR is desired, I could merge the definition for `GSL_DONT_TEST_CONSTEXPR` as
`#if defined(_MSC_VER) && _MSC_VER < 1910` for each block, or have one block like
`#if (defined(_MSC_VER) && _MSC_VER < 1910) || (!defined(__clang__) && defined(__GNUC__) && __GUNC__ < 6)`.

annagrin:
@johelegp should we ifdef in the library itself, not in tests, so older compilers could you narrow as well?

johelegp:
Right. Thanks for your support. Does the macro use looks right?

annagrin:
thanks @johelegp !

PR #642, opened 2018-03-16
Suppress warnings on VS

beinhaerter:
For this short code snippet
```
#include <gsl/gsl_util>
#include <vector>

void pr642()
{
	short buf[3];
	std::vector<short> vec(3);
	gsl::at(buf, 2) = gsl::narrow<short>(42);
	gsl::at(vec, 2) = gsl::narrow<short>(42);
}

```
VS 15.6.2 code analysis generates annoying warnings.
> ...\gsl\include\gsl\gsl_util(93) : warning C26472 : Don't use a static_cast for arithmetic conversions. Use brace initialization, gsl::narrow_cast or gsl::narow (type.1).
> ...\gsl\include\gsl\gsl_util(127) : warning C26472 : Don't use a static_cast for arithmetic conversions. Use brace initialization, gsl::narrow_cast or gsl::narow (type.1).
> ...\gsl\include\gsl\gsl_util(127) : warning C26482 : Only index into arrays using constant expressions(bounds.2).
> ...\gsl\include\gsl\gsl_util(127) : warning C26446 : Prefer to use gsl::at() instead of unchecked subscript operator (bounds.4).
> ...\gsl\include\gsl\gsl_util(114) : warning C26472 : Don't use a static_cast for arithmetic conversions. Use brace initialization, gsl::narrow_cast or gsl::narow (type.1).
> ...\gsl\include\gsl\gsl_util(135) : warning C26472 : Don't use a static_cast for arithmetic conversions. Use brace initialization, gsl::narrow_cast or gsl::narow (type.1).
> ...\gsl\include\gsl\gsl_util(135) : warning C26446 : Prefer to use gsl::at() instead of unchecked subscript operator (bounds.4).
> ...\gsl\include\gsl\gsl_assert(106) : warning C26440 : Function 'gsl::details::throw_exception<gsl::narrowing_error>' can be declared 'noexcept' (f.6).

gsl::narrow, gsl::narrow_cast and gsl::at are the safe variants suggested by CppCoreGuideline. It does not make sense to let VS warn inside the implementation of these functions that unsafe static_cast is used and that the safe variants shall be used.

PR #703, opened 2018-06-30
Dev/annagrin/fix cppcorecheck warnings

annagrin:
Fixed CppCoreCheck warnings in GSL and tests

- Added CMakeSettings.json for VS Open Folder configuration
	- So we can easily run CppCoreCheck in VS

- Fixed CppCorecheck warnings where it made sense

- Suppressed the rest
	- Some suppression does not work, so replaced by #pragma disable)
	- CppCoreCheck has noise, suppressed those with comments
	- Catch produces many warnings, blanket-supressed them all

- Had to fix clang formatting to keep attributes in place
	- clang-format does not support attributes, so I am using
          "CommentPragmas: '^ NO-FORMAT:'" to skip formatiting on them

### DIFF
--- a/include/gsl/gsl_assert
+++ b/include/gsl/gsl_assert
@@ -125,8 +125,7 @@ namespace details
 #if defined(GSL_TERMINATE_ON_CONTRACT_VIOLATION)
 
     template <typename Exception>
-    GSL_SUPPRESS(f.6) // NO-FORMAT: attribute
-    [[noreturn]] void throw_exception(Exception&&)
+    [[noreturn]] void throw_exception(Exception&&) noexcept
     {
         gsl::details::terminate();
     }
--- a/include/gsl/gsl_assert
+++ b/include/gsl/gsl_assert
@@ -21,6 +21,19 @@
 #include <stdexcept> // for logic_error
 
 //
+// make suppress attributes work for various compilers
+// Hopefully temporary until suppresion stardization occurs
+//
+#if defined(_MSC_VER)
+#define GSL_SUPPRESS(x) [[gsl::suppress(x)]]
+#else
+#if defined (__GNUC__)
+#define GSL_SUPPRESS(x) [[gsl::suppress("x")]]
+#endif // __GNUC__
+#endif // _MSC_VER
+
+
+//
 // Temporary until MSVC STL supports no-exceptions mode.
 // Currently terminate is a no-op in this mode, so we add termination behavior back
 //
@@ -103,7 +116,7 @@ namespace details
 #if defined(GSL_TERMINATE_ON_CONTRACT_VIOLATION)
 
     template <typename Exception>
-    [[gsl::suppress(f.6)]] // NO-FORMAT: attribute
+    GSL_SUPPRESS(f.6) // NO-FORMAT: attribute
     [[noreturn]] void throw_exception(Exception&&)
     {
         gsl::details::terminate();
--- a/include/gsl/gsl_assert
+++ b/include/gsl/gsl_assert
@@ -24,14 +24,15 @@
 // make suppress attributes work for various compilers
 // Hopefully temporary until suppresion stardization occurs
 //
-#if defined(_MSC_VER)
+#if defined (_MSC_VER)
 #define GSL_SUPPRESS(x) [[gsl::suppress(x)]]
 #else
-#if defined (__GNUC__)
+#if defined (__GNUC__) || defined(__clang__)
 #define GSL_SUPPRESS(x) [[gsl::suppress("x")]]
-#endif // __GNUC__
+#else
+#define GSL_SUPPRESS(x)
+#endif // __GNUC__ || __clang__
 #endif // _MSC_VER
-
 
 //
 // Temporary until MSVC STL supports no-exceptions mode.
@@ -39,6 +40,8 @@
 //
 #if defined(_MSC_VER) && defined(_HAS_EXCEPTIONS) && !_HAS_EXCEPTIONS
 #define GSL_MSVC_USE_STL_NOEXCEPTION_WORKAROUND
+#include <intrin.h>
+#define RANGE_CHECKS_FAILURE 0
 #endif
 
 //
@@ -95,10 +98,16 @@ namespace details
 #if defined(GSL_MSVC_USE_STL_NOEXCEPTION_WORKAROUND)
 
     typedef void  (__cdecl *terminate_handler)();
+    
+    GSL_SUPPRESS(f.6) // NO-FORMAT: attribute
+    [[noreturn]] inline void __cdecl default_terminate_handler()
+    {
+        __fastfail(RANGE_CHECKS_FAILURE);
+    }
 
     inline gsl::details::terminate_handler& get_terminate_handler() noexcept
     {
-        static terminate_handler handler = &abort;
+        static terminate_handler handler = &default_terminate_handler;
         return handler;
     }
 
--- a/include/gsl/gsl_assert
+++ b/include/gsl/gsl_assert
@@ -103,6 +103,7 @@ namespace details
 #if defined(GSL_TERMINATE_ON_CONTRACT_VIOLATION)
 
     template <typename Exception>
+    [[gsl::suppress(f.6)]] // NO-FORMAT: attribute
     [[noreturn]] void throw_exception(Exception&&)
     {
         gsl::details::terminate();
--- a/include/gsl/gsl_assert
+++ b/include/gsl/gsl_assert
@@ -21,13 +21,13 @@
 #include <stdexcept> // for logic_error
 
 //
-// make suppress attributes work for various compilers
+// make suppress attributes parse for some compilers
 // Hopefully temporary until suppresion stardization occurs
 //
 #if defined (_MSC_VER)
 #define GSL_SUPPRESS(x) [[gsl::suppress(x)]]
 #else
-#if defined (__GNUC__) || defined(__clang__)
+#if defined(__clang__)
 #define GSL_SUPPRESS(x) [[gsl::suppress("x")]]
 #else
 #define GSL_SUPPRESS(x)
--- a/include/gsl/gsl_assert
+++ b/include/gsl/gsl_assert
@@ -31,7 +31,7 @@
 #define GSL_SUPPRESS(x) [[gsl::suppress("x")]]
 #else
 #define GSL_SUPPRESS(x)
-#endif // __GNUC__ || __clang__
+#endif // __clang__
 #endif // _MSC_VER
 
 //
--- a/include/gsl/gsl_byte
+++ b/include/gsl/gsl_byte
@@ -18,13 +18,13 @@
 #define GSL_BYTE_H
 
 //
-// make suppress attributes work for various compilers
+// make suppress attributes work for some compilers
 // Hopefully temporary until suppresion stardization occurs
 //
 #if defined(_MSC_VER)
 #define GSL_SUPPRESS(x) [[gsl::suppress(x)]]
 #else
-#if defined(__GNUC__) || defined(__clang__)
+#if defined(__clang__)
 #define GSL_SUPPRESS(x) [[gsl::suppress("x")]]
 #else
 #define GSL_SUPPRESS(x)
--- a/include/gsl/gsl_byte
+++ b/include/gsl/gsl_byte
@@ -17,11 +17,21 @@
 #ifndef GSL_BYTE_H
 #define GSL_BYTE_H
 
+//
+// make suppress attributes work for various compilers
+// Hopefully temporary until suppresion stardization occurs
+//
+#if defined(_MSC_VER)
+#define GSL_SUPPRESS(x) [[gsl::suppress(x)]]
+#else
+#if defined(__GNUC__)
+#define GSL_SUPPRESS(x) [[gsl::suppress("x")]]
+#endif // __GNUC__
+#endif // _MSC_VER
+
 #include <type_traits>
 
 #ifdef _MSC_VER
-
-#pragma warning(push)
 
 #ifndef GSL_USE_STD_BYTE
 // this tests if we are under MSVC and the standard lib has std::byte and it is enabled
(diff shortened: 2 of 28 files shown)


beinhaerter:
It looks like you addressed much more warnings than I do in PR #642 and in a slightly different way (suppress before the line with the warning, not at function level). But I am surprised you did not #ifdef to check for different compiler support for gsl::support (MSVC vs. clang vs. gcc).

How do we want to proceed? Do you extend your PR and I close mine? Or shall we keep both open?

With my code in PR #642 and your changes I still get this warning:

> ...\gsl\include\gsl\gsl_util(141): warning C26472: Don't use a static_cast for arithmetic conversions. Use brace initialization, gsl::narrow_cast or gsl::narow (type.1).

If you add a `[[gsl::suppress(type.1)]] // NO-FORMAT: attribute"` before line 141 in gsl_util we get rid of that warning.

annagrin:
@beinhaerter I was experimenting:) I am happy to merge your changes into mine, but I would like to avoid ifdefs - currently I am trying my luck with forcing the compilers to ignore unknown attributes.

annagrin:
@beinhaerter 

> If you add a [[gsl::suppress(type.1)]] // NO-FORMAT: attribute" before line 141 in gsl_util we get rid of that warning.

Thanks, I missed that. I will use narrow_cast to fix the warning instead of suppressing in this case.

include/gsl/gsl_assert
```diff
@@ -20,6 +20,20 @@
...
 
+//
+// make suppress attributes parse for some compilers
+// Hopefully temporary until suppresion stardization occurs
+//
+#if defined (_MSC_VER)
+#define GSL_SUPPRESS(x) [[gsl::suppress(x)]]
+#else
+#if defined(__clang__)
+#define GSL_SUPPRESS(x) [[gsl::suppress("x")]]
```
neilmacintosh: Has someone opened a bug report against Clang to get this syntax variation fixed?

include/gsl/gsl_assert
```diff
@@ -110,10 +125,7 @@ namespace details
 #if defined(GSL_TERMINATE_ON_CONTRACT_VIOLATION)
 
     template <typename Exception>
-#if defined(__clang__) || defined(__GNUC__)
-#else
-    [[gsl::suppress(f.6)]]
-#endif
+    GSL_SUPPRESS(f.6) // NO-FORMAT: attribute
```
neilmacintosh: Couldn't we use a conditional noexcept for these cases, based on the throw/terminate build-mode? Yes, it might involve a macro, but probably less directly and would be no worse than suppressions. But I may be missing something else here.

include/gsl/gsl_byte
```diff
@@ -17,14 +17,28 @@
...
+
 #include <type_traits>
 
 #ifdef _MSC_VER
 
 #pragma warning(push)
 
-// don't warn about function style casts in byte related operators
-#pragma warning(disable : 26493)
+// Turn off rules that generate too much noise. To be fixed in the static analysis tool.
```
neilmacintosh: If that is MSVC /analyze, perhaps we should just call it that?

include/gsl/gsl_byte
```diff
@@ -153,6 +166,7 @@ constexpr byte to_byte_impl(T t) noexcept
     return static_cast<byte>(t);
 }
 template <>
+// GSL_SUPPRESS(type.4) // NO-FORMAT: attribute // TODO: suppression does not work
```
neilmacintosh: I can't remember, but is there a reason this can't be just `return {t};`? If there is, perhaps we should leave behind a comment to explain to others reading the code.

include/gsl/multi_span
```diff
@@ -44,6 +44,12 @@
 #pragma warning(disable : 4127) // conditional expression is constant
 #pragma warning(disable : 4702) // unreachable code
 
+// Turn off rules that generate too much noise. To be fixed in the static analysis tool.
```
beinhaerter: If you have some MS internal insights and expect these fixes to be in version x of MSVC then maybe we could only use them `#if _MSC_VER < x`.

include/gsl/multi_span
```diff
@@ -109,107 +109,127 @@ public:
...
+    GSL_SUPPRESS(bounds.2) // NO-FORMAT: attribute
+    GSL_SUPPRESS(bounds.4) // NO-FORMAT: attribute
+    constexpr const_reference operator[](std::size_t component_idx) const noexcept
     {
         Expects(component_idx < Rank); // Component index must be less than rank
         return elems[component_idx];
     }
 
-    constexpr bool operator==(const multi_span_index& rhs) const GSL_NOEXCEPT
+    GSL_SUPPRESS(bounds .1) // NO-FORMAT: attribute
```
beinhaerter: Here and in the next line we could remove the space in "bounds ." and align it with the format in the other suppresses.

include/gsl/multi_span
```diff
@@ -446,19 +478,28 @@ namespace details
...
 
-        size_type elementNum() const GSL_NOEXCEPT { return CurrentRange; }
+        GSL_SUPPRESS(c.128) // NO-FORMAT: attribute // no pointers to BoundsRanges should be ever used
+        size_type elementNum() const noexcept
+        {
+            return CurrentRange;
+        }
 
-        size_type elementNum(std::size_t dim) const GSL_NOEXCEPT
+        GSL_SUPPRESS(c.128) // NO-FORMAT: attribute // no pointers to BoundsRanges should be ever used
```
beinhaerter: If this is required would it make sense to add a private `operator&` to prevent someone taking a pointer to the object?

include/gsl/multi_span
```diff
@@ -717,60 +754,59 @@ public:
...
 
-    constexpr size_type total_size() const GSL_NOEXCEPT
+    GSL_SUPPRESS(bounds.4) // NO-FORMAT: attribute
+    constexpr size_type total_size() const noexcept
     {
         size_type ret = 0;
-        for (std::size_t i = 0; i < rank; ++i) {
-            ret += (m_extents[i] - 1) * m_strides[i];
-        }
+        for (std::size_t i = 0; i < rank; ++i) { ret += (m_extents[i] - 1) * m_strides[i]; }
```
beinhaerter: In my opinion the old loop code block was better readable than this one liner.

include/gsl/multi_span
```diff
@@ -717,60 +754,59 @@ public:
...
 
-    constexpr size_type size() const GSL_NOEXCEPT
+    GSL_SUPPRESS(bounds.4) // NO-FORMAT: attribute
+    constexpr size_type size() const noexcept
     {
         size_type ret = 1;
-        for (std::size_t i = 0; i < rank; ++i) {
-            ret *= m_extents[i];
-        }
+        for (std::size_t i = 0; i < rank; ++i) { ret *= m_extents[i]; }
```
beinhaerter: In my opinion the old loop code block was better readable than this one liner.

include/gsl/multi_span
```diff
@@ -1517,85 +1559,86 @@ public:
...
 
     template <typename OtherValueType, std::ptrdiff_t... OtherDimensions,
-              typename = std::enable_if_t<std::is_same<
-                  std::remove_cv_t<value_type>, std::remove_cv_t<OtherValueType>>::value>>
-    constexpr bool
-    operator==(const multi_span<OtherValueType, OtherDimensions...>& other) const GSL_NOEXCEPT
+              typename = std::enable_if_t<std::is_same<std::remove_cv_t<value_type>,
+                                                       std::remove_cv_t<OtherValueType>>::value>>
+    constexpr bool operator==(const multi_span<OtherValueType, OtherDimensions...>& other) const
+        noexcept
```
beinhaerter: Previously the `GSL_NOEXCEPT` was on the same line. If there is no reason to put it on a new line, I would leave it there. Or is this something the editor automatically does?

include/gsl/multi_span
```diff
@@ -2019,96 +2058,90 @@ private:
 
     pointer data_;
     const Span* m_validator;
-    void validateThis() const
-    {
+
+    GSL_SUPPRESS(bounds.1) // NO-FORMAT: attribute
+    void validateThis() const {
```
beinhaerter: "{" in the next line would be more consistent with the rest of the code.

include/gsl/multi_span
```diff
@@ -2019,96 +2058,90 @@ private:
...
     contiguous_span_iterator(const Span* container, bool isbegin)
         : data_(isbegin ? container->data_ : container->data_ + container->size())
         , m_validator(container)
-    {
-    }
+    {}
 
 public:
-    reference operator*() const GSL_NOEXCEPT
+    reference operator*() const
```
beinhaerter: Why is noexcept removed here? I know that `validateThis` is not noexcept, but noexcept is not removed for `operator ->` which looks quite similar.

include/gsl/span
```diff
@@ -408,15 +400,19 @@ public:
     }
 
     template <std::size_t N, class ArrayElementType = std::remove_const_t<element_type>>
+    // GSL_SUPPRESS(bounds.4) // NO-FORMAT: attribute // TODO: parser bug
     constexpr span(std::array<ArrayElementType, N>& arr) noexcept
         : storage_(&arr[0], details::extent_type<N>())
     {
+
```
beinhaerter: In other places you are replacing "{\n}" by "{}", here you are adding a new line?

tests/CMakeLists.txt
```diff
@@ -49,6 +49,7 @@ target_compile_options(gsl_tests_config INTERFACE
         -Werror
         -Wextra
         -Wno-missing-braces
+        -Wno-unknown-attributes
```
beinhaerter: I bet you have tested it, but I cannot find any information online describing what `-Wunknown-attributes` or `-Wno-unknown-attributes` do. And confusingly enough I find code that uses `-Wunknown-attribute` (no 's'). Do you have a link to a page describing it (gcc?, clang?)?

beinhaerter:
I like getting rid of `GSL_NOEXCEPT` and seeing more warnings being suppressed or addressed. Also thanks for doing other cleanup like adding const.

I would have split this into one PR for the `GSL_NOEXCEPT` -> noexcept change (no-brainer) and another (one or two) for the warnings.

include/gsl/gsl_byte
annagrin: The only reason is that it does not work with c++14... will add comment.

include/gsl/gsl_byte
```diff
@@ -17,14 +17,28 @@
...
 #include <type_traits>
 
 #ifdef _MSC_VER
 
 #pragma warning(push)
 
-// don't warn about function style casts in byte related operators
-#pragma warning(disable : 26493)
+// Turn off rules that generate too much noise. To be fixed in the static analysis tool.
+#pragma warning(disable : 26493) // don't use c-style casts
```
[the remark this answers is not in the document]
annagrin: Had to use #pragma suppress because MSVC has a bug where suppression does not always work in templates. To be fixed...

include/gsl/gsl_assert
```diff
@@ -20,6 +20,20 @@
 #include <exception>
 #include <stdexcept> // for logic_error
 
+//
+// make suppress attributes parse for some compilers
+// Hopefully temporary until suppresion stardization occurs
```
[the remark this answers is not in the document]
annagrin: thanks!

include/gsl/gsl_byte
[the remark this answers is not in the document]
annagrin: Added comment to clarify

include/gsl/multi_span
annagrin: Not yet, will do when we have the fixes planned.

include/gsl/multi_span
annagrin: There seems to be an issue with clang-format where it inserts the space even when it is supposed to ignore the commented line. Will fix.

include/gsl/multi_span
annagrin: Makes sense, will add as a separate PR later, thanks!

include/gsl/multi_span
annagrin: Those are clang-format changes... Normally I would not introduce formatting changes at the same time as code changes, but had to do so in this case (clang-format does not support attributes)

include/gsl/multi_span
annagrin: yeah, this is clang-format

include/gsl/multi_span
annagrin: will try to keep revert it, will see if clang-format will change it again

include/gsl/multi_span
annagrin: mistake in `operator->` ... will fix

include/gsl/span
annagrin: thanks for catching!

include/gsl/string_span
```diff
@@ -510,7 +501,7 @@ bool operator!=(const T& one, gsl::basic_string_span<CharT, Extent> other) GSL_N
 template <typename CharT, std::ptrdiff_t Extent = gsl::dynamic_extent, typename T,
           typename = std::enable_if_t<std::is_convertible<
               T, gsl::basic_string_span<std::add_const_t<CharT>, Extent>>::value>>
-bool operator<(gsl::basic_string_span<CharT, Extent> one, const T& other) GSL_NOEXCEPT
+bool operator<(gsl::basic_string_span<CharT, Extent> one, const T& other)
```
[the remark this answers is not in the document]
annagrin: thy shouldn't, thanks!

tests/CMakeLists.txt
annagrin: I couldn't find any documentation but clang gives  the following warnings on unknown attrobutes:

> warning: unknown attribute ... ignored [-Wunknown-attributes]

So I took a guess at removing that:)

annagrin: Can't reply to your comment below - sounds reasonable, created PR #719

neilmacintosh:
Thanks @annagrin, that's a huge diff! Looks good to me.

PR #708, opened 2018-07-17
Make library importable by CMake using find_package

### LINKED ISSUE #715: CMake project not importable using find_package
Currently no GSLConfig.cmake file is installed so find_package can not be used with GSL.

Exporting the targets and having an install script is a crucial requirement for GSL to be used with CMake based package manager like [hunter](https://hunter.sh). Even without any package manager `find_package` is the most clean way to import dependencies in CMake.

The CMakeLists.txt should be changed to include an export and install rule.

ithron:
Currently no GSLConfig.cmake file was installed so find_package could not be used with GSL.
Fixed CMakeLists.txt to install GSLConfig.cmake so that find_package can be used.
Fixes #715.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,7 +74,7 @@ if ((CMAKE_VERSION GREATER 3.7.9) OR (CMAKE_VERSION EQUAL 3.7.9))
     endif()
 endif()
 
-install(TARGETS GSL EXPORT GSLConfig
+install(TARGETS GSL EXPORT Microsoft.GSLConfig
   PUBLIC_HEADER DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
 )
 install(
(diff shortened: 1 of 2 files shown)


OvermindDL1:
Doesn't this conflict with GSLConfig.cmake from the GSL scientific library project that this project's name conflicts with?  Is there any way to prevent one from overwriting the other on the same system?

ithron:
Yes, there is a conflict with GNU Scientific Library.
Changed the package name to Microsoft.GSL. The exported target GSL now resides in the namespace Microsoft.GSL::
So:

`find_package(Microsoft.GSL)`
`target_link_libraries(MyTarget PUBLIC Microsoft.GSL::GSL)`

I think this might be a good compromise.

OvermindDL1:
The namespace is great, but it looks like the filename itself was still `GSLConfig.cmake` on line 86?

ithron:
Ah, yes, only changed the install line. Fixed that. (Microsoft.GSLConfig.cmake)

ithron:
Should be ready now

jimvonmoon:
Has this PR gone stale? It is a great initiative to make GSL available as an imported target. This seems to be the standard approach now.

Warchant:
Please land this fix

pr0g:
Hey there,

I wound up implementing this myself recently for a library I wanted to use the GSL with.

I've created a fork here:

    https://github.com/pr0g/GSL

and the diff looks like this:

    https://github.com/microsoft/GSL/compare/master...pr0g:cmake-find-package.

It's pretty similar to this change - I'm using my fork in my project right now.

I know there's also this PR https://github.com/microsoft/GSL/pull/784 so I don't want to step on anyone's toes but I would love if support could officially be added. I'd be happy to create a PR with my fork/branch if that would help or try and update this one?

Thanks!


Tom

JordanMaples:
Maintainer's call: Marking as duplicate of PR #784

PR #709, opened 2018-07-25
Lower required cmake version to 3.1.3

ithron:
The original project required cmake 3.1.3.
Since cmake 3.7 is not required here, changed it back to 3.1.3


ithron:
Sorry, started to PR against the wrong repo.

Issue #699, opened 2018-06-18
2fc94db made gsl::not_null constructor explicit, why?


kivadiu:
Since 2fc94db3ebfb1b066edeafac1837f34d6111bff4, this program does not work anymore without explicit cast to not_null (here tested with clang 5.0.2 on linux with -std=c++14):
```
#include <gsl/pointers>
#include <iostream>

void f(gsl::not_null<const int *> i) {
  std::cout << *i << '\n';
}

int main() {
  int i = 42;
  f(&i); // no matching function for call to 'f'
         // note: candidate function not viable: no known conversion from 'int *' to 'gsl::not_null<const int *>' for 1st argument
  return 0;
}
```

If this is really wanted, could you explain why?

Kind regards

ericLemanissier:
I guess you already checked the discussions pointed by the commit you gave: https://github.com/Microsoft/GSL/pull/659 and https://github.com/Microsoft/GSL/issues/395
It is true that the specific example you give here does not gain on security nor readability with this change. Maybe a constructor of not_null taking reference as parameter would help ?

kivadiu:
I have tried the following program with/without explicit constructor of `gsl::not_null`. I get exactly the same behaviour at runtime. The only difference is at compile time where I must add a lot of text if the constructor is explicit: `gsl::not_null<const int*>(...)`. It is so painful that it discourages to use `gsl::not_null`. Could you elaborate on the real usefulness of this explicit constructor? It does not seem to me that the check is done earlier because the construction of the `gsl::not_null` is always done before the function is called, at the point of the call.

```
#include <gsl/pointers>
#include <iostream>

void f(gsl::not_null<const int *> i) {
  std::cerr << static_cast<const void *>(i.get());
  std::cerr << ": " << *i << '\n';
}

int *never_null_int() {
  return new int{42};
}

int *sometimes_null_int(int choice) {
  return choice == 0 ? nullptr : new int{43};
}

int main() {
  f(gsl::not_null<const int *>(never_null_int()));
  f(gsl::not_null<const int *>(sometimes_null_int(1)));
  f(gsl::not_null<const int *>(sometimes_null_int(0)));
  return 0;
}
```

twadrianlis:
I am also surprised by this change - the whole point of not_null to me was an easy drop-in contract check encoded in function signatures that informed the caller that:

- hey, I expect you play nice and give me decent pointer, otherwise I will bail-out

[...]

PR #711, opened 2018-08-01
Dev/annagrin/make not null

annagrin:
Following suggestion from [@kivadiu](https://github.com/kivadiu), added `make_not_null` helper to create a `not_null`
    
Introduction of explicit `not_null` constructor made it cumbersome to create `not_null`s, especially in before c++17. Adding `make_not_null` helper. Usage (see tests):
    
```
    int i = 42;
    
    auto x = make_not_null(&i);
    helper(make_not_null(&i));
    helper_const(make_not_null(&i));
```

See #699

### DIFF
--- a/include/gsl/pointers
+++ b/include/gsl/pointers
@@ -120,6 +120,11 @@ private:
 };
 
 template <class T>
+auto make_not_null(T&&t) {
+    return gsl::not_null<std::remove_cv_t<std::remove_reference_t<T>>>{t};
+}
+
+template <class T>
 std::ostream& operator<<(std::ostream& os, const not_null<T>& val)
 {
     os << val.get();
--- a/.travis.yml
+++ b/.travis.yml
@@ -124,7 +124,7 @@ matrix:
         apt:
           packages:
             - clang-5.0
-            - g++-5
+            - g++-7
           sources:
             - ubuntu-toolchain-r-test
             - llvm-toolchain-trusty-5.0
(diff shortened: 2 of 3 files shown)


include/gsl/pointers
```diff
@@ -119,6 +119,11 @@ private:
     T ptr_;
 };
 
+template <class T>
+auto make_not_null(T&&t) {
```
neilmacintosh: maybe run clang-format? `T&&t` looks funny to me.

include/gsl/pointers
```diff
@@ -119,6 +119,11 @@ private:
...
+    return gsl::not_null<std::remove_cv_t<std::remove_reference_t<T>>>{t};
```
neilmacintosh: If the purpose of `make_not_null` is just to forward to the `not_null` constructor, isn't `std::forward` the right thing to use here instead?

kivadiu: Nice addition, thanks.
And I agree with @neilmacintosh, should it be
```
template <class T>
auto make_not_null(T&&t) {
    return gsl::not_null<std::remove_cv_t<std::remove_reference_t<T>>>{std::forward<T>(t)};
}
```

kivadiu:
I have tested your PR in my code with gcc 8.2.0 and clang 5.0.2. Works fine. Thanks. It helps a lot.
Looking forward to pull the merge from master.

PR #712, opened 2018-08-02
Dev/annagrin/sloppy not null

annagrin:
Created a helper class to help move code that uses implicit `not_null` constructor to explicit. The transition would work as follows:

- replace all occurrences of `not_null` in your code by `sloppy_not_null `
- compile - compilation should be successful 
 - replace some` sloppy_not_null`s by `not_null`, fix compilation errors,  
redesign as needed, compile and test 
- repeat until no `sloppy_not_null`s remain 

The transition to explicit `not_null` constructor is worth it:

- it minimized checking for null
  Including cases when `not_null` is silently transformed into a raw pointer and back along the call chain
- it makes API and developer intentions cleaner
 It propagates not_null types on API to their correct place, removes assertions and simplifies code
- it follows C++ Core Guidelines
  `not_null` should be a stricter type than a raw pointer, so implicit conversion is hiding useful information and potential failures

In the near future, tooling should be created to help insert explicit constructor calls and make the transition easier.

This is partially resolves issue #699 (the rest needs to be done in tooling).


annagrin:
Done, thanks!

PR #719, opened 2018-08-13
Dev/annagrin/remove gsl noexcept

annagrin:
removed remaining GSL_NOEXCEPT to make sure we have consistent APIs
    
    - Removed GSL_NOEXCEPT macro
    - Replaced by noexcept keyword when needed
    - removed noexcept where a function could throw


neilmacintosh:
Thanks @annagrin! I take it that you plan to refine which functions should actually have `noexcept` (for example, I think the comparison ops in `string_span` probably should not), in a subsequent PR and this one is just switching from macro to keyword in a compilable way?

PR #720, opened 2018-08-13
Fix finally when passing a non const lvalue invokable thing

menuet:
Hi,

I believe that the function finally should be modified because the deduced type is an lvalue reference when we pass it an lvalue. Then the member f_ would be a reference and this would be dangerous if we move a final action out of the created scope.
I propose a fix. What do you think ?

### DIFF
--- a/include/gsl/gsl_util
+++ b/include/gsl/gsl_util
@@ -52,6 +52,8 @@ template <class F>
 class final_action
 {
 public:
+    static_assert(!std::is_reference<F>::value && !std::is_const<F>::value && !std::is_volatile<F>::value, "Final_action should store its callable by value");
+
     explicit final_action(F f) noexcept : f_(std::move(f)) {}
 
     final_action(final_action&& other) noexcept : f_(std::move(other.f_)), invoke_(std::exchange(other.invoke_, false)) {}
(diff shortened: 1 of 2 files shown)


include/gsl/gsl_util
```diff
@@ -74,17 +76,26 @@ private:
 };
 
 // finally() - convenience function to generate a final_action
+//#define TRY_WITH_BUG // DEFINE THIS TO SEE THAT TEST CASE "finally_lambda_lvalue" fails to compile
```
beinhaerter: In my opinion this #if/#else shouldn't be in the final commit, it should be a comment for the commit.

menuet: yes, sorry, I should have stated here that my #ifdef was only for temporarily demonstrating the issue.
Of course, it should not be part of the final commit.
I can prepare a cleaner commit if you like.

menuet:
I have removed the #ifdef in my branch.
Do you think it is ok now ?

beinhaerter:
@menuet Just for your information, I am not in the position to accept or reject a pull request, so please don't expect me to get active.

annagrin:
Thanks @menuet!

menuet:
Hi @annagrin 
I had forgotten about this pull request :-)
So, what is the status of it? Don't you think something should be changed in the function finally?
I think this function should have only the overload with && and use std::remove_cvref_t inside it.

JordanMaples:
@menuet this change looks good, but would you mind rebasing it and updating the tests? Thanks.

PR #724, opened 2018-08-19
Dev/annagrin/better check for branch opt

annagrin:
Made a check in branch optimization inside `span::operator[]` simpler an saner.


neilmacintosh:
Looks good to me, thanks @annagrin!

PR #725, opened 2018-08-22
Allow #include <gsl/...> without repeating gsl

### LINKED ISSUE #71: Template specialization for ArrayViewTypeTraits uses incorrect static member of is_reference
```C++
struct ArrayViewTypeTraits<Traits, typename std::is_reference<typename Traits::array_view_traits &>::type>
```

```std::is_reference<T>::type``` isn't valid. It should be using ```std::is_reference<T>::value```.

kivadiu:
Headers are named inconsistently: algorithm, assert, byte and util
headers are prefixed with gsl_. This is to avoid name clash with
standard headers for people including directly from the gsl directory.

However, for people including from the parent directory of gsl, in
addition to looking inconsistent, this also looks redundant:
#include <gsl/gsl_assert>

This adds algorithm, assert, byte and util headers that just forward to
gsl_ headers. It is then possible to include like this:
#include <gsl/assert>

Fixes #71

### DIFF
--- a/include/gsl/algorithm
+++ b/include/gsl/algorithm
@@ -1,0 +1,22 @@
+///////////////////////////////////////////////////////////////////////////////
+//
+// Copyright (c) 2018 Microsoft Corporation. All rights reserved.
+//
+// This code is licensed under the MIT License (MIT).
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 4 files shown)


JordanMaples:
Maintainer's call: We see your point about the inconsistently named headers. We agree this is... unfortunate. If we were to fix it, we'd probably do it similarly to how you suggest. However, we don't think it's worth the effort of creating additional headers at this time, it could cause confusion for newcomers to see a lot of headers that are duplicates.

Issue #680, opened 2018-05-09
Lack of deduction guides for gsl::span and similar types


effolkronium:
Pls, implement something like that (if C++17 support exist)

```
namespace gsl
{
    template<class T>
    span(T&)->span<typename T::value_type>;

    template<class T, size_t N>
    span(T(&)[N])->span<typename std::remove_all_extents<T>::type>;
}
```

annagrin:
some work for non_null is done in #689

johelegp:
The [standard's](http://eel.is/c++draft/span.overview) has some.

annagrin:
BTW, I laid the grounds for this work to start with my addition of c++17 tests (#692, #697). Gave a starter with  not_null constructor from T and  template arguments deduction tests (#689). PRs are welcome for span deduction guides! We might need to bear in mind that some deduction guides could be just missing constructors...

JordanMaples:
Maintainers' call: Sorry for the delay. Now that span has been standardized, the only delta between gsl::span and std::span is the bounds checking. We are following the standard interface exactly.

PR #729, opened 2018-09-02
Initial support for CTAD for spans.

LYP951018:
Related issue #680 .

### DIFF
--- a/include/gsl/span
+++ b/include/gsl/span
@@ -623,6 +623,26 @@ private:
     }
 };
 
+
+template<typename T, std::size_t N>
+span(std::array<T, N>&) -> span<std::remove_const_t<T>, static_cast<std::ptrdiff_t>(N)>;
+
+template<typename T, std::size_t N>
+span(const std::array<T, N>&) -> span<std::add_const_t<T>, static_cast<std::ptrdiff_t>(N)>;
+
+template<typename T, std::size_t N>
+span(T(&)[N]) -> span<T, static_cast<std::ptrdiff_t>(N)>;
+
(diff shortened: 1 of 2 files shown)


LYP951018:
It seems that I forgot to wrap deduction guides with cplusplus. I'll fix that soon.

LYP951018:
I think we need some macro to indicate whether deduction guide is available.

annagrin:
Thanks @LYP951018.  How about this one: __cpp_deduction_guides? 

I found it in https://en.cppreference.com/w/cpp/feature_test, but haven't tried myself yet.

jwakely:
`#if __cpp_deduction_guides >= 201606` is the right check.

LYP951018:
My concerns here is that there will be many places require something like:

```cpp
#if __cpp_deduction_guides >= 201606
```

And I should introduce some macro:

```cpp
#if __cpp_deduction_guides >= 201606
#define GSL_HAS_DEDUCTION_GUIDES
#endif 
```

Where should I put the macro definition in?

MikeGitb:
How is checking for that macro better than directly checking for the standard feature macro?

Btw.: Why do we have to check for a particular version instead of just checking for the presence?

jwakely:
You don't *have* to, but the value tells you if the version you care about is supported. For some features different values mean different things are supported.

hsutter:
Guidelines editors call: Yes please add CTAD. Thanks!

JordanMaples:
Closing this PR. Deduction guides were was added as a part of #844.

PR #731, opened 2018-09-10
Suppress Warnings in GSL Headers

hannesweisbach:
This fixes the issue raised in PR #465. While the proposed solution in #465 depended on compiler-specific macros, this patch uses a build-system feature and therefore does not depend on compiler-specific behaviour.
Additionally, it treats the headers only as system headers for non-standalone builds, so that GSL-developers are still able to see and fix legitimate warnings.


annagrin:
Thanks @hannesweisbach!

Issue #487, opened 2017-04-16
CMake version


tiagomacarios:
Is there a reason for using CMake 2.8.7 (2012?)?
If a newer version was used one could improve the CMake files to have a more modern pattern. GSL could be an interface target for example.

neilmacintosh:
We started using a "lowest common denominator" version at first to ensure broad accessibility. We also want to use continuous integration in Git, although I don't think that is currently limiting us to 2.8.7 (I haven't checked recently to be honest).

I'd be happy to look at a PR that upgrades the requirement if it demonstrates a function improvement that we can't have with today's CMake.

rianquinn:
@neilmacintosh One thing that I ran into with VS2017 is that it has a hacked up version of Git built into it, which breaks Windows Git SCM. Although the issue is not specific to CMake, it was showing up as an issue with CMake because ExternalProject_Add (something this repo should might want to use in the future instead of submodules), was broken as a result. There is a fix for it in CMake, but it requires the latest 3.8. 

Just food for thought.

CaseyCarter:
The rationale for using ancient cmake was that Travis CI uses cmake 2.8.ancient. The Travis and AppVeyor scripts now install a more recent version of cmake (IIRC 3.7.2) manually before they configure and build the tests.

tiagomacarios:
Not really sure what version would be good. If it is too old you cannot use newer CMake features, if it is too new people normally don't have it on their system. Anyway, I [forked a branch](https://github.com/tiagomacarios/GSL/commit/691fd1cd0cab0666ce3227a1db20d3cf543fe65c) were the CMake files use newer features.

Main things:
CMake itself will automatically check for c++14 support.
GSL is now an INTERFACE and one only need to link against that interface (include paths and natvis files will be taken care automatically).
When generating a VS project the gsl tests will be properly nested into a test subfolder.
`if(compiler)` expression were modified to generator expressions.

I can try to break this changes into smaller ones to see what the CI will accept or not. Let me know what you think.

@rianquinn What is the reason to use ExternalProject_Add instead of the submodule? Just curious. I normally think that submodules make things easier, but it is not unusual to find people that disagree =)

neilmacintosh:
Thanks for the update @tiagomacarios I'd be happy to look at PRs that brought the changes over in discrete chunks.

a) so we can ensure CI remains happy at each step
b) because I like to review changes that are logically related rather than one big "all the CMake things" PR

rianquinn:
[...]

PR #741, opened 2018-10-15
Clang-cl: Fix #739 compatibility macros & add to AppVeyor

### LINKED ISSUE #739: Compatibility fix GSL_SUPPRESS not working with clang-cl.exe
When building with clang-cl on Windows, the compatibility fix for clang is not used.

clang-cl.exe defines both the macros `_MSC_VER` and `__clang__`.
But the compatibility fix, as added with #703, checks for `_MSC_VER` first.

Applies to:
[include/gsl/gsl_byte](https://github.com/Microsoft/GSL/blob/master/include/gsl/gsl_byte)
[include/gsl/gsl_assert](https://github.com/Microsoft/GSL/blob/master/include/gsl/gsl_assert)

clang v7.0.0

### LINKED ISSUE #745: [Travis CI] Test broken size fails with GCC 6.5
The span.size() bug reported in #590 remains in GCC 6.5 (as now used in Travis CI).
The test added with #704 now breaks the CI: [GSL build report](https://travis-ci.org/Microsoft/GSL/jobs/454937553)

According to the [gcc bugtracker: 83116](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=83116), the fix was supposed to have landed in GCC 6.5.

I [added a fix](https://github.com/Microsoft/GSL/pull/741/commits/84f57949abd4b69ac4003daa16537202096ea71f) while working on #741, should I make it a separate pull request?

Farwaykorse:
As described in #739.
When building on Windows with clang-cl.exe (the compatible replacement for cl.exe), both the macros `_MSC_VER` and `__clang__` are defined.

To get the temporary compatibility fix (#703) working with clang-cl, the check for `__clang__` has to come first.

----
Adding compilation tests with AppVeyor:
- [x] Investigate curious test failure in Travis CI
  Build 26: on Linux `COMPILER=g++-6 BUILD_TYPE=Release GSL_CXX_STANDARD=14`
  Resolves #745
- [x] Compiling with the latest clang-cl at [AppVeyor](https://www.appveyor.com/docs/windows-images-software/#llvm)
  Using the VS 2017 and the LLVM toolset integration (vs extension available on AppVeyor). This tests the use of clang-cl from VS projects.
- [x] ~~Temporary fix for the lld-linker and vcpkg (version installed on AppVeyor)~~ --> No issue with Ninja
  ~~(link using link.exe or update the installed vcpkg version)~~
  - [x] Request update of vcpkg with AppVeyor appveyor/ci#2742
  -  ~~Replace the `git pull` of vcpkg with a fallback to the msvc linker~~
    Works locally but can't get it working on AppVeyor. ~~Therefore sticking with pulling vcpkg.~~ --> Ninja
  - [x] ~~The `git pull` command on AppVeyor appears to generate a 'remote exception' bce6ed2~~ --> Ninja
- [x] Blocking errors
  - [x] "error : cannot use 'try'/'throw' with exceptions disabled" while building Catch
  - [x] `-fno-exceptions` is not recognised by clang-cl'; results in: `-Wunknown-argument` 100f2b9
  - [x] No support for `-std=c++1z`; ...
    `clang-cl.exe : error : unknown argument ignored in clang-cl: '-std=c++1z' [-Werror,-Wunknown-argument]`
    Fix: When building with clang-cl use the version of CMake installed on AppVeyor (v3.12.2) . 9a625ab
    __Is there a reason for using CMake v3.8.0?__ Yes: Issue #487 discusses the use of older CMake versions to ensure accessibility.
- [x] Potentially valid warnings
  - [x] `-Winvalid-noreturn` 2238c47  __Enabled by Default__
    `gsl\include\gsl/gsl_assert(123): error : function declared 'noreturn' should not return [-Werror,-Winvalid-noreturn]`
  - [x] `-Wnewline-eof` 137367b __-Weverything__
    `gsl\tests\multi_span_tests.cpp(1785): error : no newline at end of file [-Werror,-Wnewline-eof]`
  - [x] `-Wunused-member-function` 337a03c __-Weverything__
    `gsl\tests\span_tests.cpp(57): error : unused member function 'operator&' [-Werror,-Wunused-member-function]`
  - [x] `-Wdouble-promotion` f810ce7 __-Weverything__
    `gsl\tests\notnull_tests.cpp(196): error : implicit conversion increases floating-point precision: 'float' to 'double' [-Werror,-Wdouble-promotion]`
  - [x] `-Wmissing-noreturn` 1b15a73 __-Weverything__
    `gsl\tests\no_exception_ensure_tests.cpp(29): error : function 'test_terminate' could be declared with attribute 'noreturn' [-Werror,-Wmissing-noreturn]`
    `gsl\tests\no_exception_throw_tests.cpp(28): error : function 'test_terminate' could be declared with attribute 'noreturn' [-Werror,-Wmissing-noreturn]`
  - [x] `-Wmissing-variable-declarations` 329da83 __-Weverything__
    `gsl\tests\utils_tests.cpp(83): error : no previous extern declaration for non-static variable 'j' [-Werror,-Wmissing-variable-declarations]`
- [x] Update README.md: supported platform Clang-cl 7.0.0 on Windows. 07ec87b

Extra:
- [x] ~~Replace MSBuild with~~ Add Ninja-build. AppVeyor execution time reduction with 60%.
  - [x] Manually add target architecture flags to clang-cl 449eec4
    [Cmake issue: Ninja generator fails with clang-cl when mixing 64/32 bit clang and Visual Studio Environment](https://gitlab.kitware.com/cmake/cmake/issues/16259)
  - [x] No need for the lld-link workaround when using Ninja. 76375f1

Improvements: (After review by annagrin)
  - [x] Separate CMake generation from build. 4631e6e
  - [x] "Clearer"/less convoluted compile flag configuration for the tests. caa29d0
    \+ 'flatter' decision structure; - some duplication; + more specific to each compiler
    - [x] Remove redundant warning flags for clang-cl (checked documentation since v4.0.0)
      Note: `/W4` enables `-Wall` and `-Wextra`. 2bdb8f9 767f34c 595a726
    - [x] Remove flag for clang-cl `-fno-strict-aliasing` no issues building without it. b1e63fd
  - [x] Move suppression of `-Wunused-member-function` from `span_tests.cpp` to configuration.
    note: With `-Weverything` add: `$<$<EQUAL:${GSL_CXX_STANDARD},14>:-Wno-unused-member-function>` c98c546
  - [x] Why those additional `-Wno-*` flags? - Local issue. ~~Remove flags. 7b85156~~ - pushed warning level.
    - [x] Check changes. _(marked in this checklist)_
  - [x] Different solution for [[noreturn]] in 2238c4760e86fe
  - ~~Find first working CMake version.~~ Proper flags for clang-cl. v7.0.0 does not recognize  `-std=c++1z`.
    Both CMake and clang on AppVeyor are updated regularly. Unless both can be pinned, similar issues will appear again. Besides, this exercises the usage of the latest CMake release.
  - [x] Return MSBuild generators. 2197612
  - [x] Push warning level.
    - [x] LLVM/clang-cl: `-Weverything` 9ef9808
      - [x] `-Wc++98-compat` and `-Wc++98-compat-pedantic` 57ac640, bef30b8
      - [x] `-Wmissing-prototypes` in tests f96eca3
      - [x] `-Wundef` on macro `__GNUC__`  a987075
    - [x] LLVM/clang++ (Linux & OSX): `-Weverything` 
      - [x] `-Wweak-vtables` - suppressed fcbbe1f
      - [x] `-Wundef` on macro `_MSC_VER` eb60986
      - [x] `-Wpadded` in gsl_util:51 and multi_span:1231 - suppressed fe882ce
      - [x] `-Wundefined-func-template` only in clang v5.0.2. e7812cb
      - [x] `-Wdeprecated` solved e7812cb
        > definition of implicit copy assignment operator for 'static_bounds' is deprecated because it has a user-declared copy constructor
  - [x] Clang-cl compilation with VS2015 (using the v140 MS STL). f6f7579
    - [x] `-Wkeyword-macro` Clang does not require the `constexpr` disabling macro. f40a69e

Resolves #739


annagrin:
@Farwaykorse thanks for the fix, can we also add test cases in travis.yml to make this always true?

Farwaykorse:
Compiling with clang-cl in Appveyor seems possible, but I don't have any experience with this.

Farwaykorse:
I've made an error while reordering the options.
I'll wait for Appveyor to finish the last run, to see if there are more issues, before committing.

Farwaykorse:
Issue with lld-link and the vcpkg install:
- https://github.com/Microsoft/vcpkg/issues/4359
- https://bugs.llvm.org/show_bug.cgi?id=38985

appveyor.yml
```diff
@@ -15,29 +15,48 @@ image:
...
-      $env:PATH="C:\cmake-3.8.0-win32-x86\bin;$env:PATH"
+      # Workaround for lld bug with vcpkg integration
+      if ($env:USE_TOOLSET -eq "LLVM" -and
+          (Get-Content C:\Tools\vcpkg\toolsrc\VERSION.txt).Replace('.','').Trim('"') -lt 20181020
+      ) {
+        echo "Updating vcpkg..."
+        cd C:\Tools\vcpkg
+        git pull > $NULL
+        cd "$env:APPVEYOR_BUILD_FOLDER"
+      } else { echo "==> vcpkg has been updated, please remove the workaround from appvayor.yml." }
```
CaseyCarter:
```suggestion
      } else { echo "==> vcpkg has been updated, please remove the workaround from appveyor.yml." }
```

tests/span_tests.cpp
```diff
@@ -54,7 +54,6 @@ struct DerivedClass : BaseClass
 };
 struct AddressOverloaded
 {
-    AddressOverloaded operator&() const { return {}; }
```
CaseyCarter: This `operator&` overload is the *raison dêtre* for having this type and using it in the tests - it's to ensure that everything which should be using `std::addressof` instead of `&` is doing so.

CaseyCarter: You *could* add a call somewhere specifically to silence the warning and avoid all the boilerplate, but I agree that the boilerplate makes it blatantly clear that this function is unused for good reason.

Farwaykorse:
How about replacing the MSBuild configurations in AppVeyor with Ninja?
This reduces the roundtrip time for the builds with quite a lot.

tests/span_tests.cpp
Farwaykorse: This being the test suite I think the boilerplate would be acceptable.

Farwaykorse:
@annagrin 
This PR now contains a collection of related changes.
- The original fix for the compatibility macros.
- The requested AppVeyor configuration for Clang-cl.
- A series of minor changes to the GSL based on Clang warnings, that do not show up on in Travis.
- The fix for #745.
  Here to pass the TravisCI compilation, but could have been its own PR.

And two changes in the AppVeyor implementation:
- Use of the CMake version on AppVeyor. (Only when compiling with clang.)
  Clang-cl v7.0.0 is not recognised and receives the wrong flags.
  Issue #487 discusses the use of older CMake versions to ensure accessibility.
  __Q:__ Is it acceptable to use the latest CMake with clang-cl? Or should the first working version be used?
- Replacing MSBuild with Ninja on AppVeyor.
  This significantly reduces the build times. But it is a little less like the default usage scenario with Visual Studio.
  __Q:__ Do you agree with the buildsystem switch?

include/gsl/gsl_assert
```diff
@@ -113,19 +113,20 @@ namespace details
 
 #endif
 
-    [[noreturn]] inline void terminate() noexcept
-    {
 #if defined(GSL_MSVC_USE_STL_NOEXCEPTION_WORKAROUND)
-        (*gsl::details::get_terminate_handler())();
+    inline void terminate() noexcept { (*gsl::details::get_terminate_handler())(); }
```
annagrin: what is the reason for removing the [[noreturn]] here?

include/gsl/gsl_assert
```diff
@@ -113,19 +113,20 @@ namespace details
...
-        std::terminate();
+    [[noreturn]] inline void terminate() noexcept { std::terminate(); }
 #endif
-    }
 
 #if defined(GSL_TERMINATE_ON_CONTRACT_VIOLATION)
 
     template <typename Exception>
+#if defined(GSL_MSVC_USE_STL_NOEXCEPTION_WORKAROUND)
+    void throw_exception(Exception&&) noexcept
```
annagrin: same question as above

tests/CMakeLists.txt
```diff
@@ -34,7 +34,7 @@ endif()
 # please try to keep entries ordered =)
 add_library(gsl_tests_config INTERFACE)
 target_compile_options(gsl_tests_config INTERFACE
-    $<$<CXX_COMPILER_ID:MSVC>:
+    $<$<AND:$<PLATFORM_ID:Windows>,$<OR:$<CXX_COMPILER_ID:MSVC>,$<CXX_COMPILER_ID:Clang>>>:
```
annagrin: The logic got complicated here, looks like it's worth creating separate cases for msvc, clang-cl, clang, gcc options, and splitting the code accordingly, even if some options are repeated sometimes - it will help the code look intentional and maintainable.

For example, before, we had msvc vs the rest cases, now it looks like clang-cl will have a mix of msvc and clang options, so it looks like a logic error. Is it intentional?

tests/CMakeLists.txt
```diff
@@ -57,6 +57,12 @@ target_compile_options(gsl_tests_config INTERFACE
         -Wshadow
         -Wsign-conversion
     >
+    $<$<CXX_COMPILER_ID:Clang>:
```
annagrin: Can you please clarify why those are needed - anything different in clang-cl that we didn't need in clang? or something added in latest version of clang?

annagrin:
@Farwaykorse Thank you very much for the considerable chunk of work! I left a few comments, mostly asking for clarification,  and requested a change for the Makelists.txt to make the intentions clearer.

appveyor.yml
```diff
@@ -13,44 +13,79 @@ image:
...
+          $env:CXXFLAGS = "-m64";
+        }
       }
-      echo generator="$env:generator"
-      cmake .. -G "$env:generator" -DGSL_CXX_STANDARD="$env:GSL_CXX_STANDARD"
 
 build_script:
-  - cmake --build . --config %CONFIGURATION% -- /m /v:minimal
+  - call "%VCVARSALL%" %ARCHITECTURE%
+  - cmake .. -G "Ninja" -DGSL_CXX_STANDARD="%GSL_CXX_STANDARD%"
```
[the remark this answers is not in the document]
Farwaykorse: _After some testing:_ Yes, these can move to the `before_build` section.
A left-over from some difficulty with the scope of `vcvarsall.bat`.

include/gsl/gsl_assert
Farwaykorse: Clang (v7.0.0) sees this as returning `terminate_handler` and throws the warning `-Winvalid-noreturn` here.

I don't understand this workaround for the STL's exceptions _(top of `namespace details` ~line 100)_.
But, if this was not intended (and it is not a bug in clang's implementation), than it could be changed there.

If you want to maintain the `[[noreturn]]` or the issue is with Clang, than I could disable the warning.
I chose not do this, since I couldn't figure out the validity of the warning. Considering that when a code or compiler update warrants adding `[[noreturn]]` again, a new warning will show up.

tests/CMakeLists.txt
Farwaykorse: _A case of "works on my machine"-reversed._
Neither, those were needed to compile with `/Weverything`.

When compiling with CMake in Visual Studio these warnings where show. Not on AppVeyor.
They should not unless `/Weverything` is active, but that flag is not set, (`-v` on ninja)
... Deleting the CMakeCache solved it. Bug in VS's CMake?

Now I have to check if any other change was unnecessary ...
On the bright-side, disabling 6 warnings is enough to compile with every warning in clang enabled.

tests/CMakeLists.txt
Farwaykorse: It was intentional. Clang-cl expects the flags from cl.exe, but anything not in MSVC, like the warnings (besides the sets `-Wall` and `-Wextra`) requires the 'normal' clang flags.

I'll make it a dual configuration, MSVC-like and GNUC-like.
(Like clang, Intel C++ also ships in two such platform-dependent versions.)
With a generator expression per compiler for specific flags. Some duplication, but it'll limit the if-statement nesting to two levels.

tests/span_tests.cpp
```diff
@@ -54,9 +55,18 @@ struct DerivedClass : BaseClass
 };
 struct AddressOverloaded
 {
+#if (__cplusplus > 201402L)
+    [[maybe_unused]]
+#elif defined(__clang__)
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wunused-member-function"
```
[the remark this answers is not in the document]
Farwaykorse: After finding the [local issue](#discussion_r237531050), it turns out this warning `-Wunused-member-function` is only part of the `-Weverything` set.

The warning exists only with Clang, C++14 and if specifically enabled or with `-Weverything`.
I'll remove the pragma's, if the configuration is changed to use `-Weverything` it can be disabled for C++14.
I'll leave the `[[maybe_unused]]`. As discussed with CaseyCarter, it does state the purpose of the function. And is future-proof for improved diagnostics.

include/gsl/gsl_assert
annagrin: I suppose this is because clang_cl gives a warning on invalind noreturn. This `terminate()` should never return, and we shouldn't have difference in API based on whether we use exceptions or not. Can we instead detect that we are compiling with clang and use `__attribute((noreturn))` in the function pointer typedef? msvc does not have a problem with this code, and gcc does not run on it.

`typedef void (__attribute((noreturn))  *func )();`

annagrin:
@Farwaykorse realized that I didn't answer your questions, my apologies.

> Q: Is it acceptable to use the latest CMake with clang-cl? Or should the first working version be used?

We should use the first working version, to make sure this works for al many people as possible.

> Replacing MSBuild with Ninja on AppVeyor.

This is great test, but we shoud add it if possible, not replace - we are actually testing that GSL is working with Visual Studio and msbuild.

> Q: Do you agree with the buildsystem switch?

As I mentioned above, this is a great addtion but we don't want to remove the test of msbuild

include/gsl/gsl_assert
annagrin: Is clan_cl using MS STL? if yes, it has the same problem of std::terminate not terminating as msvc does, hence the workaround:)

tests/CMakeLists.txt
annagrin: Looks good!

tests/CMakeLists.txt
annagrin: > ... Deleting the CMakeCache solved it. Bug in VS's CMake?

Possible, do you think you have a repro, or repro steps? You can use "clone GSL" as a repro step. I can send it to the CMake team.

annagrin: > On the bright-side, improvements have been made, and disabling 7 warnings is enough to compile with /WX and every warning in clang enabled.

Great! As for the unnecessary disabling, we should re-enable them if the don't fire. To be sure we do not  miss warnings in the future.

include/gsl/gsl_assert
Farwaykorse: Yes, clang-cl uses the MS STL.

tests/CMakeLists.txt
Farwaykorse: > repro, or repro steps?

I'll give it a try, when I have some time.
> As for the unnecessary disabling, we should re-enable them if they don't fire. To be sure we do not miss warnings in the future.

Done, pushed the warning-level with clang-cl to -Weverything.
Nothing special needed, just adding a `#if defined(__GNUC__)` in a few places to solve `-Wundef`.

Farwaykorse:
Changes:
- The warning level with LLVM/clang is now pushed to `-Weverything`.
  - No issues with clang-cl on Windows.
  - Compilation with clang++ on Linux and OSX.
    - `-Wdeprecated` solved, but please review.
    - `-Wpadded` currently suppressed, but might be solvable.
    - `-Wweak-vtables` currently suppressed, but might be solvable.
- The AppVeyor script now supports both MSBuild and Ninja-build.
  The current setup uses MSBuild by default and Ninja for two cases:
  - VS2017: There is a bug in the LLVM integration extension on AppVeyor.
  - VS2015: MSBuild is not supported, since LLVM v7.0.0.
- Add clang-cl compilation with VS2015 (using the v140 MS STL).

include/gsl/gsl_assert
Farwaykorse: It took me a while to get to some understanding.
And indeed, inconsistent use of `[[noreturn]]` in an interface would be an issue.

Considering using `__attribute__((noreturn))`:
It becomes part of the type `terminate_handler` and a requirement for every instantiation of the type and any custom terminate handler. Since the attribute is not the same as `[[noreturn]]` this creates a separate interface just for clang on Windows.
See: Farwaykorse/GSL commit: 9a171ce14a68c157c6db834fbcd038678eafda96

To me, this looks like an even worse situation.

Suppressing the warning locally seems the best option.
There seem to be no usage or compilation issues, when ignoring the warning, and it is limited to the workaround.
For end-users the noreturn interface properties should be maintained, and without warnings.

If I'm wrong here and there would be a better solution, please let me know. Currently I don't see it.

annagrin:
@Farwaykorse I will resolve conflicts caused by my changes. Thanks @Farwaykorse!

annagrin:
added PR #762 to resolve conflicts

Farwaykorse:
Thanks @annagrin and @CaseyCarter for for your friendly and constructive feedback.

PR #753, opened 2018-11-28
Use the c++17 constexpr if in CheckRange if possible

kile0:


MikeGitb:
How about checking for `__cpp_if_constexpr`? I haven't checked if it is supported by all relevant computers though.

annagrin:
Thanks @kile0 !

Issue #754, opened 2018-12-04
span operator[] not constexpr due to CheckRange


alexhgreen07:
The span element access `operator[]` cannot be used in a `constexpr` due to the `CheckRange` not being marked as `constexpr`. Instead the developer is forced to use iterators (which may not be appropriate sometimes). Making `CheckRange` `constexpr` should allow for compile time checking of ranges assuming  a compatible contract policy.

PR #755, opened 2018-12-04
Make CheckRange constexpr

alexhgreen07:
Added support for `constexpr` range checking to allow for `constexpr span::operator[]` to work in a constant expression.


alexhgreen07:
#754

annagrin:
Thanks @alexhgreen07 !

PR #759, opened 2019-01-07
Fixes CMAKE_VERSION check

### LINKED ISSUE #758: Wrong CMake version comparison. No natvis included
https://github.com/Microsoft/GSL/blob/0f68d133fa6fd2973951b8aaab481e34bbfd2cf4/CMakeLists.txt#L70

The check for the cmake version is wrong (-> always false), this prevents the inclusion of gsl natvis in a visual studio solution.

The check should be:
if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.7.9)

k-brac:
Fixes #758 by using the correct comparison operator for a version
See documentation at https://cmake.org/cmake/help/latest/command/if.html
Thanks to this change, gsl natvis is properly included in a visual studio solution

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,7 +67,7 @@ else()
 endif()
 
 
-if ((CMAKE_VERSION GREATER 3.7.9) OR (CMAKE_VERSION EQUAL 3.7.9))
+if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.7.9)
     if (MSVC_IDE)
         option(VS_ADD_NATIVE_VISUALIZERS "Configure project to use Visual Studio native visualizers" TRUE)
     else()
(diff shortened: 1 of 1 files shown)


johelegp:
This would require bumping
https://github.com/Microsoft/GSL/blob/1c8310480df612342d9e7ff6802b9356a034101c/CMakeLists.txt#L1
to 3.7.0, see https://cmake.org/cmake/help/latest/release/3.7.html#commands.

k-brac:
you are right. I have been fooled by the use of VERSION_LESS. I didn't check if VERSION_GREATER_EQUAL was added in the same version of cmake.
I changed the comparison to work with cmake < 3.7

annagrin:
thanks @k-brac !

PR #764, opened 2019-01-16
Fixed check if std::byte is available

beinhaerter:
The check for std::byte existance fails on gcc 8.2. According to https://en.cppreference.com/w/cpp/utility/feature_test it is necessary to `#include <cstddef>` before checking `__cpp_lib_byte`.

### DIFF
--- a/include/gsl/gsl_byte
+++ b/include/gsl/gsl_byte
@@ -56,6 +56,7 @@
 #else // _MSC_VER
 
 #ifndef GSL_USE_STD_BYTE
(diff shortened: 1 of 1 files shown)


beinhaerter:
Three months without reaction on such a simple PR. Is contribution still appreciated? Or is the project dead?

annagrin:
Thanks @beinhaerter, apologies for the delay. We are figuring out resourcing for the project, should improve shortly. @JordanMaples  is joining the support of GSL.

PR #766, opened 2019-01-31
* adding cancel to final_action

jchorst:
There are scenarios where it would be useful to allow a final_action to be cancelled. The mechanism already exists to support moves -- this patch exposes it with a new public method 'cancel'.


JordanMaples:
Maintainer's call: linking to isocpp/cppcoreguidelines#688

JordanMaples:
Maintainers' call: The Core Guidelines still hasn't added this, please reopen this PR whenever cancellation is added to final_action. Thanks!

Issue #765, opened 2019-01-19
Using difference_type


jack17529:
In line - https://github.com/Microsoft/GSL/blob/master/include/gsl/span#L144 , one can just use "difference_type" instead of "typename Span::index_type" declared before, right?

PR #769, opened 2019-02-20
Update usage of difference_type in span

jack17529:
About my Issue - https://github.com/Microsoft/GSL/issues/765

### DIFF
--- a/include/gsl/span
+++ b/include/gsl/span
@@ -141,7 +141,7 @@ namespace details
 
         span_iterator() = default;
 
-        constexpr span_iterator(const Span* span, typename Span::index_type idx) noexcept
+        constexpr span_iterator(const Span* span, difference_type idx) noexcept
             : span_(span), index_(idx)
         {}
 


RedDwarf69:
I have not looked at it in detail. I was just looking for an update on index_type, but I don't think this PR would be OK in the future since `index_type` and `difference_type` are different types (see https://en.cppreference.com/w/cpp/container/span or https://github.com/martinmoene/span-lite#select-index-type)

jack17529:
@RedDwarf69 no, you have connected the wrong dots.
@annagrin can we please get this PR merged, one other PR by me is also remaining to get merged.

annagrin:
@JordanMaples is this boiling down to gsl::span not following the standard? if yes, lets have a bigger PR making it so.

JordanMaples:
Maintainer's call: Looks good, thank you.

RedDwarf69:
What have I got wrong here:

This constructor is going to be called by span::end(). span::end() is going to use span::size() as the "idx" parameter of this constructor. span::size() returns a `span::index_type`. After this PR "idx" is now of type `span_iterator::difference_type`.
As it is right now `span_iterator::difference_type` == `span::index_type` == "the type returned by span::size()" == "the type used by span::end()", no issues.

But whenever GSL's span is made to follow the standard somebody is going to create `span::difference_type` and will redefine `span_iterator::difference_type` as `span::difference_type`. Now if I have a span of size SIZE_MAX, and SIZE_MAX > PTRDIFF_MAX, span::end() is going to fail (and would have not before this PR).

?

PR #784, opened 2019-05-09
Update "Make library importable by CMake using find_package"

hanst99:
Exporting CMake targets directly is very useful for a variety of build systems. There was [this PR](https://github.com/microsoft/GSL/pull/708) to implement this, but there hasn't been any movement on it for a couple of months so I took it over (there was only one merge conflict to fix on it, I believe all the comments in the original PR had already been addressed by its author).

I don't know if this is the 'normal' way to deal with a situation like this but I'd hate for this to go to waste as it'd make GSL much easier to use.


hanst99:
w.r.t. to CLA, not sure if I can sign it considering pretty much all of this PR is based on work submitted by someone else (again, I only did the final merge and I'm willing to address further comments if there are any), so I'm not sure if this meets the "originality" clause here. I think considering the original work was submitted under signing the CLA that means I can use it, and I'm tentatively agreeing based on that, but I'll withdraw the submission if that isn't the case.

pr0g:
Hey there,

I wound up implementing this myself recently for a library I wanted to use the GSL with.

I've created a fork here:

> https://github.com/pr0g/GSL

and the diff looks like this:

> https://github.com/microsoft/GSL/compare/master...pr0g:cmake-find-package.

It's pretty similar to this change - I'm using the fork in another project right now.

I know this PR #784 already exists so I don't want to step on anyone's toes but I would love if support could officially be added. I'd be happy to create a PR with my fork/branch if that would help or try and update this one?

Thanks!

Tom

Adnn:
Hi,

Is there something we can do to get this PR in a state it will be merged?

pr0g:
I got impatient and made a fork I could use in my project 😝 https://github.com/pr0g/GSL/tree/cmake-find-package?files=1

JordanMaples:
Thanks for this PR. Sorry about the wait on getting it reviewed/merged. 

Would you mind adding an update to the readme's "Using the libraries" section? We need to have instructions for the end user on how they need to modify their `CMakeLists.txt` files to take advantage of this.

Adnn:
Thank you for your feedback @JordanMaples.

JordanMaples:
@ras0219-msft do you have any thoughts about the changes to CMake proposed here?

CMakeLists.txt
```diff
@@ -54,15 +57,13 @@ target_compile_definitions(GSL INTERFACE
...
+        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+        $<INSTALL_INTERFACE:include>
     )
 else()
     target_include_directories(GSL SYSTEM INTERFACE
-        $<BUILD_INTERFACE:
-            ${CMAKE_CURRENT_SOURCE_DIR}/include
-        >
+        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+        $<INSTALL_INTERFACE:include>
```
ras0219-msft: This should be `${CMAKE_INSTALL_INCLUDEDIR}` to match the command below.

CMakeLists.txt
```diff
@@ -54,15 +57,13 @@ target_compile_definitions(GSL INTERFACE
 # the SYSTEM keyword suppresses warnings for users of the library
 if(GSL_STANDALONE_PROJECT)
     target_include_directories(GSL INTERFACE
-        $<BUILD_INTERFACE:
-            ${CMAKE_CURRENT_SOURCE_DIR}/include
-        >
+        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+        $<INSTALL_INTERFACE:include>
```
ras0219-msft: This should be `${CMAKE_INSTALL_INCLUDEDIR}` to match the command below.

CMakeLists.txt
```diff
@@ -77,15 +78,21 @@ if (CMAKE_VERSION VERSION_GREATER 3.7.8)
...
     if(VS_ADD_NATIVE_VISUALIZERS)
         target_sources(GSL INTERFACE
-            ${CMAKE_CURRENT_SOURCE_DIR}/GSL.natvis
+            $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/GSL.natvis>
         )
     endif()
 endif()
 
+install(TARGETS GSL EXPORT Microsoft.GSLConfig
+  PUBLIC_HEADER DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
```
ras0219-msft: Because you're using `$<INSTALL_INTERFACE:...>` above and `install(DIRECTORY include/gsl ...)` below, you shouldn't use `PUBLIC_HEADER` here (it does nothing).

CMakeLists.txt
```diff
@@ -77,15 +78,21 @@ if (CMAKE_VERSION VERSION_GREATER 3.7.8)
...
+)
 install(
     DIRECTORY include/gsl
-    DESTINATION include
+    DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
 )
+# Make library importable by other projects
+install(EXPORT Microsoft.GSLConfig NAMESPACE Microsoft.GSL:: DESTINATION share/Microsoft.GSL/cmake)
+export(TARGETS GSL NAMESPACE Microsoft.GSL:: FILE Microsoft.GSLConfig.cmake)
```
ras0219-msft: Adding an alias target after the `install/export` commands would be useful to make things "symmetric" across all three possible use modes (`add_subdirectory()`, `find_package()` into the build dir, `find_package()` into the install dir).
```cmake
add_library(Microsoft.GSL::GSL ALIAS GSL)
```

Adnn:
Thank you @ras0219-msft for taking time to review!

> I'd also recommend that the `target_compile_features()` and `target_compile_options()` are wrapped in `$<BUILD_INTERFACE:...>` expressions -- there is not a CMake-wide standard practice around whether these kinds of settings should be exported to consumers

Unless I am missing something here, BUILD_INTERFACE generator expression would not prevent propagation to consumers: it would export the value only for the build tree (as opposed to the install tree). It is true that projects are most commonly consumed from their install tree, yet it is perfectly possible to consume GSL from its install tree, it which case you'd have different behaviours depending on where the project is consumed from, which imo is worst.

On the other hand, both CMake functions you mention can be given `PRIVATE` argument, which would effectively prevent propagation to downstream (consumers).

> and I personally find it alarming that simply
> ```cmake
> find_package(Microsoft.GSL)
> target_link_libraries(main PRIVATE Microsoft.GSL::GSL)
> ```
> 
> will suddenly upgrade my project to C++14.

This is a good point. Yet, it is possible to argue that if the `INTERFACE` requirement of the project actually is C++14 (i.e., its public headers will not compile withou C++14), then it is a good thing that this build requirement is implicitly forwarded. As otherwise, a consuming project which is not upgraded will not build anyway.

hanst99:
@ras0219-msft @Adnn I added some commits which I hope should address (most) of your comments. I’ve not done anything about the C++14 `INTERFACE` requirement, I lean towards agreeing with @Adnn here that it’s unlikely to be harmful if GSL requires C++14 anyway (assuming that is indeed the case).

If you think this is a blocker I can also remove it though.

Also sorry for taking so long to get back to this, I didn’t realise there was activity here.

hanst99:
Also a side note, maybe there should a CI test for making sure the installation bit actually works? I tested it locally, but it’d probably be better to have something there to avoid breaking it...

Adnn:
Thank you @hanst99 for your changes, as far as I can see they cover every point raised here (except the debated requirement propagation, as you already stated).

@JordanMaples Do you think it is possible to merge this PR in? To give a few arguments:

* this change enables a new way to consume the project via a CMake package, but does not have any impact on the already existing ways to consume it, so disruption to existing consumers is not expected
* those changes do not set anything in stone regarding the CMake package, and we can address further potential questions/required changes via separate PRs.
* in particular, it will allow to produce a Conan recipe PR, which I plan to start when this PR is merged in ; )

Adnn:
@JordanMaples PRs #811 and #820 seem to implement the same feature as this PR. I'd say this proves that users really need this feature to use this library, as so many try to implement it.

This PR is likely the most mature and reviewed of the set, do you think it can be merged in so we are done with it?

JordanMaples:
I think this one's ready to go in. Thank you all for your work on this.

PR #791, opened 2019-06-09
Update span

jack17529:
Removed size_type.


JordanMaples:
Maintainer's call: looks good, however we need to check if user code relies on `size_type` before accepting.

JordanMaples:
@CaseyCarter @BjarneStroustrup 

Is it safe to remove size_type without breaking too much user code? Are you ok with doing this?
Note: std::span does not have size_type.

JordanMaples:
Maintainer's call: Revising earlier comment. This is now outdated following #844.

PR #800, opened 2019-08-05
Fixes #798 : Dropping inclusion of `gsl/pointers` in `string_span`

### LINKED ISSUE #798: string_view header includes gsl/pointers, doesn't seem to need it
In the [`string_view` header](https://raw.githubusercontent.com/microsoft/GSL/master/include/gsl/string_span), we have the following line:
```
#include <gsl/pointers>   // for not_null
```
The thing is, `string_view` doesn't use `not_null` anywhere. So, is it really needed in there, or can it be dropped?

eyalroz:
This PR fixes issue #798  by  dropping inclusion of `gsl/pointers` in `string_span`.

### DIFF
--- a/include/gsl/string_span
+++ b/include/gsl/string_span
@@ -20,7 +20,6 @@
 #include <gsl/gsl_assert> // for Ensures, Expects
 #include <gsl/gsl_util>   // for narrow_cast
 #include <gsl/span>       // for operator!=, operator==, dynamic_extent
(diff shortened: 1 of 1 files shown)


eyalroz:
FYI - the unsuccessful check is due to a VM update command failing, not an actual test failing.

JordanMaples:
Maintainer's call: Looks great, thanks.

PR #803, opened 2019-08-31
Fixed memory leaks and out of bound access

### LINKED ISSUE #737: Invalid use of pointer in notnull_tests.cpp TEST_CASE("TestNotNullostream")
Detected with the help of valgrind which shows strlen accessing invalid memory.  In the excerpt below, p is a pointer to a single char, but the ostream::operator()<< does not know this and assume it's a pointer to char, and will end up calling strlen via _Traits::length(__s)). Turns out there is a null terminator in memory, 29 bytes after the 'c', so no crash and the test passes. I solved this by casting to void * p and &v.
```
TEST_CASE("TestNotNullostream")
{
...
    ostream_helper<char>('c');
...
 }

void ostream_helper(T v)
{
    not_null<T*> p(&v);
    {
        std::ostringstream os;
        std::ostringstream ref;
        os << p;   ///// HEREHERE
        ref << &v; ///// HEREHERE
        CHECK(os.str() == ref.str());
    }
}
```
For reference here is the code to operator<<
```

template<class _Traits>
    inline basic_ostream<char, _Traits>&
    operator<<(basic_ostream<char, _Traits>& __out, const char* __s)
    {
      if (!__s)
	__out.setstate(ios_base::badbit);
      else
	__ostream_insert(__out, __s,
			 static_cast<streamsize>(_Traits::length(__s))); ///// HEREHERE
      return __out;
    }


```

B1Z0N:
I looked through all tests with `valgrind --leak-check=full --leak-kinds=all` and found some memory errors. So this PR is just minor fixes of ones.

--------
Fixes #737. Actually this issue gave me an idea to run valgrind on all files. So thanks to @stayprivates. It's been over a year since this issue is on hold, so i don't know why he/she doesn't PR the solution already.

### DIFF
--- a/tests/multi_span_tests.cpp
+++ b/tests/multi_span_tests.cpp
@@ -1220,6 +1220,8 @@ TEST_CASE("md_access")
             expected += 3;
         }
     }
+
+    delete[] image_ptr;
 }
 
 GSL_SUPPRESS(con.4) // NO-FORMAT: attribute
@@ -1616,6 +1618,8 @@ TEST_CASE("span_structure_size")
     multi_span<const double, dynamic_range, 6, 4> av2 =
         as_multi_span(av1, dim(5), dim<6>(), dim<4>());
     (void) av2;
+
+    delete[] arr;
 }
 
 GSL_SUPPRESS(con.4) // NO-FORMAT: attribute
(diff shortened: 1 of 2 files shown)


B1Z0N:
@annagrin @JordanMaples any thoughts on this?

JordanMaples:
Maintainers' call: Looks good, thanks.

stayprivates:
Sorry for not participating more, at the time git/github was new to me and I did not have the time to get acquinted with it. Then I forgot about it.

PR #805, opened 2019-09-05
Add nodiscard to finally

### LINKED ISSUE #657: add nodiscard to finally
add nodiscard to finally

B1Z0N:
Fixes #657

### DIFF
--- a/include/gsl/gsl_util
+++ b/include/gsl/gsl_util
@@ -43,10 +43,13 @@
 #define GSL_CONSTEXPR_NARROW 1
 #endif
 
-#if defined(__cplusplus) && (__cplusplus >= 201703L)
-#define GSL_NODISCARD [[nodiscard]]
-#else
-#define GSL_NODISCARD
+#ifdef __has_cpp_attribute
+#  if __has_cpp_attribute(nodiscard)
+#    define GSL_NODISCARD [[nodiscard]]
+#  endif
+#endif
+#ifndef GSL_NODISCARD
+#    define GSL_NODISCARD
 #endif
 
 namespace gsl
--- a/include/gsl/gsl_util
+++ b/include/gsl/gsl_util
@@ -44,14 +44,10 @@
 #endif
 
 #if defined(__cplusplus) && (__cplusplus >= 201703L)
-
 #define GSL_NODISCARD [[nodiscard]]
-
-#else // defined(__cplusplus) && (__cplusplus >= 201703L)
-
+#else
 #define GSL_NODISCARD
-
-#endif // defined(__cplusplus) && (__cplusplus >= 201703L)
+#endif
 
 namespace gsl
 {
--- a/include/gsl/gsl_util
+++ b/include/gsl/gsl_util
@@ -43,15 +43,6 @@
 #define GSL_CONSTEXPR_NARROW 1
 #endif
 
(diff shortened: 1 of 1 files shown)


include/gsl/gsl_util
```diff
@@ -43,6 +43,12 @@
 #define GSL_CONSTEXPR_NARROW 1
 #endif
 
+#if defined(__cplusplus) && (__cplusplus >= 201703L)
```
johelegp: This should have the form
```C++
#ifdef __has_cpp_attribute                      // Check if __has_cpp_attribute is present
#  if __has_cpp_attribute(deprecated)           // Check for an attribute
#    define DEPRECATED(msg) [[deprecated(msg)]]
#  endif
#endif
#ifndef DEPRECATED
#    define DEPRECATED(msg)
#endif
```
See https://en.cppreference.com/w/cpp/feature_test.

omartijn: I wonder if it's even necessary. GSL targets c++14 and up. Attributes were standardized in c++11. Wouldn't it make sense to use [[nodiscard]] directly?

johelegp: Neither form is foolproof. The current check isn't enough to suggest the existence of `[[nodiscard]]` and not all compilers with `__cplusplus >= 201703L` implement the relaxation of allowing non-existent attributes in case they haven't implemented `nodiscard`. I'd guess some users that use `-Werror` would get a hard error as such. My suggestion itself requires a C++20 feature, so I can't say it's better. `nodiscard` itself is a C++17 addition, as noted in the first table of the link.

B1Z0N: I think `__has_cpp_attribute` is the best we can do for now. I am glad that this feature is being added in c++20. But at the same time I hate that it took so long to realize the need.

B1Z0N:
Any thoughts on how i could fix this?
My suggestion is to add this after previous checks:
```c++

#if defined(__clang__) && defined(__llvm__)
#define GSL_NODISCARD __attribute__((warn_unused_result))
#endif

```

I am a bit new to multicompiler support, so i'd be happy if somebody could give me some advices.

JordanMaples:
While I'm not opposed to the feature test, every compiler we've tested with C++14 has no problem compiling the `[[nodiscard]]`. Can you give an example of a compiler that has a problem with the code as it is?

omartijn:
@B1Z0N I don't really get what you're trying to achieve with that last example. Clang is an example of a compiler which definitely doesn't have a problem with `[[nodiscard]]`. Why would you write it so that this specific compiler uses a non-standard attribute?

I'm with @JordanMaples on this. As said before, just using `[[nodiscard]]` directly seems like the best option here, unless you have a concrete example of a compiler that doesn't support it (while supporting c++14 in other respects).

B1Z0N:
1. Now I see that visual studio with `c++14` doesn't support `[[nodiscard]]`, that is logical, since it was added in `c++17`.  Should I test for version of the cpp standard to enable it in `c++14`?

2. By the way I don't know any other way of testing the piece of code on all major compilers, except the usual, manual way. What do you do in such cases?

JordanMaples:
Maintainers' call: Closing this as nodiscard was added in #908 using a macro approach.

PR #807, opened 2019-09-15
Add a table of features

kuzkry:
Hello!
I believe it would compose nicely with https://github.com/isocpp/CppCoreGuidelines/pull/1516.
And please help me with `strided_span` cause I have no idea what that beast might be.

PR #808, opened 2019-09-15
Implement joining_thread

kuzkry:
Note that this is rebased on top of #807 but it consists of one commit only (therefore I suggest going to the commit section and simplifying your review this way).

This PR implements a simple version of joining_thread. Appropriate thread library is detected and supplied by CMake.
Reasons why joining_thread derives privately from std::thread:
* basically private inheritance works like composition with different semantics
* it allows to get bunch of usings instead of redefining the whole std::thread interface
* no problems with undefined behaviour if a user would try to delete joining_thread through the base class pointer
* prevention of doing tricks, such as:
```c++
    joining_thread a([]{});
    a.detach() // compile error, detach() is deleted
    a.std::thread::detach(); // no error, though a loophole
````
Last but not least, I didn't care much about exceptions thrown by joinable(), I assumed it would have to be an incredibly crazy scenario for this function to throw, but noexcept(false) could *theoretically* be added there. I say "theoretically", because the same function is called by the destructor of std::thread anyway which would terminate in this case.


kuzkry:
WIP: Needs some improvement + possibly adding the cancellation mode.

PR #813, opened 2019-10-04
Deprecate multi_span, strided_span, etc

JordanMaples:
A number of features exist in the GSL that are not a part of the Core Guidelines. 

This PR marks all classes and structs within the multi_span header as deprecated.


nitronoid:
This is sad. 
The multi-span implementation provided here is the best one I've found and suits all my needs out of the box.
Would you please consider supporting it separately instead of completely abandoning the idea?

JordanMaples:
Maintainers' call: Approving this. If you require this, I recommend creating a copy for yourself as we will no longer be maintaining this portion going forward.

JordanMaples:
@nitronoid It looks correct to me, however I'm not an expert with licensing. 
@gdr-at-ms or @hsutter would either of you mind looking at @nitronoid's licensing on the multi_span mirror (https://github.com/nitronoid/multi_span)?

PR #814, opened 2019-10-06
Fix typos in gsl/pointers

sizmailov:


JordanMaples:
Maintainers' call: Thanks for the corrections.

PR #815, opened 2019-10-07
Added std::exchange (C++14) to make the intent more explicit

matt77hias:


JordanMaples:
Maintainers' call: thanks.

Issue #816, opened 2019-10-13
MSVC min/max macro collision with numeric_limits


andreduvoisin:
MSVC with windows.h included and without NOMINMAX defined causes a min/max macro collision with numeric_limits in multi_span. For projects with warnings as errors enabled, this breaks the build.

MSVC 14.2 Warning:

```
C:\path\to\GSL\include\gsl\multi_span(86): warning C4003: not enough arguments for function-like macro invocation 'max'
```

Clang Diagnostic:

```
Error while processing RedactedFile.h.
path\to\GSL\include\gsl/multi_span:86:78: error: too few arguments provided to function-like macro invocation [clang-diagnostic-error]
        static const SizeType max_value = std::numeric_limits<SizeType>::max();
                                                                             ^
C:\Program Files (x86)\Windows Kits\10\include\10.0.17763.0\shared\minwindef.h:193:9: note: macro 'max' defined here
#define max(a,b)            (((a) > (b)) ? (a) : (b))
        ^
```

andreduvoisin:
PR submitted with fix. #817

SanderBouwhuis:
So, is this now fixed? I discovered GSL yesterday and was interested to use it.
I ran into the same issue.

andreduvoisin:
Not yet. The PR is submitted and all tests pass — it just needs to be reviewed by a maintainer and merged. Not sure why it hasn’t been looked at yet, but other PRs submitted after this one also haven’t gained any traction.

Maybe @JordanMaples or @annagrin could help out?

SanderBouwhuis:
Thanks Andre. I simply did the changes you suggested and it seems to work.

Very strange that Microsoft hasn't even tested this in their own compiler.

JordanMaples:
What is the aversion to using NOMINMAX? The problem lies with presence of the min and max macro definitions. I'm hesitant to take this fix because it's not an issue with the implementation of the GSL itself. 

To clarify the lack of traction, generally I only address GSL related issues on Wednesdays, but I was out of the office last week.

Thank you for your patience.

JordanMaples:
@SanderBouwhuis 

GSL itself should be able to compile out of the box to be used in your applications without any sort of an additional overhead. Is it only when you're trying to use it in conjunction with windows.h that you're having issues or are you running into issues in general?

[...]

PR #817, opened 2019-10-13
Fix MSVC min/max macro collision with numeric_limits.

andreduvoisin:
Related issue: https://github.com/microsoft/GSL/issues/816

Wraps all usages of `std::numeric_limits<T>::max` and `std::numeric_limits<T>::min` in parenthesis to prevent function-style macro expansion.


JordanMaples:
Maintainers' call: GSL is made to work with Standard C++ and we encourage conventional function call syntax. `min` and `max` macros are non-standard so you should use NOMINMAX.

Issue #818, opened 2019-10-13
throw_exception() is not appropriately defined for GSL_UNENFORCED_ON_CONTRACT_VIOLATION


alexkaratarakis:
According to the `gsl_assert` header:
```
// There are three configuration options for this GSL implementation's behavior
// when pre/post conditions on the GSL types are violated:
//
// 1. GSL_TERMINATE_ON_CONTRACT_VIOLATION: std::terminate will be called (default)
// 2. GSL_THROW_ON_CONTRACT_VIOLATION: a gsl::fail_fast exception will be thrown
// 3. GSL_UNENFORCED_ON_CONTRACT_VIOLATION: nothing happens
```

When `GSL_UNENFORCED_ON_CONTRACT_VIOLATION` is selected, nothing
should happen, but throw_exception()'s behavior is not consistent
with this.

`throw_exception()` is notably used in `narrow()`, which in turn is used by the `gsl::span(Container)` constructor.

JordanMaples:
Maintainer's call: The Core Guidelines editors have given guidance in the following issue
isocpp/CppCoreGuidelines#1512 that contract violations should not throw, they should terminate.

We need to remove the throwing options entirely, not just when they are unenforced.

JordanMaples:
Maintainer's call: Closing issue as contract violation now only results in termination.

PR #819, opened 2019-10-13
Fix throw_exception() for GSL_UNENFORCED_ON_CONTRACT_VIOLATION

alexkaratarakis:
There are three configuration options for when conditions are violated.
When GSL_UNENFORCED_ON_CONTRACT_VIOLATION is selected, nothing
should happen, but throw_exception()'s behavior is not consistent
with this.

The ifdefs for the three config options (throw, terminate, unenforced)
have been reordered to be consistent with the order they appear in
the documentaton at the start of the header as well as the definition
of GSL_CONTRACT_CHECK right below these changes

#818

 Properly define noexcept specification for narrow() depending on the selected contract violation configuration

#818

### DIFF
--- a/include/gsl/gsl_assert
+++ b/include/gsl/gsl_assert
@@ -128,15 +128,7 @@ namespace details
 #endif
     }
 
-#if defined(GSL_TERMINATE_ON_CONTRACT_VIOLATION)
-
-    template <typename Exception>
-    [[noreturn]] void throw_exception(Exception&&) noexcept
-    {
-        gsl::details::terminate();
-    }
-
-#else
+#if defined(GSL_THROW_ON_CONTRACT_VIOLATION)
 
     template <typename Exception>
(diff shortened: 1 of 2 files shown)


JordanMaples:
Maintainer's call: The Core Guidelines editors have given guidance in the following issue
isocpp/CppCoreGuidelines#1512 that contract violations should not throw, they should terminate.

We need to remove the throwing options entirely, not just when they are unenforced.

johelegp:
I find that decision unfortunate. The throwing behaviour's pretty great when used in the tests of a library. I understand the conservative decision of the guideline's editors considering what happened to contracts, but that doesn't mean you shouldn't offer this alternative behaviour.

JordanMaples:
Maintainer's call: Thank you for the PR. Closing as GSL_UNENFORCED_ON_CONTRACT_VIOLATION no longer exists. Contract violation now always results in termination.

PR #823, opened 2019-11-08
Add missing #include if built with GSL_USE_STD_BYTE=1

RedDwarf69:
cstddef was being included to detect whether GSL_USE_STD_BYTE should be
set to 0 or 1. But if the user provided GSL_USE_STD_BYTE no detection is
required and the header was never included.


ombre5733:
:+1: We have the same patch in place to use the GSL together with `gcc-arm-none-eabi-7.2.1`.

PR #824, opened 2019-11-13
Expects should force terminate on contract violation for non-test code

JordanMaples:
This change makes GSL_TERMINATE_ON_CONTRACT_VIOLATION the default behavior for non-test environments while also removing GSL_UNENFORCED_ON_CONTRACT_VIOLATION. 

Throwing can still be enabled by defining GSL_TEST, this is to convey to the developer that this is not the desired behavior in production.

### DIFF
--- a/include/gsl/gsl_assert
+++ b/include/gsl/gsl_assert
@@ -51,16 +51,31 @@
 #endif
 
 //
-// There are three configuration options for this GSL implementation's behavior
-// when pre/post conditions on the GSL types are violated:
-//
-// 1. GSL_TERMINATE_ON_CONTRACT_VIOLATION: std::terminate will be called (default)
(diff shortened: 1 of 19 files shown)


JordanMaples:
Addressing @hsutter's comment from isocpp/CppCoreGuidelines#1512 clarifying that the intent of [GSL.assert](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#gslassert-assertions) is to terminate on contract violation.

JordanMaples:
After discussing this with the C++ Core Guidelines maintainers, the test logic should always terminate, regardless of being in test mode.
The correct way to test the contract itself is to have one contract per test execution and ensuring that the terminate handler is invoked to log contract violation.

Issue #810, opened 2019-09-19
Is there a reason there are no free-standing begin() and end() functions for gsl::span?


omartijn:
I usually use free-standing `begin()` and `end()` for getting iterators on my containers. This has worked fine for me, since it uses ADL to find the function in the relevant namespace, while allowing me to override it locally.

Now these functions aren't defined in the GSL namespace. Is there a specific reason for not doing this? If not, I could easily whip up a PR.

Note about standardization, when `std::span` is standardized it would obviously work with that, since then the span is defined in the std namespace, which has `begin()` and `end()` functions, so adding these functions now will make it behave more like `std::span`.

mbeutel:
Don't you do the two-step dance anyway when writing generic code?

```c++
template <typename R>
    auto myAlgorithm(R&& range)
{
    using std::begin;
    using std::end;
    auto e = end(range);
    for (auto i = begin(range); i != e; ++i) { ... }
}
```

This makes your algorithm work for plain arrays as well as for `gsl::span<>`.

In C++20 you can simplify this by using `ranges::begin()`/`end()`.

omartijn:
@mbeutel I _could_ do that, indeed. I don't normally do that, though. It'd be nice if this would just work with gsl::span.

mbeutel:
I disagree with the idea that custom containers should provide custom freestanding `begin()`, `end()`, `size()`, `ssize()` functions.

What is the benefit of writing `begin(c)` over `c.begin()` in your scenario?

JordanMaples:
Maintainers' call: You're right, we should provide begin, end, size, and ssize functions as non-members.

PR #825, opened 2019-11-14
Implement free-standing (c|r|cr)(begin|end) and ssize functions

omartijn:
This issue implements the free-standing iterator and size function discussed in https://github.com/microsoft/GSL/issues/810.

Closes https://github.com/microsoft/GSL/issues/810

### DIFF
--- a/include/gsl/span
+++ b/include/gsl/span
@@ -774,6 +774,62 @@ constexpr ElementType& at(span<ElementType, Extent> s, index i)
     return s[i];
 }
 
+// [span.obs] Free observer functions
+template <class ElementType, std::ptrdiff_t Extent>
+constexpr typename span<ElementType, Extent>::index_type ssize(const span<ElementType, Extent> &span) noexcept
+{
+    return span.size();
+}
+
+// [span.iter] Free functions for begin/end functions
+template <class ElementType, std::ptrdiff_t Extent>
+constexpr typename span<ElementType, Extent>::iterator begin(const span<ElementType, Extent> &span) noexcept
+{
+    return span.begin();
+}
+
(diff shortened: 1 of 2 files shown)


JordanMaples:
Thanks for working on this @omartijn. Would you mind adding some tests to span_tests.cpp for the new functions?

I'm also getting the same failures in my build so I think something might be wrong with AppVeyor. I'll start looking into AppVeyor to get everything unblocked.

omartijn:
I have added tests for the new functions, @JordanMaples. Note that I check the return type of the free-standing functions and compare them with the member functions. Since this is all compile-time known, I could have just as well done this with a static_assert instead of a run-time check. Is there any preference known for this?

A static_assert could also be added in the implementation file, instead of in the tests - which is the way I personally prefer doing things.

omartijn:
I've rebased to the current master in the hopes of getting this merged. I am still getting AppVeyor errors originating in tests/bounds_tests.cpp - which appear unrelated to my changes. I think this has to do with the recent switch to gtest.

@JordanMaples do you know more about this? What can I do to get this merged?

JordanMaples:
Thanks for the ping. You're right, the failures are because of the recent GTest merge. It's strange that these warnings didn't show up in the tests that were run before the merge. I'll make a quick fix, get that pushed through, and will let you know as soon as it's all clean.

JordanMaples:
I just pushed a change that should fix the build issues. I'll check in an hour or so to see if Appveyor is clean. If it's not I'll continue pushing fixes.

JordanMaples:
Maintainer's call: Looks good, thanks!

PR #834, opened 2020-01-11
Travis fixes/upgrades

JordanMaples:
Changes to the Travis matrix:

Upgrading the distro from Trusty to Bionic because Trusty is EOL.
Upgrading XCode to the latest versions from XCode 9, 10, and 11
Removing Clang 3.6 - 3.9
Adding Clang 8 and 9.

Changes to the test files:
The diagnostic inconsistent-missing-destructor-override is missing in Clang 4 and earlier. I added major version detection to the warning suppression to only add this flag to versions that support it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,101 +22,42 @@ matrix:
     # so leave this first for the overall build to be faster
     ##########################################################################
 
-    # XCode 8.3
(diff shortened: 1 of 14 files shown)


JordanMaples:
abandoning this in favor of the changes made by @Farwaykorse in #835

PR #835, opened 2020-01-14
Improve CI with TravisCI

### LINKED ISSUE #760: CI uses a cmake version greater than 3.1.3
From the feedback I got on #759 I noticed that CI is using a cmake version greater than the minimum recommended one. So CI won't automatically catch incompatible use of cmake

Farwaykorse:
- Extend CI reach
  - [x] Add compilers:
    -  GCC 8 and GCC 9
    - LLVM/Clang 8, 9 and 10
    - AppleClang / Xcode 9.4, 10.1 and 11.3 (different compilers)
  - [x] Build both C++14 and C++17 with all capable compilers.
  - [x] Add validation of the CMake configuration with all supported versions. Fixes #760
    Generate with `-Wdev -Werror=dev --warn-uninitialized`
  -  ~~Libc++ with Clang?~~ (already covered by AppleClang)
- Improve Job performance
  - [x] Use the default CMake versions on all images. This significantly improves set-up times.
  - [x] Change default Linux image to Ubuntu Bionic Beaver. (Use Ubuntu Xenial where needed for older compilers.)
    Get Clang form the Ubuntu packages where available.
- Other
  - [x] Use readable names for builds
  - [x] Add verification of variables (Bash) + fix setting path
  - [x] Use "stages" to end CI after jobs with the latest compilers have failed.
    The latest compilers have the best error messages. There is no benefit in running all legacy compilers before the error is fixed.
    Same for ending the CI run after a CMake error is found.
  - [x] Clang 5.0.1 and AppleClang 9.1 report (false positive?) warnings for `-Wundefined-func-template`. In tests and only in C++17 mode.
  - [x] Increase CMake requirement for AppleClang to CMake 3.2.3.
    GoogleTest uses feature detection for C++11, AppleClang features information is added in [CMake 3.2](https://cmake.org/cmake/help/latest/release/3.2.html#other).
  - [x] Failure with GCC-8

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -261,6 +261,41 @@ matrix:
     - name: GCC-7 C++17 Release
       env: CXX=g++-7 BUILD_TYPE=Release GSL_CXX_STANDARD=17
 
+    # GCC 8
+    - name: GCC-8 C++14 Debug
+      env: CXX=g++-8 BUILD_TYPE=Debug GSL_CXX_STANDARD=14
+      addons: &gcc8
+        apt:
+          packages: g++-8
+    - name: GCC-8 C++14 Release
+      env: CXX=g++-8 BUILD_TYPE=Release GSL_CXX_STANDARD=14
+      addons: *gcc8
+    - name: GCC-8 C++17 Debug
+      env: CXX=g++-8 BUILD_TYPE=Debug GSL_CXX_STANDARD=17
+      addons: *gcc8
+    - name: GCC-8 C++17 Release
+      env: CXX=g++-8 BUILD_TYPE=Release GSL_CXX_STANDARD=17
+      addons: *gcc8
+
+    # GCC 9
+    - name: GCC-9 C++14 Debug
+      env: CXX=g++-9 BUILD_TYPE=Debug GSL_CXX_STANDARD=14
+      addons: &gcc9
+        apt:
+          sources:
+          - sourceline: ppa:ubuntu-toolchain-r/test
+          packages:
+          - g++-9
+    - name: GCC-9 C++14 Release
+      env: CXX=g++-9 BUILD_TYPE=Release GSL_CXX_STANDARD=14
+      addons: *gcc9
+    - name: GCC-9 C++17 Debug
+      env: CXX=g++-9 BUILD_TYPE=Debug GSL_CXX_STANDARD=17
+      addons: *gcc9
+    - name: GCC-9 C++17 Release
+      env: CXX=g++-9 BUILD_TYPE=Release GSL_CXX_STANDARD=17
+      addons: *gcc9
+
 install:
   - ${CXX} --version
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,7 +44,8 @@ matrix:
     - name: CMake validation on OSX
       os: osx
       osx_image: xcode11.3
-      env: *CMAKE_VERSION_LIST
+      env:
+      - CMAKE_VERSION: '"3.16.2 3.15.6 3.14.7 3.13.5 3.12.4 3.11.4 3.10.3 3.9.6 3.8.2 3.7.2 3.6.3 3.5.2 3.4.3 3.3.2 3.2.3"'
       script:
       - |
         cd ./build
--- a/.travis.yml
+++ b/.travis.yml
@@ -298,7 +298,6 @@ matrix:
 
     # Clang 9
     - name: Clang-9 C++14 Debug
-      stage: Latest
       env: CXX=clang++-9 BUILD_TYPE=Debug GSL_CXX_STANDARD=14
       addons: &clang9
         apt:
@@ -316,6 +315,27 @@ matrix:
     - name: Clang-9 C++17 Release
       env: CXX=clang++-9 BUILD_TYPE=Release GSL_CXX_STANDARD=17
       addons: *clang9
+
+    # Clang 10
+    - name: Clang-10 C++14 Debug
+      stage: Latest
+      env: CXX=clang++-10 BUILD_TYPE=Debug GSL_CXX_STANDARD=14
+      addons: &clang10
+        apt:
+          sources:
+          - sourceline: 'deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic-10 main'
+            key_url: https://apt.llvm.org/llvm-snapshot.gpg.key
+          packages:
+          - clang-10
+    - name: Clang-10 C++14 Release
+      env: CXX=clang++-10 BUILD_TYPE=Release GSL_CXX_STANDARD=14
+      addons: *clang10
+    - name: Clang-10 C++17 Debug
+      env: CXX=clang++-10 BUILD_TYPE=Debug GSL_CXX_STANDARD=17
+      addons: *clang10
+    - name: Clang-10 C++17 Release
+      env: CXX=clang++-10 BUILD_TYPE=Release GSL_CXX_STANDARD=17
+      addons: *clang10
 
     ##########################################################################
     # GCC on Linux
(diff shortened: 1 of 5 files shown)


JordanMaples:
Thanks for writing this up. I have some thoughts on changes that can be made:

- For Clang versions <= 4.0, You'll run into issues where the build will fail because the warning suppression for inconsistent-missing-destructor-override doesn't exist in those versions. To mitigate that consider changing each of the test files that include the suppression to have the clang version check. 
```cpp
#if __clang_major__ > 4
#pragma GCC diagnostic ignored "-Winconsistent-missing-destructor-override"
#endif // __clang_major__ > 4
```
- You can remove Clang versions older than 4.
- For XCode, It's sufficient to only have 9.4, 10.3, and 11.3 in the matrix. 

I didn't look too closely at the g++-8 failures.

Farwaykorse:
@JordanMaples 

> For Clang versions <= 4.0 ...

Curious. In my working/ experimental branch, I removed the suppressing of `-Winconsistent-missing-destructor-override` from the files and none of the builds triggered the warning.
Are you sure this warning needs to be suppressed?
[Latest test run on Travis](https://travis-ci.com/Farwaykorse/GSL/builds/144411070)

> ... Clang versions older than 4 ... only have 9.4, ...

Shouldn't the earliest compatible compilers be in the CI?

> ... g++-8 failures.

I'm seeing an warning specific to GCC-8, Release builds, both C++14 and 17.
https://travis-ci.com/Farwaykorse/GSL/jobs/275589611#L500

I'll create a separate PR about moving the warning suppression for the tests to CMake.

Farwaykorse:
Rebased onto master and removed Xcode-11.1.
[TravisCI](https://travis-ci.org/microsoft/GSL/builds/638046156)
Build failures with older LLVM/Clang and AppleClang versions are solved in #836 by removing the relevant warning suppression pragma.

GCC-8 issue still needs to be looked at.

JordanMaples:
I tried looking into the GCC-8 issue and I'm not sure what's going on. I'm unable to locally reproduce the issue with gcc-8.3 on either WSL Ubuntu or WSL Kali. Godbolt has no problem compiling the code either, so I think it may be an issue with Travis. I can test natively on my home pc later on tonight, but I doubt the results will be different.

Do you have any thoughts?

Farwaykorse:
I'm seeing the same error when building on the WSL (Ubuntu 18.04.3 LTS) with gcc 8.3.0 installed from the Ubuntu devel package repository (g++-8 (Ubuntu 8.3.0-6ubuntu1~18.04.1) 8.3.0).

The error only shows for Release configurations.

JordanMaples:
Sorry that it's been a month since I got back to you on this. Is the only bug in multi_span? If it is, I'm tempted to just remove the multi_span_test from the build matrix because as it is the code is deprecated.

Farwaykorse:
It is, and only appears with this specific compiler configuration. Even including the file from a different location does not trigger it. (Probably the reason it won't reproduce on Godbolt.)

.travis.yml
```diff
@@ -7,262 +7,494 @@ notifications:
...
+        cd ./build
+        ( set -eu
+          for CMAKE in ${CMAKE_path[@]}; do test_CMake_generate $CMAKE; done
+        )
+
+    ##########################################################################
+    # AppleClang on OSX
     ##########################################################################
 
     # XCode 8.3
```
kainjow:
```suggestion
    # Xcode 8.3
```

.travis.yml
```diff
@@ -7,262 +7,494 @@ notifications:
...
-    - env: COMPILER=clang++ BUILD_TYPE=Release GSL_CXX_STANDARD=14
+    - name: AppleClang Xcode-8.3 C++14 Release
+      env: BUILD_TYPE=Release GSL_CXX_STANDARD=14
       os: osx
       osx_image: xcode8.3
-      compiler: clang
 
-    # XCode 9.1
-    - env: COMPILER=clang++ BUILD_TYPE=Debug GSL_CXX_STANDARD=14
+    # XCode 9.0 earliest C++17 support
```
kainjow:
```suggestion
    # Xcode 9.0 earliest C++17 support
```

.travis.yml
```diff
@@ -7,262 +7,494 @@ notifications:
...
+    - name: AppleClang Xcode-9.0 C++14 Debug
+      env: BUILD_TYPE=Debug GSL_CXX_STANDARD=14
+      os: osx
+      osx_image: xcode9
+    - name: AppleClang Xcode-9.0 C++14 Release
+      env: BUILD_TYPE=Release GSL_CXX_STANDARD=14
+      os: osx
+      osx_image: xcode9
+
+    # XCode 9.4
```
kainjow:
```suggestion
    # Xcode 9.4
```

.travis.yml
```diff
@@ -7,262 +7,494 @@ notifications:
...
+      env: BUILD_TYPE=Debug GSL_CXX_STANDARD=17
+      os: osx
+      osx_image: xcode9.4
+    - name: AppleClang Xcode-9.4 C++17 Release
+      env: BUILD_TYPE=Release GSL_CXX_STANDARD=17
+      os: osx
+      osx_image: xcode9.4
 
-    - env: COMPILER=clang++ BUILD_TYPE=Release GSL_CXX_STANDARD=14
+    # XCode 10.1
```
kainjow:
```suggestion
    # Xcode 10.1
```

.travis.yml
```diff
@@ -7,262 +7,494 @@ notifications:
...
+      os: osx
+      osx_image: xcode10.1
+    - name: AppleClang Xcode-10.1 C++17 Release
+      env: BUILD_TYPE=Release GSL_CXX_STANDARD=17
       os: osx
-      osx_image: xcode9.1
-      compiler: clang
+      osx_image: xcode10.1
+
+    # XCode 10.3
```
kainjow:
```suggestion
    # Xcode 10.3
```

.travis.yml
```diff
@@ -7,262 +7,494 @@ notifications:
...
+    - name: AppleClang Xcode-10.3 C++17 Debug
+      env: BUILD_TYPE=Debug GSL_CXX_STANDARD=17
+      os: osx
+      osx_image: xcode10.3
+    - name: AppleClang Xcode-10.3 C++17 Release
+      env: BUILD_TYPE=Release GSL_CXX_STANDARD=17
+      os: osx
+      osx_image: xcode10.3
+
+    # XCode 11.3
```
kainjow:
```suggestion
    # Xcode 11.3
```

Farwaykorse:
Thanks for looking after this PR @JordanMaples 

The GCC-8 error for multi-span still persists after merging the latest master.
I've tried replacing the GTest macro "EXPECT_DEATH" with "EXPECT_DEBUG_DEATH" on  [multi_span_tests.cpp#L1045](https://github.com/microsoft/GSL/blob/d90fefea6de716fdf5e90fe3cfd8281a704e3be2/tests/multi_span_tests.cpp#L1045) but it still triggers in release builds.
Since `multi-span` is deprecated and probably won't see a fix, I've disabled the line with a macro.

All builds on TravisCI should now pass.

JordanMaples:
Fantastic work! I'll merge these as soon as Appveyor finishes. Thanks again for all of your work @Farwaykorse!

JordanMaples:
@Farwaykorse it looks like MSVC is failing in the latest changes in multi_span_tests. I'm guessing the problem is the 'not' keyword. I'm not sure if the msvc preprocessor supports it. Would you mind taking a look at it? If I don't hear back, I'll pull this in the next 3 or so hours and experiment locally.

Farwaykorse:
@JordanMaples Yes, thank you for bringing it to my attention.
````
warning C4067: unexpected tokens following preprocessor directive
````
I believe the alternative tokens where added in VS2017, apparently not for the C++14 mode.

Issue #782, opened 2019-05-01
Make gsl::span::index_type std::size_t


RedDwarf69:
Currently gsl::span::index_type is std::ptrdiff_t, but C++20 will use std::size_t (https://en.cppreference.com/w/cpp/container/span) and span-lite is planing to do the move (https://github.com/martinmoene/span-lite#select-index-type).

anordal:
Ditto for `size()`: #559

JordanMaples:
See isocpp/CppCoreGuidelines#1511

hsutter:
Guidelines editors' call: Yes, let's change it to unsigned. We'll update the Guidelines.

natersoz:
What do you do with `rend()` which needs to index to -1 ?

anordal:
> rend()

I would turn that question on its head and ask when -1 needs to be < 0.

JordanMaples:
Maintainer's call: Closing this gsl::span is now unsigned.

Issue #559, opened 2017-10-08
gsl::span::size returns signed integer


Lectem:
Is there any rationale behind this ? Right now it returns an `index_type` which is `ptr_diff_t`,a signed type.
It is quite inconsistent with the rest of the standard library where `size()` returns unsigned integers (usually `size_t`), for example `std::vector`.

I know that many in the committee see the use of unsigned as an error of conception, but still I'd prefer it being consistent with the standard library.

gdr-at-ms:
Yes, there is a rationale behind this, and it has been explained numerous times.  Please, use the search facility.

Lectem:
Sorry, should have looked for it before. While I understand why it was chosen, I still think it's a pain.
Yes the standard library was wrong, but isn't the GSL made to work with it ?
If I can not use both together easily, I think it doesn't really make sense to use the GSL.

mapron:
So, if current issue is now reopened, is it possible to reopen gsl::index issues? Not trolling.
I'm pretty annoyed I not able to use gsl::index examples because VC++ complaints for signed/unsigned comparison. And no, I could not just turn off that warning in our codebase.

JordanMaples:
Maintainer's call: gsl::span is now unsigned following #844, so gsl::span::size returns unsigned values. For those who still require signed values, please use gsl::span::ssize.

PR #844, opened 2020-02-03
gsl::span: Migrate from signed to unsigned

JordanMaples:
Migrate gsl::span and gsl::string_span's index_type to be unsigned instead of signed. This aligns the gsl::span's implementation to the std.

#559, #782, isocpp/CppCoreGuidelines#1511

### DIFF
--- a/include/gsl/span
+++ b/include/gsl/span
@@ -199,7 +199,8 @@ namespace details
         constexpr span_iterator operator+(const difference_type n) const noexcept
         {
             span_iterator ret = *this;
-            return ret += n;
+            ret += n;
+            return ret;
         }
 
         friend constexpr span_iterator operator+(const difference_type n,
@@ -220,7 +221,8 @@ namespace details
         constexpr span_iterator operator-(const difference_type n) const noexcept
         {
             span_iterator ret = *this;
-            return ret -= n;
+            ret -= n;
+            return ret;
         }
 
         template <
@@ -342,8 +344,6 @@ namespace details
     {
     public:
         using size_type = std::size_t;
-
-        static_assert(Ext != dynamic_extent, "A fixed-size span must not have size == dynamic_extent");
 
         constexpr extent_type() noexcept = default;
 
--- a/include/gsl/span
+++ b/include/gsl/span
@@ -552,16 +552,6 @@ public:
         return data()[size() - 1];
     }
 
-    // at and operator() are deprecated to align to the public member functions of std::span
-    [[deprecated("Use operator[]")]] constexpr reference at(size_type idx) const noexcept
-    {
-        return this->operator[](idx);
-    }
-    [[deprecated("Use operator[]")]] constexpr reference operator()(size_type idx) const noexcept
-    {
-        return this->operator[](idx);
-    }
-
     constexpr pointer data() const noexcept { return storage_.data(); }
 
     // [span.iter], span iterator support
@@ -817,67 +807,67 @@ constexpr ElementType& at(span<ElementType, Extent> s, index i)
 
 // [span.obs] Free observer functions
 template <class ElementType, std::size_t Extent>
-constexpr typename span<ElementType, Extent>::size_type
-ssize(const span<ElementType, Extent>& span) noexcept
+constexpr std::ptrdiff_t
+ssize(const span<ElementType, Extent>& s) noexcept
 {
-    return span.size();
+    return static_cast<std::ptrdiff_t>(s.size());
 }
 
 // [span.iter] Free functions for begin/end functions
 template <class ElementType, std::size_t Extent>
 constexpr typename span<ElementType, Extent>::iterator
-begin(const span<ElementType, Extent>& span) noexcept
+begin(const span<ElementType, Extent>& s) noexcept
 {
-    return span.begin();
+    return s.begin();
 }
 
 template <class ElementType, std::size_t Extent = dynamic_extent>
 constexpr typename span<ElementType, Extent>::iterator
-end(const span<ElementType, Extent>& span) noexcept
+end(const span<ElementType, Extent>& s) noexcept
 {
-    return span.end();
+    return s.end();
 }
 
 template <class ElementType, std::size_t Extent>
 constexpr typename span<ElementType, Extent>::const_iterator
-cbegin(const span<ElementType, Extent>& span) noexcept
+cbegin(const span<ElementType, Extent>& s) noexcept
 {
-    return span.cbegin();
+    return s.cbegin();
 }
 
 template <class ElementType, std::size_t Extent>
 constexpr typename span<ElementType, Extent>::const_iterator
-cend(const span<ElementType, Extent>& span) noexcept
+cend(const span<ElementType, Extent>& s) noexcept
 {
-    return span.cend();
+    return s.cend();
 }
 
 template <class ElementType, std::size_t Extent>
 constexpr typename span<ElementType, Extent>::reverse_iterator
-rbegin(const span<ElementType, Extent>& span) noexcept
+rbegin(const span<ElementType, Extent>& s) noexcept
 {
-    return span.rbegin();
+    return s.rbegin();
 }
 
 template <class ElementType, std::size_t Extent>
 constexpr typename span<ElementType, Extent>::reverse_iterator
-rend(const span<ElementType, Extent>& span) noexcept
+rend(const span<ElementType, Extent>& s) noexcept
 {
-    return span.rend();
+    return s.rend();
 }
 
 template <class ElementType, std::size_t Extent>
 constexpr typename span<ElementType, Extent>::const_reverse_iterator
-crbegin(const span<ElementType, Extent>& span) noexcept
+crbegin(const span<ElementType, Extent>& s) noexcept
 {
-    return span.crbegin();
+    return s.crbegin();
 }
 
 template <class ElementType, std::size_t Extent>
 constexpr typename span<ElementType, Extent>::const_reverse_iterator
-crend(const span<ElementType, Extent>& span) noexcept
+crend(const span<ElementType, Extent>& s) noexcept
 {
-    return span.crend();
+    return s.crend();
 }
 
 } // namespace gsl
--- a/include/gsl/span
+++ b/include/gsl/span
@@ -296,7 +296,7 @@ namespace details
         constexpr void _Verify_offset(const difference_type n) const noexcept
         { // test that *this + n is within the span of this iterator STL
             // algorithm call
-            Expects((current_ + n) >= begin_ && (current_ + n) <= end_);
+            Expects( n <= (current_ - begin_)  && n <= (end_ - current_));
         }
 
         GSL_SUPPRESS(bounds.1) // NO-FORMAT: attribute
@@ -331,13 +331,13 @@ namespace details
     class extent_type
     {
     public:
-        using index_type = std::size_t;
+        using size_type = std::size_t;
 
         static_assert(Ext != dynamic_extent, "A fixed-size span must be >= 0 in size.");
 
         constexpr extent_type() noexcept {}
 
-        template <index_type Other>
+        template <size_type Other>
         constexpr extent_type(extent_type<Other> ext)
         {
             static_assert(Other == Ext || Other == dynamic_extent,
@@ -345,30 +345,30 @@ namespace details
             Expects(ext.size() == Ext);
         }
 
-        constexpr extent_type(index_type size) { Expects(size == Ext); }
+        constexpr extent_type(size_type size) { Expects(size == Ext); }
 
-        constexpr index_type size() const noexcept { return Ext; }
+        constexpr size_type size() const noexcept { return Ext; }
     };
 
     template <>
     class extent_type<dynamic_extent>
     {
     public:
-        using index_type = std::size_t;
+        using size_type = std::size_t;
 
-        template <index_type Other>
+        template <size_type Other>
         explicit constexpr extent_type(extent_type<Other> ext) : size_(ext.size())
         {}
 
-        explicit constexpr extent_type(index_type size) : size_(size)
+        explicit constexpr extent_type(size_type size) : size_(size)
         {
             Expects(size != dynamic_extent);
         }
 
-        constexpr index_type size() const noexcept { return size_; }
+        constexpr size_type size() const noexcept { return size_; }
 
     private:
-        index_type size_;
+        size_type size_;
     };
 
     template <class ElementType, std::size_t Extent, std::size_t Offset, std::size_t Count>
@@ -388,7 +388,7 @@ public:
     // constants and types
     using element_type = ElementType;
     using value_type = std::remove_cv_t<ElementType>;
-    using index_type = std::size_t;
+    using size_type = std::size_t;
     using pointer = element_type*;
     using reference = element_type&;
     using difference_type = std::ptrdiff_t;
@@ -398,12 +398,12 @@ public:
     using reverse_iterator = std::reverse_iterator<iterator>;
     using const_reverse_iterator = std::reverse_iterator<const_iterator>;
 
-    using size_type = index_type;
+    using index_type [[deprecated("use size_type instead of index_type")]] = size_type;
 
 #if defined(GSL_USE_STATIC_CONSTEXPR_WORKAROUND)
-    static constexpr const index_type extent{Extent};
+    static constexpr const size_type extent{Extent};
 #else
-    static constexpr index_type extent{Extent};
+    static constexpr size_type extent{Extent};
 #endif
 
     // [span.cons], span constructors, copy, assignment, and destructor
@@ -414,7 +414,7 @@ public:
     constexpr span() noexcept : storage_(nullptr, details::extent_type<0>())
     {}
 
-    constexpr span(pointer ptr, index_type count) noexcept : storage_(ptr, count) {}
+    constexpr span(pointer ptr, size_type count) noexcept : storage_(ptr, count) {}
 
     constexpr span(pointer firstElem, pointer lastElem) noexcept
         : storage_(firstElem, static_cast<std::size_t>(std::distance(firstElem, lastElem)))
@@ -451,7 +451,7 @@ public:
                   std::is_convertible<typename Container::pointer, pointer>::value &&
                   std::is_convertible<typename Container::pointer,
                                       decltype(std::declval<Container>().data())>::value>>
-    constexpr span(Container& cont) noexcept : span(cont.data(), narrow<index_type>(cont.size()))
+    constexpr span(Container& cont) noexcept : span(cont.data(), narrow<size_type>(cont.size()))
     {}
 
     template <class Container,
@@ -461,7 +461,7 @@ public:
                   std::is_convertible<typename Container::pointer,
                                       decltype(std::declval<Container>().data())>::value>>
     constexpr span(const Container& cont) noexcept
-        : span(cont.data(), narrow<index_type>(cont.size()))
+        : span(cont.data(), narrow<size_type>(cont.size()))
     {}
 
     constexpr span(const span& other) noexcept = default;
@@ -504,36 +504,36 @@ public:
         return {data() + Offset, Count == dynamic_extent ? size() - Offset : Count};
     }
 
-    constexpr span<element_type, dynamic_extent> first(index_type count) const noexcept
+    constexpr span<element_type, dynamic_extent> first(size_type count) const noexcept
     {
         Expects(count <= size());
         return {data(), count};
     }
 
-    constexpr span<element_type, dynamic_extent> last(index_type count) const noexcept
+    constexpr span<element_type, dynamic_extent> last(size_type count) const noexcept
     {
         Expects(count <= size());
         return make_subspan(size() - count, dynamic_extent, subspan_selector<Extent>{});
     }
 
-    constexpr span<element_type, dynamic_extent> subspan(index_type offset,
-                                                         index_type count = dynamic_extent) const
+    constexpr span<element_type, dynamic_extent> subspan(size_type offset,
+                                                         size_type count = dynamic_extent) const
         noexcept
     {
         return make_subspan(offset, count, subspan_selector<Extent>{});
     }
 
     // [span.obs], span observers
-    constexpr index_type size() const noexcept { return storage_.size(); }
-    constexpr index_type size_bytes() const noexcept
+    constexpr size_type size() const noexcept { return storage_.size(); }
+    constexpr size_type size_bytes() const noexcept
     {
-        return size() * narrow_cast<index_type>(sizeof(element_type));
+        return size() * narrow_cast<size_type>(sizeof(element_type));
     }
     constexpr bool empty() const noexcept { return size() == 0; }
 
     // [span.elem], span element access
     GSL_SUPPRESS(bounds.1) // NO-FORMAT: attribute
-    constexpr reference operator[](index_type idx) const noexcept
+    constexpr reference operator[](size_type idx) const noexcept
     {
         Expects(idx < size());
         return data()[idx];
@@ -552,11 +552,11 @@ public:
     }
 
     // at and operator() are deprecated to align to the public member functions of std::span
-    [[deprecated("Use operator[]")]] constexpr reference at(index_type idx) const noexcept
+    [[deprecated("Use operator[]")]] constexpr reference at(size_type idx) const noexcept
     {
         return this->operator[](idx);
     }
-    [[deprecated("Use operator[]")]] constexpr reference operator()(index_type idx) const noexcept
+    [[deprecated("Use operator[]")]] constexpr reference operator()(size_type idx) const noexcept
     {
         return this->operator[](idx);
     }
@@ -564,13 +564,30 @@ public:
     constexpr pointer data() const noexcept { return storage_.data(); }
 
     // [span.iter], span iterator support
-    constexpr iterator begin() const noexcept { return {data(), data() + size(), data()}; }
-    constexpr iterator end() const noexcept { return {data(), data() + size(), data() + size()}; }
+    constexpr iterator begin() const noexcept
+    {
+        const auto data = storage_.data();
+        return {data, data + size(), data};
+    }
 
-    constexpr const_iterator cbegin() const noexcept { return {data(), data() + size(), data()}; }
+    constexpr iterator end() const noexcept
+    {
+        const auto data = storage_.data();
+        const auto size = storage_.size();
+        return {data, data + size, data + size};
+    }
+
+    constexpr const_iterator cbegin() const noexcept
+    {
+        const auto data = storage_.data();
+        return {data, data + size(), data};
+    }
+
     constexpr const_iterator cend() const noexcept
     {
-        return {data(), data() + size(), data() + size()};
+        const auto data = storage_.data();
+        const auto size = storage_.size();
+        return {data, data + size, data + size};
     }
 
     constexpr reverse_iterator rbegin() const noexcept { return reverse_iterator{end()}; }
@@ -635,7 +652,7 @@ private:
 
     // The rest is needed to remove unnecessary null check
     // in subspans and constructors from arrays
-    constexpr span(KnownNotNull ptr, index_type count) : storage_(ptr, count) {}
+    constexpr span(KnownNotNull ptr, size_type count) : storage_(ptr, count) {}
 
     template <std::size_t CallerExtent>
     class subspan_selector
@@ -643,7 +660,7 @@ private:
     };
 
     template <std::size_t CallerExtent>
-    span<element_type, dynamic_extent> make_subspan(index_type offset, index_type count,
+    span<element_type, dynamic_extent> make_subspan(size_type offset, size_type count,
                                                     subspan_selector<CallerExtent>) const
     {
         const span<element_type, dynamic_extent> tmp(*this);
@@ -651,7 +668,7 @@ private:
     }
 
     GSL_SUPPRESS(bounds.1) // NO-FORMAT: attribute
-    span<element_type, dynamic_extent> make_subspan(index_type offset, index_type count,
+    span<element_type, dynamic_extent> make_subspan(size_type offset, size_type count,
                                                     subspan_selector<dynamic_extent>) const
     {
         Expects(size() >= offset);
@@ -665,7 +682,7 @@ private:
 
 #if defined(GSL_USE_STATIC_CONSTEXPR_WORKAROUND)
 template <class ElementType, std::size_t Extent>
-constexpr const typename span<ElementType, Extent>::index_type span<ElementType, Extent>::extent;
+constexpr const typename span<ElementType, Extent>::size_type span<ElementType, Extent>::extent;
 #endif
 
 // [span.comparison], span comparison operators
@@ -748,7 +765,7 @@ as_writeable_bytes(span<ElementType, Extent> s) noexcept
 //
 template <class ElementType>
 constexpr span<ElementType> make_span(ElementType* ptr,
-                                      typename span<ElementType>::index_type count)
+                                      typename span<ElementType>::size_type count)
 {
     return span<ElementType>(ptr, count);
 }
@@ -799,7 +816,7 @@ constexpr ElementType& at(span<ElementType, Extent> s, index i)
 
 // [span.obs] Free observer functions
 template <class ElementType, std::size_t Extent>
-constexpr typename span<ElementType, Extent>::index_type
+constexpr typename span<ElementType, Extent>::size_type
 ssize(const span<ElementType, Extent>& span) noexcept
 {
     return span.size();
--- a/include/gsl/span
+++ b/include/gsl/span
@@ -71,11 +71,7 @@ namespace gsl
 {
 
 // [views.constants], constants
-#if (defined(GSL_USE_STATIC_CONSTEXPR_WORKAROUND))
 constexpr const std::size_t dynamic_extent = static_cast<std::size_t>(-1);
-#else
-constexpr std::size_t dynamic_extent = static_cast<std::size_t>(-1);
-#endif
 
 template <class ElementType, std::size_t Extent = dynamic_extent>
 class span;
@@ -137,7 +133,7 @@ namespace details
 
 #ifdef _MSC_VER
         using _Unchecked_type = pointer;
-#endif
+#endif // _MSC_VER
         constexpr span_iterator() = default;
 
         constexpr span_iterator(pointer begin, pointer end, pointer current)
@@ -348,14 +344,14 @@ namespace details
     public:
         using size_type = std::size_t;
 
-        static_assert(Ext != dynamic_extent, "A fixed-size span must be >= 0 in size.");
+        static_assert(Ext != dynamic_extent, "A fixed-size span must not have size == dynamic_extent");
 
         constexpr extent_type() noexcept {}
 
         template <size_type Other>
         constexpr extent_type(extent_type<Other> ext)
         {
-            static_assert(Other == Ext || Other == dynamic_extent,
+            static_assert(Other == Ext,
                           "Mismatch between fixed-size extent and size of initializing data.");
             Expects(ext.size() == Ext);
         }
@@ -414,8 +410,6 @@ public:
     using const_iterator = details::span_iterator<const ElementType>;
     using reverse_iterator = std::reverse_iterator<iterator>;
     using const_reverse_iterator = std::reverse_iterator<const_iterator>;
-
-    using index_type [[deprecated("use size_type instead of index_type")]] = size_type;
 
 #if defined(GSL_USE_STATIC_CONSTEXPR_WORKAROUND)
     static constexpr const size_type extent{Extent};
@@ -805,18 +799,6 @@ as_writable_bytes(span<ElementType, Extent> s) noexcept
     return {reinterpret_cast<byte*>(s.data()), s.size_bytes()};
 }
 
-template <class ElementType, std::size_t Extent,
-          std::enable_if_t<!std::is_const<ElementType>::value, int> = 0>
-[[deprecated(
-    "use as_writable_bytes")]] span<byte, details::calculate_byte_size<ElementType, Extent>::value>
-as_writeable_bytes(span<ElementType, Extent> s) noexcept
-{
-    // clang-format off
-    GSL_SUPPRESS(type.1) // NO-FORMAT: attribute
-    // clang-format on
-    return {reinterpret_cast<byte*>(s.data()), s.size_bytes()};
-}
-
 //
 // make_span() - Utility functions for creating spans
 //
@@ -868,7 +850,7 @@ constexpr ElementType& at(span<ElementType, Extent> s, index i)
 {
     // No bounds checking here because it is done in span::operator[] called below
     Ensures(i >= 0);
-    return s[narrow_cast<std::size_t>(i)];
+    return s[static_cast<std::size_t>(i)];
 }
 
 // [span.obs] Free observer functions
--- a/include/gsl/span
+++ b/include/gsl/span
@@ -418,7 +418,8 @@ public:
               // "Dependent" is needed to make "std::enable_if_t<Dependent || Extent == 0 || Extent
               // == dynamic_extent>" SFINAE, since "std::enable_if_t<Extent == 0 || Extent ==
               // dynamic_extent>" is ill-formed when Extent is greater than 0.
-              class = std::enable_if_t<(Dependent || Extent == 0 || Extent == dynamic_extent)>>
+              class = std::enable_if_t<(Dependent ||
+                                        details::is_allowed_extent_conversion<0, Extent>::value)>>
     constexpr span() noexcept : storage_(nullptr, details::extent_type<0>())
     {}
 
@@ -430,34 +431,29 @@ public:
     constexpr span(pointer firstElem, pointer lastElem) noexcept
         : storage_(firstElem, static_cast<std::size_t>(std::distance(firstElem, lastElem)))
     {
-        if (Extent != dynamic_extent) { Expects(lastElem - firstElem == static_cast<difference_type>(Extent)); }
+        if (Extent != dynamic_extent)
+        { Expects(lastElem - firstElem == static_cast<difference_type>(Extent)); }
     }
 
-    template <std::size_t N>
+    template <std::size_t N,
+              std::enable_if_t<details::is_allowed_extent_conversion<N, Extent>::value, int> = 0>
     constexpr span(element_type (&arr)[N]) noexcept
         : storage_(KnownNotNull{std::addressof(arr[0])}, details::extent_type<N>())
     {}
 
-    template <std::size_t N, std::size_t Ext = Extent,
-              std::enable_if_t<(N == Ext || Ext == dynamic_extent), int> = 0>
-    constexpr span(std::array<value_type, N>& arr) noexcept
+    template <std::size_t N,
+              std::enable_if_t<details::is_allowed_extent_conversion<N, Extent>::value, int> = 0>
+    constexpr span(std::array<element_type, N>& arr) noexcept
         : storage_(KnownNotNull{arr.data()}, details::extent_type<N>())
     {}
 
-    template <std::size_t N, std::size_t Ext = Extent,
-              std::enable_if_t<((N == Ext || Ext == dynamic_extent) &&
-                                std::is_convertible<const value_type*, element_type*>::value),
+    template <std::size_t N,
+              std::enable_if_t<(details::is_allowed_extent_conversion<N, Extent>::value &&
+                                details::is_allowed_element_type_conversion<const value_type,
+                                                                            element_type>::value),
                                int> = 0>
     constexpr span(const std::array<value_type, N>& arr) noexcept
         : storage_(KnownNotNull{arr.data()}, details::extent_type<N>())
-    {}
-
-    constexpr span(std::array<value_type, 0>&) noexcept
-        : storage_(static_cast<pointer>(nullptr), details::extent_type<0>())
-    {}
-
-    constexpr span(const std::array<value_type, 0>&) noexcept
-        : storage_(static_cast<pointer>(nullptr), details::extent_type<0>())
     {}
 
     // NB: the SFINAE here uses .data() as a incomplete/imperfect proxy for the requirement
@@ -474,6 +470,7 @@ public:
     template <class Container,
               class = std::enable_if_t<
                   std::is_const<element_type>::value && !details::is_span<Container>::value &&
+                  !details::is_std_array<Container>::value &&
                   std::is_convertible<typename Container::pointer, pointer>::value &&
                   std::is_convertible<decltype(std::declval<Container&>().data()), pointer>::value>>
     constexpr span(const Container& cont) noexcept : span(cont.data(), cont.size())
@@ -677,7 +674,7 @@ private:
 
     template <std::size_t CallerExtent>
     constexpr span<element_type, dynamic_extent> make_subspan(size_type offset, size_type count,
-                                                    subspan_selector<CallerExtent>) const
+                                                              subspan_selector<CallerExtent>) const
     {
         const span<element_type, dynamic_extent> tmp(*this);
         return tmp.subspan(offset, count);
@@ -686,8 +683,8 @@ private:
     // clang-format off
     GSL_SUPPRESS(bounds.1) // NO-FORMAT: attribute
     // clang-format on
-    constexpr span<element_type, dynamic_extent> make_subspan(size_type offset, size_type count,
-                                                    subspan_selector<dynamic_extent>) const
+    constexpr span<element_type, dynamic_extent>
+    make_subspan(size_type offset, size_type count, subspan_selector<dynamic_extent>) const
     {
         Expects(size() >= offset);
 
--- a/include/gsl/span
+++ b/include/gsl/span
@@ -271,7 +271,7 @@ namespace details
             std::enable_if_t<std::is_same<std::remove_cv_t<Type2>, value_type>::value, int> = 0>
         constexpr bool operator<=(const span_iterator<Type2>& rhs) const noexcept
         {
-            return !(*this > rhs);
+            return !(rhs < *this);
         }
 
         template <
@@ -279,7 +279,7 @@ namespace details
             std::enable_if_t<std::is_same<std::remove_cv_t<Type2>, value_type>::value, int> = 0>
         constexpr bool operator>=(const span_iterator<Type2>& rhs) const noexcept
         {
-            return *!(this < rhs);
+            return !(*this < rhs);
         }
 
 #ifdef _MSC_VER
--- a/include/gsl/span
+++ b/include/gsl/span
@@ -187,7 +187,7 @@ namespace details
         {
             Expects(begin_ && current_ && end_);
             if (n > 0) Expects(end_ - current_ >= n);
-            if (n < 0) Expects(end_ - current_ >= -n);
+            if (n < 0) Expects(current_ - begin_ >= -n);
             current_ += n;
             return *this;
         }
@@ -294,7 +294,7 @@ namespace details
         }
 
         constexpr void _Verify_offset(const difference_type n) const noexcept
-        { // test that the iterator *this + n is a valid range in an STL
+        { // test that *this + n is within the span of this iterator STL
             // algorithm call
             Expects((current_ + n) >= begin_ && (current_ + n) <= end_);
         }
--- a/include/gsl/span
+++ b/include/gsl/span
@@ -25,7 +25,7 @@
 #include <array>     // for array
 #include <cstddef>   // for ptrdiff_t, size_t, nullptr_t
 #include <iterator>  // for reverse_iterator, distance, random_access_...
-#include <memory> // for std::addressof
+#include <memory>    // for std::addressof
 #include <stdexcept>
 #include <type_traits> // for enable_if_t, declval, is_convertible, inte...
 #include <utility>
@@ -475,7 +475,7 @@ public:
         typename details::calculate_subspan_type<ElementType, Extent, Offset, Count>::type
     {
         Expects((size() >= Offset) &&
-                (Count == dynamic_extent || (Offset + Count <= size())));
+                (Count == dynamic_extent || (Count <= size() - Offset)));
 
         return {data() + Offset, Count == dynamic_extent ? size() - Offset : Count};
     }
--- a/include/gsl/span
+++ b/include/gsl/span
@@ -517,13 +517,13 @@ public:
 
     constexpr reference front() const noexcept
     {
-        Expects(size() != dynamic_extent && size() > 0);
+        Expects(size() > 0);
         return data()[0];
     }
 
     constexpr reference back() const noexcept
     {
-        Expects(size() != dynamic_extent && size() > 0);
+        Expects(size() > 0);
         return data()[size() - 1];
     }
 
--- a/include/gsl/span
+++ b/include/gsl/span
@@ -263,7 +263,7 @@ namespace details
             std::enable_if_t<std::is_same<std::remove_cv_t<Type2>, value_type>::value, int> = 0>
         constexpr bool operator>(const span_iterator<Type2>& rhs) const noexcept
         {
-            return !(*this < rhs);
+            return rhs < *this;
         }
 
         template <
--- a/include/gsl/span
+++ b/include/gsl/span
@@ -592,7 +592,7 @@ private:
         }
         else
         {
-            return idx >= 0 && idx < size;
+            return idx < size;
         }
     }
 
--- a/include/gsl/span
+++ b/include/gsl/span
@@ -333,7 +333,7 @@ namespace details
     public:
         using index_type = std::size_t;
 
-        static_assert(Ext >= 0, "A fixed-size span must be >= 0 in size.");
+        static_assert(Ext != dynamic_extent, "A fixed-size span must be >= 0 in size.");
 
         constexpr extent_type() noexcept {}
 
@@ -417,7 +417,7 @@ public:
     constexpr span(pointer ptr, index_type count) noexcept : storage_(ptr, count) {}
 
     constexpr span(pointer firstElem, pointer lastElem) noexcept
-        : storage_(firstElem, std::distance(firstElem, lastElem))
+        : storage_(firstElem, static_cast<std::size_t>(std::distance(firstElem, lastElem)))
     {}
 
     template <std::size_t N>
@@ -536,7 +536,7 @@ public:
     GSL_SUPPRESS(bounds.1) // NO-FORMAT: attribute
     constexpr reference operator[](index_type idx) const noexcept
     {
-        Expects(CheckRange(idx, storage_.size()));
+        Expects(idx < size());
         return data()[idx];
     }
 
@@ -594,32 +594,6 @@ public:
 #endif // _MSC_VER
 
 private:
-    static constexpr bool CheckRange(index_type idx, index_type size) noexcept
-    {
-        // Optimization:
-        //
-        // idx >= 0 && idx < size
-        // =>
-        // static_cast<size_t>(idx) < static_cast<size_t>(size)
-        //
-        // because size >=0 by span construction, and negative idx will
-        // wrap around to a value always greater than size when casted.
-
-        // check if we have enough space to wrap around
-#if defined(__cpp_if_constexpr)
-        if constexpr (sizeof(index_type) <= sizeof(size_t))
-#else
-        if (sizeof(index_type) <= sizeof(size_t))
-#endif
-        {
-            return narrow_cast<size_t>(idx) < narrow_cast<size_t>(size);
-        }
-        else
-        {
-            return idx < size;
-        }
-    }
-
     // Needed to remove unnecessary null check in subspans
     struct KnownNotNull
     {
--- a/include/gsl/span
+++ b/include/gsl/span
@@ -26,16 +26,18 @@
 #include <cstddef>   // for ptrdiff_t, size_t, nullptr_t
 #include <iterator>  // for reverse_iterator, distance, random_access_...
 #include <limits>
+#include <memory> // for std::addressof
 #include <stdexcept>
 #include <type_traits> // for enable_if_t, declval, is_convertible, inte...
 #include <utility>
-#include <memory> // for std::addressof
 
 #if defined(_MSC_VER) && !defined(__clang__)
 #pragma warning(push)
 
 // turn off some warnings that are noisy about our Expects statements
 #pragma warning(disable : 4127) // conditional expression is constant
+#pragma warning(                                                                                   \
+    disable : 4146) // unary minus operator applied to unsigned type, result still unsigned
 #pragma warning(disable : 4702) // unreachable code
 
 // Turn MSVC /analyze rules that generate too much noise. TODO: fix in the tool.
@@ -70,9 +72,9 @@ namespace gsl
 {
 
 // [views.constants], constants
-constexpr const std::ptrdiff_t dynamic_extent = -1;
+constexpr const std::size_t dynamic_extent = std::numeric_limits<std::size_t>::max();
 
-template <class ElementType, std::ptrdiff_t Extent = dynamic_extent>
+template <class ElementType, std::size_t Extent = dynamic_extent>
 class span;
 
 // implementation details
@@ -83,7 +85,7 @@ namespace details
     {
     };
 
-    template <class ElementType, std::ptrdiff_t Extent>
+    template <class ElementType, std::size_t Extent>
     struct is_span_oracle<gsl::span<ElementType, Extent>> : std::true_type
     {
     };
@@ -108,7 +110,7 @@ namespace details
     {
     };
 
-    template <std::ptrdiff_t From, std::ptrdiff_t To>
+    template <std::size_t From, std::size_t To>
     struct is_allowed_extent_conversion
         : public std::integral_constant<bool, From == To || From == gsl::dynamic_extent ||
                                                   To == gsl::dynamic_extent>
@@ -151,10 +153,10 @@ namespace details
             : span_iterator(other.span_, other.index_)
         {}
 
-        GSL_SUPPRESS(bounds.1) // NO-FORMAT: attribute
+        GSL_SUPPRESS(bounds .1) // NO-FORMAT: attribute
         constexpr reference operator*() const
         {
-            Expects(index_ != span_->size());
+            Expects(static_cast<size_t>(index_) != span_->size());
             return *(span_->data() + index_);
         }
 
@@ -166,7 +168,7 @@ namespace details
 
         constexpr span_iterator& operator++() noexcept
         {
-            Expects(0 <= index_ && index_ != span_->size());
+            Expects(0 <= index_ && static_cast<size_t>(index_) != span_->size());
             ++index_;
             return *this;
         }
@@ -180,7 +182,7 @@ namespace details
 
         constexpr span_iterator& operator--() noexcept
         {
-            Expects(index_ != 0 && index_ <= span_->size());
+            Expects(index_ != 0 && static_cast<size_t>(index_) <= span_->size());
             --index_;
             return *this;
         }
@@ -198,7 +200,8 @@ namespace details
             return ret += n;
         }
 
-        friend constexpr span_iterator operator+(difference_type n, span_iterator const& rhs) noexcept
+        friend constexpr span_iterator operator+(difference_type n,
+                                                 span_iterator const& rhs) noexcept
         {
             return rhs + n;
         }
@@ -272,7 +275,7 @@ namespace details
             Expects((index_ + n) >= 0 && (index_ + n) <= span_->size());
         }
 
-        GSL_SUPPRESS(bounds.1) // NO-FORMAT: attribute
+        GSL_SUPPRESS(bounds .1) // NO-FORMAT: attribute
         constexpr pointer _Unwrapped() const noexcept
         { // after seeking *this to a high water mark, or using one of the
             // _Verify_xxx functions above, unwrap this span_iterator to a raw
@@ -287,7 +290,7 @@ namespace details
 #else
         static constexpr bool _Unwrap_when_unverified = false;
 #endif
-        GSL_SUPPRESS(con.3) // NO-FORMAT: attribute // TODO: false positive
+        GSL_SUPPRESS(con .3) // NO-FORMAT: attribute // TODO: false positive
         constexpr void _Seek_to(const pointer p) noexcept
         { // adjust the position of *this to previously verified location p
             // after _Unwrapped
@@ -300,11 +303,11 @@ namespace details
         std::ptrdiff_t index_ = 0;
     };
 
-    template <std::ptrdiff_t Ext>
+    template <std::size_t Ext>
     class extent_type
     {
     public:
-        using index_type = std::ptrdiff_t;
+        using index_type = std::size_t;
 
         static_assert(Ext >= 0, "A fixed-size span must be >= 0 in size.");
 
@@ -327,13 +330,16 @@ namespace details
     class extent_type<dynamic_extent>
     {
     public:
-        using index_type = std::ptrdiff_t;
+        using index_type = std::size_t;
 
         template <index_type Other>
         explicit constexpr extent_type(extent_type<Other> ext) : size_(ext.size())
         {}
 
-        explicit constexpr extent_type(index_type size) : size_(size) { Expects(size >= 0); }
+        explicit constexpr extent_type(index_type size) : size_(size)
+        {
+            Expects(size != dynamic_extent);
+        }
 
         constexpr index_type size() const noexcept { return size_; }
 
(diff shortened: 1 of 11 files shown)


JordanMaples:
Would you mind taking a look at this @StephanTLavavej @billyoneal @sunnychatterjee?

JordanMaples:
Forgot to tag @CaseyCarter as well.

include/gsl/span
```diff
@@ -512,23 +518,25 @@ public:
...
-    
-    // at and operator() are deprecated to align to the public member functions of std::span
-    [[deprecated("Use operator[]")]]
-    constexpr reference at(index_type idx) const noexcept{ return this->operator[](idx); }
-    [[deprecated("Use operator[]")]]
-    constexpr reference operator()(index_type idx) const noexcept{ return this->operator[](idx); }
-
 
+    // at and operator() are deprecated to align to the public member functions of std::span
+    [[deprecated("Use operator[]")]] constexpr reference at(index_type idx) const noexcept
```
BillyONeal: If you're already forcing this big breaking change on users to change signedness, perhaps removing at() at this time would also be a good idea?

include/gsl/span
```diff
@@ -451,45 +455,47 @@ public:
...
-    template <std::ptrdiff_t Offset, std::ptrdiff_t Count = dynamic_extent>
-    GSL_SUPPRESS(bounds.1) // NO-FORMAT: attribute
+    template <std::size_t Offset, std::size_t Count = dynamic_extent>
+    GSL_SUPPRESS(bounds .1) // NO-FORMAT: attribute
     constexpr auto subspan() const noexcept ->
         typename details::calculate_subspan_type<ElementType, Extent, Offset, Count>::type
     {
-        Expects((Offset >= 0 && size() - Offset >= 0) &&
+        Expects((size() >= Offset) &&
                 (Count == dynamic_extent || (Count >= 0 && Offset + Count <= size())));
```
BillyONeal: This `Count >= 0` is a no-op.

include/gsl/span
```diff
@@ -629,69 +637,69 @@ private:
...
         const span<element_type, dynamic_extent> tmp(*this);
         return tmp.subspan(offset, count);
     }
 
-    GSL_SUPPRESS(bounds.1) // NO-FORMAT: attribute
+    GSL_SUPPRESS(bounds .1) // NO-FORMAT: attribute
     span<element_type, dynamic_extent> make_subspan(index_type offset, index_type count,
                                                     subspan_selector<dynamic_extent>) const
     {
-        Expects(offset >= 0 && size() - offset >= 0);
```
BillyONeal: Another no-op >= 0. Probably best to audit for these.

include/gsl/string_span
```diff
@@ -108,18 +108,20 @@ namespace details
 // Will fail-fast if sentinel cannot be found before max elements are examined.
 //
 template <typename T, const T Sentinel>
-span<T, dynamic_extent> ensure_sentinel(T* seq, std::ptrdiff_t max = PTRDIFF_MAX)
+span<T, dynamic_extent> ensure_sentinel(T* seq,
```
BillyONeal: Preexisting perf: probably special casing a 1 overload version instead of default arguments will win

tests/span_tests.cpp
```diff
@@ -158,27 +158,27 @@ TEST(span_test, from_pointer_length_constructor)
     int arr[4] = {1, 2, 3, 4};
 
     {
-        for (int i = 0; i < 4; ++i)
+        for (std::size_t i = 0; i < 4; ++i)
```
BillyONeal: Technically this one should still work as int because the loop ensures it is in the bounds [0, 4). Is changing to size_t for warning resistance?

include/gsl/span
```diff
@@ -300,11 +303,11 @@ namespace details
         std::ptrdiff_t index_ = 0;
```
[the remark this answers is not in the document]
JordanMaples: I'm preparing another PR to handle the iterators.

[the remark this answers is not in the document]
BillyONeal: Maybe I tagged the wrong member. I'm just indicating that in several places you are static_cast-ing this thing to size_t.

include/gsl/span
JordanMaples: Yeah, this is as good of a time as any to remove at().

tests/span_tests.cpp
BillyONeal: (Occurs elsewhere in these tests)

include/gsl/span
JordanMaples: good catch, I thought I removed this before I pushed.

tests/span_tests.cpp
JordanMaples: Yeah, as an int this was generating an additional warning. Just trying to avoid that.

include/gsl/span
JordanMaples: removed.

include/gsl/span
```diff
@@ -451,45 +454,47 @@ public:
...
-    GSL_SUPPRESS(bounds.1) // NO-FORMAT: attribute
+    template <std::size_t Offset, std::size_t Count = dynamic_extent>
+    GSL_SUPPRESS(bounds .1) // NO-FORMAT: attribute
     constexpr auto subspan() const noexcept ->
         typename details::calculate_subspan_type<ElementType, Extent, Offset, Count>::type
     {
-        Expects((Offset >= 0 && size() - Offset >= 0) &&
-                (Count == dynamic_extent || (Count >= 0 && Offset + Count <= size())));
+        Expects((size() >= Offset) &&
+                (Count == dynamic_extent || (Offset + Count <= size())));
```
CaseyCarter: `Count` could be so enormous that `Offset + Count` wraps and is `< size()` despite that `Count > size()` (See https://cplusplus.github.io/LWG/issue3103).
```suggestion
                (Count == dynamic_extent || (Count <= size() - Offset)));
```

include/gsl/span
```diff
@@ -512,23 +517,25 @@ public:
 
     constexpr reference front() const noexcept
     {
-        Expects(size() > 0);
+        Expects(size() != dynamic_extent && size() > 0);
```
CaseyCarter: When will it ever be the case that `size() == dynamic_extent`? (Ditto line 526.)

include/gsl/string_span
```diff
@@ -23,7 +23,7 @@
 
 #include <algorithm> // for equal, lexicographical_compare
 #include <array>     // for array
-#include <cstddef>   // for ptrdiff_t, size_t, nullptr_t
+#include <cstddef>   // for size_t, nullptr_t
 #include <cstdint>   // for PTRDIFF_MAX
 #include <cstring>
 #include <string>      // for basic_string, allocator, char_traits
```
CaseyCarter:
```suggestion
#include <string>    // for basic_string, allocator, char_traits
```

include/gsl/string_span
```diff
@@ -108,18 +108,20 @@ namespace details
...
 template <typename T, const T Sentinel>
-span<T, dynamic_extent> ensure_sentinel(T* seq, std::ptrdiff_t max = PTRDIFF_MAX)
+span<T, dynamic_extent> ensure_sentinel(T* seq,
+                                        std::size_t max = std::numeric_limits<std::size_t>::max())
 {
     Ensures(seq != nullptr);
 
-    GSL_SUPPRESS(f.23) // NO-FORMAT: attribute // TODO: false positive // TODO: suppress does not work
+    GSL_SUPPRESS(
+        f .23) // NO-FORMAT: attribute // TODO: false positive // TODO: suppress does not work
```
CaseyCarter: // TODO: NO-FORMAT doesn't work? :trollface:

include/gsl/span
JordanMaples: Shouldn't be, I'll remove those lines really quick.

include/gsl/string_span
JordanMaples: dang... gotta go fix all of that.

BillyONeal:
Do you have death tests for each of the "Expects"? We tested span with this death testing infrastructure:

```
// Copyright (c) Microsoft Corporation.
// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception

#pragma once
#include <crtdbg.h>
#include <stdint.h>
#include <stdio.h>
#include <stdlib.h>
#include <string>
#include <test_windows.h>
#include <vector>

namespace std_testing {

    constexpr int internal_failure = 103;
    using normal_function_t        = void (*)();
    using death_function_t         = void (*)();

    [[noreturn]] inline void api_unexpected(const char* const api_name) {
        const auto last_error = static_cast<unsigned int>(::GetLastError());
        printf("%s failed; LastError: 0x%08X\n", api_name, last_error);
        abort();
    }


    class death_test_executive {
        const normal_function_t run_normal_tests;
        std::vector<death_function_t> death_tests;

        int execute_death_test(const char* const test_id) const {
            const auto testId = static_cast<size_t>(atoi(test_id));

            if (errno == 0) {
                _set_abort_behavior(0, _WRITE_ABORT_MSG);
                _CrtSetReportMode(_CRT_ASSERT, _CRTDBG_MODE_FILE);
                _CrtSetReportFile(_CRT_ASSERT, _CRTDBG_FILE_STDOUT);

                death_tests[testId]();
                return 0;
            } else {
                puts("failed to parse test_id");
                return internal_failure;
            }
        }

        DWORD dispatch_death_test(const size_t test_id, const wchar_t* const this_program) const {
            STARTUPINFOW si{};
            si.cb = sizeof(si);

            PROCESS_INFORMATION pi{};

            std::wstring test_id_str(1, L' ');
            test_id_str.append(std::to_wstring(test_id));
            if (::CreateProcessW(this_program, &test_id_str[0], nullptr, nullptr, FALSE, 0, nullptr, nullptr, &si, &pi)
                == 0) {
                api_unexpected("CreateProcessW");
            }

            if (::WaitForSingleObject(pi.hProcess, INFINITE) != WAIT_OBJECT_0) {
                api_unexpected("WaitForSingleObject");
            }

            DWORD exit_code = 0;
            if (::GetExitCodeProcess(pi.hProcess, &exit_code) == 0) {
                api_unexpected("GetExitCodeProcess");
            }

            ::CloseHandle(pi.hThread);
            ::CloseHandle(pi.hProcess);

            return exit_code;
        }

        static std::wstring get_current_process_path() {
            std::wstring result(MAX_PATH, L'\0');
            for (;;) {
                const DWORD result_size = ::GetModuleFileNameW(nullptr, &result[0], static_cast<DWORD>(result.size()));
                const size_t str_size   = result.size();
                if (result_size == str_size) {
                    // buffer was not big enough
                    const size_t str_max_size    = result.max_size();
                    const size_t result_max_size = str_max_size - str_max_size / 2;
                    if (result_size >= result_max_size) {
                        api_unexpected("GetModuleFileNameW");
                    }

                    result.resize(result_size + result_size / 2);
                } else if (result_size == 0) {
                    api_unexpected("GetModuleFileNameW");
                } else {
                    result.resize(result_size);
                    break;
                }
            }

            return result;
        }

    public:
        explicit death_test_executive(const normal_function_t normal_tests_function)
            : run_normal_tests(normal_tests_function) {}


        template <size_t TestsCount>
        void add_death_tests(const death_function_t (&tests)[TestsCount]) {
            death_tests.insert(death_tests.end(), tests, std::end(tests));
        }

        int run(int argc, char* argv[]) const {
            if (argc == 1) {
                // first pass, run normal tests and sub-process loop
                printf("running normal tests...");
                run_normal_tests();
                puts(" passed!");

                ::SetErrorMode(SEM_FAILCRITICALERRORS | SEM_NOGPFAULTERRORBOX);
                const auto this_program       = get_current_process_path();
                const size_t death_tests_size = death_tests.size();
                for (size_t idx = 0; idx < death_tests_size; ++idx) {
                    printf("running death test %zu... ", idx);
                    const DWORD death_test_result = dispatch_death_test(idx, this_program.c_str());
                    if (death_test_result <= 1000U) {
                        printf("returned %lu", death_test_result);
                    } else {
                        printf("returned 0x%lX", death_test_result);
                    }

                    if (death_test_result == 0 || death_test_result == 100) {
                        puts(", a success code (this is bad)");
                        puts("Terminate!");
                        return 1;
                    } else if (death_test_result == internal_failure) {
                        puts(", an internal test harness failure");
                        puts("Terminate!");
                        return 1;
                    } else {
                        puts(", a failure code (this is good)");
                    }
                }

                return 0;
            } else if (argc == 2) {
                return execute_death_test(argv[1]);
            } else {
                puts("incorrect command line");
                return 1;
            }
        }
    };
} // namespace std_testing
```

```
// Copyright (c) Microsoft Corporation.
// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception

#define _CONTAINER_DEBUG_LEVEL 1

#include <algorithm>
#include <span>
#include <stddef.h>
#include <test_death.hpp>
using namespace std;

int globalArray[5]{10, 20, 30, 40, 50};
int otherArray[5]{10, 20, 30, 40, 50};

void test_case_operator_dereference_value_initialized_iterator() {
    span<int>::iterator it; // note: for IDL to work correctly, default-init and value-init are equivalent
    (void) *it; // cannot dereference value-initialized span iterator
}

void test_case_operator_dereference_end_iterator() {
    span<int> sp(globalArray);
    span<int>::iterator it = sp.end();
    (void) *it; // cannot dereference end span iterator
}

void test_case_operator_arrow_value_initialized_iterator() {
    span<int>::iterator it;
    (void) it.operator->(); // cannot dereference value-initialized span iterator
}

void test_case_operator_arrow_end_iterator() {
    span<int> sp(globalArray);
    span<int>::iterator it = sp.end();
    (void) it.operator->(); // cannot dereference end span iterator
}

void test_case_operator_preincrement_value_initialized_iterator() {
    span<int>::iterator it;
    ++it; // cannot increment value-initialized span iterator
}

void test_case_operator_preincrement_after_end() {
    span<int> sp(globalArray);
    span<int>::iterator it = sp.end();
    ++it; // cannot increment span iterator past end
}

void test_case_operator_predecrement_value_initialized_iterator() {
    span<int>::iterator it;
    --it; // cannot decrement value-initialized span iterator
}

void test_case_operator_predecrement_before_begin() {
    span<int> sp(globalArray);
    span<int>::iterator it = sp.begin();
    --it; // cannot decrement span iterator before begin
}

void test_case_operator_advance_value_initialized_iterator() {
    span<int>::iterator it;
    it += 1; // cannot seek value-initialized span iterator
}

void test_case_operator_advance_value_initialized_iterator_zero() {
    span<int>::iterator it;
    it += 0; // OK
}

void test_case_operator_advance_before_begin() {
    span<int> sp(globalArray);
    span<int>::iterator it = sp.begin();
    it += -1; // cannot seek span iterator before begin
}

void test_case_operator_advance_after_end() {
    span<int> sp(globalArray);
    span<int>::iterator it = sp.end();
    it += 1; // cannot seek span iterator after end
}

void test_case_operator_retreat_value_initialized_iterator() {
    span<int>::iterator it;
    it -= 1; // cannot seek value-initialized span iterator
}

void test_case_operator_retreat_value_initialized_iterator_zero() {
    span<int>::iterator it;
    it -= 0; // OK
}

void test_case_operator_retreat_before_begin() {
    span<int> sp(globalArray);
    span<int>::iterator it = sp.begin();
    it -= 1; // cannot seek span iterator before begin
}

void test_case_operator_retreat_after_end() {
    span<int> sp(globalArray);
    span<int>::iterator it = sp.end();
    it -= -1; // cannot seek span iterator after end
}

void test_case_operator_subtract_incompatible_different_data() {
    span<int> sp1(globalArray);
    span<int> sp2(otherArray);
    (void) (sp1.begin() - sp2.begin()); // cannot subtract incompatible span iterators
}

void test_case_operator_subtract_incompatible_different_size() {
    span<int> sp1(globalArray, 3);
    span<int> sp2(globalArray, 4);
    (void) (sp1.begin() - sp2.begin()); // cannot subtract incompatible span iterators
}

void test_case_operator_subtract_incompatible_value_initialized() {
    span<int> sp(globalArray);
    (void) (sp.begin() - span<int>::iterator{}); // cannot subtract incompatible span iterators
}

void test_case_operator_equal_incompatible_different_data() {
    span<int> sp1(globalArray);
    span<int> sp2(otherArray);
    (void) (sp1.begin() == sp2.begin()); // cannot compare incompatible span iterators for equality
}

void test_case_operator_equal_incompatible_different_size() {
    span<int> sp1(globalArray, 3);
    span<int> sp2(globalArray, 4);
    (void) (sp1.begin() == sp2.begin()); // cannot compare incompatible span iterators for equality
}

void test_case_operator_equal_incompatible_value_initialized() {
    span<int> sp(globalArray);
    (void) (sp.begin() == span<int>::iterator{}); // cannot compare incompatible span iterators for equality
}

void test_case_operator_less_incompatible_different_data() {
    span<int> sp1(globalArray);
    span<int> sp2(otherArray);
    (void) (sp1.begin() < sp2.begin()); // cannot compare incompatible span iterators
}

void test_case_operator_less_incompatible_different_size() {
    span<int> sp1(globalArray, 3);
    span<int> sp2(globalArray, 4);
    (void) (sp1.begin() < sp2.begin()); // cannot compare incompatible span iterators
}

void test_case_operator_less_incompatible_value_initialized() {
    span<int> sp(globalArray);
    (void) (sp.begin() < span<int>::iterator{}); // cannot compare incompatible span iterators
}

void test_case_algorithm_incompatible_different_data() {
    span<int> sp1(globalArray);
    span<int> sp2(otherArray);
    (void) find(sp1.begin(), sp2.begin(), -1); // span iterators from different views do not form a range
}

void test_case_algorithm_incompatible_different_size() {
    span<int> sp1(globalArray, 3);
    span<int> sp2(globalArray, 4);
    (void) find(sp1.begin(), sp2.begin(), -1); // span iterators from different views do not form a range
}

void test_case_algorithm_incompatible_value_initialized() {
    span<int> sp(globalArray);
    (void) find(sp.begin(), span<int>::iterator{}, -1); // span iterators from different views do not form a range
}

void test_case_algorithm_incompatible_transposed() {
    span<int> sp(globalArray);
    (void) find(sp.end(), sp.begin(), -1); // span iterator range transposed
}

void test_case_constructor_first_count_incompatible_extent() {
    span<int, 3> sp(begin(globalArray), size(globalArray)); // span construction from invalid range as count != extent
    (void) sp;
}

void test_case_constructor_first_last_incompatible_extent() {
    span<int, 3> sp(
        begin(globalArray), end(globalArray)); // span construction from invalid range as last - first != extent
    (void) sp;
}

void test_case_first_excessive_compiletime_count() {
    span<int> sp(globalArray);
    (void) sp.first<6>(); // Count out of range in span::first()
}

void test_case_first_excessive_runtime_count_dynamic_extent() {
    span<int> sp(globalArray);
    (void) sp.first(6); // Count out of range in span::first(count)
}

void test_case_first_excessive_runtime_count_static_extent() {
    span<int, 5> sp(globalArray);
    (void) sp.first(6); // Count out of range in span::first(count)
}

void test_case_last_excessive_compiletime_count() {
    span<int> sp(globalArray);
    (void) sp.last<6>(); // Count out of range in span::last()
}

void test_case_last_excessive_runtime_count_dynamic_extent() {
    span<int> sp(globalArray);
    (void) sp.last(6); // Count out of range in span::last(count)
}

void test_case_last_excessive_runtime_count_static_extent() {
    span<int, 5> sp(globalArray);
    (void) sp.last(6); // Count out of range in span::last(count)
}

void test_case_subspan_excessive_compiletime_offset() {
    span<int> sp(globalArray);
    (void) sp.subspan<6>(); // Offset out of range in span::subspan()
}

void test_case_subspan_excessive_compiletime_count() {
    span<int> sp(globalArray);
    (void) sp.subspan<2, 4>(); // Count out of range in span::subspan()
}

void test_case_subspan_excessive_runtime_offset_dynamic_extent() {
    span<int> sp(globalArray);
    (void) sp.subspan(6); // Offset out of range in span::subspan(offset, count)
}

void test_case_subspan_excessive_runtime_count_dynamic_extent() {
    span<int> sp(globalArray);
    (void) sp.subspan(2, 4); // Count out of range in span::subspan(offset, count)
}

void test_case_subspan_excessive_runtime_offset_static_extent() {
    span<int, 5> sp(globalArray);
    (void) sp.subspan(6); // Offset out of range in span::subspan(offset, count)
}

void test_case_subspan_excessive_runtime_count_static_extent() {
    span<int, 5> sp(globalArray);
    (void) sp.subspan(2, 4); // Count out of range in span::subspan(offset, count)
}

void test_case_size_bytes_overflow() {
    span<int> sp(begin(globalArray), static_cast<size_t>(-2)); // undefined behavior, not detected here
    (void) sp.size_bytes(); // size of span in bytes exceeds std::numeric_limits<size_t>::max()
}

void test_case_operator_subscript_out_of_range_dynamic_extent() {
    span<int> sp(globalArray);
    (void) sp[5]; // span index out of range
}

void test_case_operator_subscript_out_of_range_static_extent() {
    span<int, 5> sp(globalArray);
    (void) sp[5]; // span index out of range
}

void test_case_front_empty_dynamic_extent() {
    span<int> sp;
    (void) sp.front(); // front of empty span
}

void test_case_back_empty_dynamic_extent() {
    span<int> sp;
    (void) sp.back(); // back of empty span
}

void test_case_front_empty_static_extent() {
    span<int, 0> sp;
    (void) sp.front(); // front of empty span
}

void test_case_back_empty_static_extent() {
    span<int, 0> sp;
    (void) sp.back(); // back of empty span
}

int main(int argc, char* argv[]) {
    std_testing::death_test_executive exec([] {
        test_case_operator_advance_value_initialized_iterator_zero();
        test_case_operator_retreat_value_initialized_iterator_zero();
    });

#if _ITERATOR_DEBUG_LEVEL != 0
    exec.add_death_tests({
        test_case_operator_dereference_value_initialized_iterator,
        test_case_operator_dereference_end_iterator,
        test_case_operator_arrow_value_initialized_iterator,
        test_case_operator_arrow_end_iterator,
        test_case_operator_preincrement_value_initialized_iterator,
        test_case_operator_preincrement_after_end,
        test_case_operator_predecrement_value_initialized_iterator,
        test_case_operator_predecrement_before_begin,
        test_case_operator_advance_value_initialized_iterator,
        test_case_operator_advance_before_begin,
        test_case_operator_advance_after_end,
        test_case_operator_retreat_value_initialized_iterator,
        test_case_operator_retreat_before_begin,
        test_case_operator_retreat_after_end,
        test_case_operator_subtract_incompatible_different_data,
        test_case_operator_subtract_incompatible_different_size,
        test_case_operator_subtract_incompatible_value_initialized,
        test_case_operator_equal_incompatible_different_data,
        test_case_operator_equal_incompatible_different_size,
        test_case_operator_equal_incompatible_value_initialized,
        test_case_operator_less_incompatible_different_data,
        test_case_operator_less_incompatible_different_size,
        test_case_operator_less_incompatible_value_initialized,
        test_case_algorithm_incompatible_different_data,
        test_case_algorithm_incompatible_different_size,
        test_case_algorithm_incompatible_value_initialized,
        test_case_algorithm_incompatible_transposed,
    });
#endif // _ITERATOR_DEBUG_LEVEL != 0

    // _CONTAINER_DEBUG_LEVEL tests
    exec.add_death_tests({
        test_case_constructor_first_count_incompatible_extent,
        test_case_constructor_first_last_incompatible_extent,
        test_case_first_excessive_compiletime_count,
        test_case_first_excessive_runtime_count_dynamic_extent,
        test_case_first_excessive_runtime_count_static_extent,
        test_case_last_excessive_compiletime_count,
        test_case_last_excessive_runtime_count_dynamic_extent,
        test_case_last_excessive_runtime_count_static_extent,
        test_case_subspan_excessive_compiletime_offset,
        test_case_subspan_excessive_compiletime_count,
        test_case_subspan_excessive_runtime_offset_dynamic_extent,
        test_case_subspan_excessive_runtime_count_dynamic_extent,
        test_case_subspan_excessive_runtime_offset_static_extent,
        test_case_subspan_excessive_runtime_count_static_extent,
        test_case_size_bytes_overflow,
        test_case_operator_subscript_out_of_range_dynamic_extent,
        test_case_operator_subscript_out_of_range_static_extent,
        test_case_front_empty_dynamic_extent,
        test_case_back_empty_dynamic_extent,
        test_case_front_empty_static_extent,
        test_case_back_empty_static_extent,
    });

    return exec.run(argc, argv);
}
```

include/gsl/span
```diff
@@ -646,52 +653,52 @@ private:
     span<element_type, dynamic_extent> make_subspan(index_type offset, index_type count,
                                                     subspan_selector<dynamic_extent>) const
     {
-        Expects(offset >= 0 && size() - offset >= 0);
+        Expects(size() >= offset && size() != dynamic_extent);
```
CaseyCarter: Ditto `size() != dynamic_extent`.

include/gsl/span
```diff
@@ -351,16 +356,17 @@ namespace details
...
 {
 public:
     // constants and types
     using element_type = ElementType;
     using value_type = std::remove_cv_t<ElementType>;
-    using index_type = std::ptrdiff_t;
+    using index_type = std::size_t;
     using pointer = element_type*;
     using reference = element_type&;
+    using difference_type = std::ptrdiff_t;
```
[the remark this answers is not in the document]
sunnychatterjee: Interesting. Based on https://en.cppreference.com/w/cpp/container/span, you do want difference_type as a ptrdiff_t after all. @BillyONeal could you please weigh in here?

include/gsl/span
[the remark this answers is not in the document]
JordanMaples: New iterator changes are live.

include/gsl/span
```diff
@@ -25,17 +25,18 @@
...
 #include <type_traits> // for enable_if_t, declval, is_convertible, inte...
 #include <utility>
-#include <memory> // for std::addressof
 
 #if defined(_MSC_VER) && !defined(__clang__)
 #pragma warning(push)
 
 // turn off some warnings that are noisy about our Expects statements
 #pragma warning(disable : 4127) // conditional expression is constant
+#pragma warning(                                                                                   \
```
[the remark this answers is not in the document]
JordanMaples: This is also being used by the n < 0 cases for the iterators.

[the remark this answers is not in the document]
StephanTLavavej: `static_cast<std::size_t>(-1)` doesn't trigger warning C4146. If that warning is being emitted, it's coming from unary minus being directly applied to `size_t`.

include/gsl/span
```diff
@@ -121,139 +122,165 @@ namespace details
     {
     };
 
-    template <class Span, bool IsConst>
+    template <class type>
```
CaseyCarter: Don't template parameters in this codebase conventionally have PascalCase names? (And on 224 and 238 and... many other places ;).)

include/gsl/span
```diff
@@ -70,9 +71,9 @@ namespace gsl
 {
 
 // [views.constants], constants
-constexpr const std::ptrdiff_t dynamic_extent = -1;
+constexpr const std::size_t dynamic_extent = static_cast<std::size_t>(-1);
```
StephanTLavavej: Pre-existing: `constexpr const` has always been redundant on scalar variables like this one. You can simply say `constexpr`.

include/gsl/span
```diff
@@ -121,139 +122,165 @@ namespace details
...
-
         using iterator_category = std::random_access_iterator_tag;
-        using value_type = std::remove_cv_t<element_type_>;
-        using difference_type = typename Span::index_type;
-
-        using reference = std::conditional_t<IsConst, const element_type_, element_type_>&;
-        using pointer = std::add_pointer_t<reference>;
+        using value_type = std::remove_cv_t<type>;
+        using difference_type = ptrdiff_t;
+        using pointer = std::add_pointer_t<type>;
```
CaseyCarter: Why pay metaprogramming overhead when you could simply write `type*`? (Ditto `type&` on 133.)

include/gsl/span
StephanTLavavej: Pre-existing: The lack of `inline constexpr` in C++17 mode is a minor difference from the Standard specification. It is unlikely to matter, but it is theoretically detectable if someone takes this variable's address.

include/gsl/span
StephanTLavavej: Style question: Does GSL permit lowercase template parameters like `type`?

include/gsl/span
```diff
@@ -121,139 +122,165 @@ namespace details
...
-        friend span_iterator<Span, true>;
-        template <bool B, std::enable_if_t<!B && IsConst>* = nullptr>
-        constexpr span_iterator(const span_iterator<Span, B>& other) noexcept
-            : span_iterator(other.span_, other.index_)
-        {}
-
-        GSL_SUPPRESS(bounds.1) // NO-FORMAT: attribute
-        constexpr reference operator*() const
+#ifdef _MSC_VER
+        using _Unchecked_type = typename pointer;
```
CaseyCarter: Belated response to your email question: the presence of `_Unchecked_type` tells an older STL (IIRC VS2017-era) that this iterator type participates in our checking and unwrapping machinery. Current STLs don't use it anymore, but STLs that you support do.

include/gsl/span
```diff
@@ -121,139 +122,165 @@ namespace details
...
-#endif
-
         using iterator_category = std::random_access_iterator_tag;
-        using value_type = std::remove_cv_t<element_type_>;
-        using difference_type = typename Span::index_type;
-
-        using reference = std::conditional_t<IsConst, const element_type_, element_type_>&;
-        using pointer = std::add_pointer_t<reference>;
+        using value_type = std::remove_cv_t<type>;
+        using difference_type = ptrdiff_t;
```
StephanTLavavej: Elsewhere, the code is careful to qualify `std::ptrdiff_t`.

include/gsl/span
```diff
@@ -121,139 +122,165 @@ namespace details
...
+        using pointer = std::add_pointer_t<type>;
+        using reference = std::add_lvalue_reference_t<type>;
```
StephanTLavavej: For simplicity, you could simply say `type *` and `type&` like the STL does. `add_pointer_t` and `add_lvalue_reference_t` are necessary in extremely generic code that has to handle types that can't be pointed/referred to (e.g. there are no pointers to abominable function types and no references to `void`). `span` elements are object types, so you can always form references/pointers to them.

include/gsl/span
StephanTLavavej: `typename` here is unnecessary and should be removed. The compiler knows it's a type because it's directly a typedef in this class.

include/gsl/span
```diff
@@ -121,139 +122,165 @@ namespace details
...
             return ret;
         }
 
-        constexpr span_iterator operator+(difference_type n) const noexcept
+        constexpr span_iterator& operator+=(const difference_type n) noexcept
         {
-            auto ret = *this;
+            Expects(begin_ && current_ && end_);
+            if (n > 0) Expects(end_ - current_ >= n);
+            if (n < 0) Expects(end_ - current_ >= -n);
```
CaseyCarter:
```suggestion
            if (n < 0) Expects(current_ - begin_ >= -n);
```

include/gsl/span
```diff
@@ -121,139 +122,165 @@ namespace details
...
+#endif
+        constexpr operator span_iterator<const type>() const noexcept
         {
-            Expects(index_ != span_->size());
-            return *(span_->data() + index_);
+            return {begin_, end_, current_};
         }
 
-        constexpr pointer operator->() const
+        constexpr reference operator*() const noexcept { return *operator->(); }
```
StephanTLavavej: In debug mode, every `operator*` (which is a common call) will pay an extra function call. Consider avoiding this cost.

include/gsl/span
```diff
@@ -121,139 +122,165 @@ namespace details
...
             return rhs + n;
         }
 
-        constexpr span_iterator& operator+=(difference_type n) noexcept
+        constexpr span_iterator& operator-=(const difference_type n) noexcept
         {
-            Expects((index_ + n) >= 0 && (index_ + n) <= span_->size());
-            index_ += n;
+            Expects(begin_ && end_ && current_);
+            if (n > 0) Expects(end_ - current_ >= n);
```
CaseyCarter:
```suggestion
            if (n > 0) Expects(current_ - begin_ >= n);
```

include/gsl/span
```diff
@@ -121,139 +122,165 @@ namespace details
...
+            auto ret{*this};
+            --*this;
             return ret;
         }
 
-        constexpr span_iterator operator+(difference_type n) const noexcept
+        constexpr span_iterator& operator+=(const difference_type n) noexcept
         {
-            auto ret = *this;
+            Expects(begin_ && current_ && end_);
```
StephanTLavavej: It should be acceptable to say `+= 0` for a value-initialized iterator.

include/gsl/span
```diff
@@ -121,139 +122,165 @@ namespace details
...
+
+        constexpr span_iterator operator+(const difference_type n) const noexcept
+        {
+            auto ret{*this};
             return ret += n;
         }
 
-        friend constexpr span_iterator operator+(difference_type n, span_iterator const& rhs) noexcept
+        friend constexpr span_iterator operator+(const difference_type n,
+                                                 span_iterator const& rhs) noexcept
```
StephanTLavavej: Pre-existing: This uses right-const when the file has previously used left-const.

include/gsl/span
```diff
@@ -121,139 +122,165 @@ namespace details
...
 
-        constexpr difference_type operator-(span_iterator rhs) const noexcept
+        template <
+            class type2,
+            std::enable_if_t<std::is_same<std::remove_cv_t<type2>, value_type>::value, int> = 0>
+        constexpr difference_type operator-(const span_iterator<type2>& rhs) const noexcept
         {
-            Expects(span_ == rhs.span_);
-            return index_ - rhs.index_;
+            Expects(begin_ == rhs.begin_);
```
CaseyCarter: `&& end_ == rhs.end_`?

include/gsl/span
```diff
@@ -121,139 +122,165 @@ namespace details
...
         {
             return rhs + n;
         }
 
-        constexpr span_iterator& operator+=(difference_type n) noexcept
+        constexpr span_iterator& operator-=(const difference_type n) noexcept
         {
-            Expects((index_ + n) >= 0 && (index_ + n) <= span_->size());
-            index_ += n;
+            Expects(begin_ && end_ && current_);
```
StephanTLavavej: Same comment, `-= 0` should be acceptable for value-initialized iterators.

include/gsl/span
```diff
@@ -121,139 +122,165 @@ namespace details
...
+            if (n > 0) Expects(end_ - current_ >= n);
+            if (n < 0) Expects(end_ - current_ >= -n);
```
StephanTLavavej: Please double-check these Expects. They appear to be wrong (as they are identical to `+=` Expects).

include/gsl/span
```diff
@@ -121,139 +122,165 @@ namespace details
...
+            auto ret{*this};
             return ret -= n;
         }
 
-        constexpr span_iterator& operator-=(difference_type n) noexcept { return *this += -n; }
+        friend constexpr span_iterator operator-(const difference_type n,
+                                                 span_iterator const& rhs) noexcept
+        {
+            return rhs - n;
+        }
```
StephanTLavavej: This operator should not exist: `5 + iter` is a required operation (also supported by pointers) but `5 - iter` doesn't make any sense.

include/gsl/span
```diff
@@ -121,139 +122,165 @@ namespace details
...
 
-        constexpr friend bool operator==(span_iterator lhs, span_iterator rhs) noexcept
+        template <
+            class type2,
+            std::enable_if_t<std::is_same<std::remove_cv_t<type2>, value_type>::value, int> = 0>
+        constexpr bool operator==(const span_iterator<type2>& rhs) const noexcept
         {
-            return lhs.span_ == rhs.span_ && lhs.index_ == rhs.index_;
+            return begin_ == rhs.begin_ && current_ == rhs.current_;
```
CaseyCarter: `&& end_ == rhs.end_` would be consistent here. Given that the standard only requires `x == y` to be valid when `x` and `y` are iterators into the same range, I would write this as `Expects(begin_ == rhs.begin_ && end_ == rhs.end_); return current_ == rhs.current_;`.

include/gsl/span
```diff
@@ -121,139 +122,165 @@ namespace details
...
         }
 
-        constexpr friend bool operator<(span_iterator lhs, span_iterator rhs) noexcept
+        template <
+            class type2,
+            std::enable_if_t<std::is_same<std::remove_cv_t<type2>, value_type>::value, int> = 0>
+        constexpr bool operator<(const span_iterator<type2>& rhs) const noexcept
         {
-            return lhs.index_ < rhs.index_;
+            Expects(begin_ == rhs.begin_);
```
CaseyCarter: `&& end_ == rhs.end_`?

include/gsl/span
```diff
@@ -121,139 +122,165 @@ namespace details
...
         }
 
-        constexpr friend bool operator<=(span_iterator lhs, span_iterator rhs) noexcept
+        template <
+            class type2,
+            std::enable_if_t<std::is_same<std::remove_cv_t<type2>, value_type>::value, int> = 0>
+        constexpr bool operator>(const span_iterator<type2>& rhs) const noexcept
         {
-            return !(rhs < lhs);
+            return !(*this < rhs);
```
CaseyCarter:
```suggestion
            return rhs < *this;
```

include/gsl/span
StephanTLavavej: This doesn't validate `end_` so it won't detect iterators to some incompatible spans (those where the begins are identical but the sizes are different).

include/gsl/span
```diff
@@ -121,139 +122,165 @@ namespace details
...
         }
 
-        constexpr friend bool operator>(span_iterator lhs, span_iterator rhs) noexcept
+        template <
+            class type2,
+            std::enable_if_t<std::is_same<std::remove_cv_t<type2>, value_type>::value, int> = 0>
+        constexpr bool operator<=(const span_iterator<type2>& rhs) const noexcept
         {
-            return rhs < lhs;
+            return *this < rhs || *this == rhs;
```
CaseyCarter:
```suggestion
            return !(rhs < *this);
```

include/gsl/span
```diff
@@ -121,139 +122,165 @@ namespace details
...
         }
 
-        constexpr friend bool operator>=(span_iterator lhs, span_iterator rhs) noexcept
+        template <
+            class type2,
+            std::enable_if_t<std::is_same<std::remove_cv_t<type2>, value_type>::value, int> = 0>
+        constexpr bool operator>=(const span_iterator<type2>& rhs) const noexcept
         {
-            return !(rhs > lhs);
+            return *this > rhs || *this == rhs;
```
CaseyCarter:
```suggestion
            return !(*this < rhs);
```

include/gsl/span
StephanTLavavej: This is incorrect. It permits incompatible span iterators to be compared (and return false), and it additionally ignores whether `end_` differs. Instead, it should Expect that `begin_` and `end_` are identical, then it should compare `current_`.

include/gsl/span
StephanTLavavej: Same issue of not Expecting `end_`.

include/gsl/span
StephanTLavavej: This is an inefficient way to express the operation (due to repeated validation). It is better to transform this into a single call to `operator<`. `left <= right` is `!(right < left)`.

include/gsl/span
StephanTLavavej: Ditto, `left >= right` is `!(left < right)`.

include/gsl/span
```diff
@@ -262,22 +289,22 @@ namespace details
         // algorithm calls
         friend constexpr void _Verify_range(span_iterator lhs, span_iterator rhs) noexcept
         { // test that [lhs, rhs) forms a valid range inside an STL algorithm
-            Expects(lhs.span_ == rhs.span_        // range spans have to match
-                    && lhs.index_ <= rhs.index_); // range must not be transposed
+            Expects(lhs.begin_ == rhs.begin_    // range spans have to match
+                    && lhs.end_ <= rhs.end_); // range must not be transposed
```
CaseyCarter: Shouldn't this check that begins and ends are equal (spans match) and that `lhs.current_ <= rhs.current_`?

include/gsl/span
```diff
@@ -262,22 +289,22 @@ namespace details
...
         }
 
         constexpr void _Verify_offset(const difference_type n) const noexcept
         { // test that the iterator *this + n is a valid range in an STL
```
CaseyCarter:
```suggestion
        { // test that *this + n is within the span of this iterator STL
```

include/gsl/span
StephanTLavavej: This has been incorrectly modified. It should be validating that begin and end are the same, and then the indices should be `<=`. That is, what used to be two tests needs to be three (because of the removal of the parent pointer).

include/gsl/span
```diff
@@ -262,22 +289,22 @@ namespace details
...
             // algorithm call
-            Expects((index_ + n) >= 0 && (index_ + n) <= span_->size());
+            Expects((current_ + n) >= begin_ && (current_ + n) <= end_);
```
StephanTLavavej: This Expects has been written to trigger pointer overflow and then look at the consequences, which is potentially problematic. I recommend directly testing the unknown `n` against known-valid expressions like `end_ - current_`.

include/gsl/span
```diff
@@ -291,20 +318,20 @@ namespace details
...
 
-    template <std::ptrdiff_t Ext>
+    template <std::size_t Ext>
     class extent_type
     {
     public:
-        using index_type = std::ptrdiff_t;
+        using index_type = std::size_t;
 
         static_assert(Ext >= 0, "A fixed-size span must be >= 0 in size.");
```
StephanTLavavej: Not pre-existing: This `static_assert` is now always true, and should instead be checking for `dynamic_extent`. Or it should be removed entirely, since there is an explicit specialization immediately below that will absorb `dynamic_extent`.

include/gsl/span
```diff
@@ -327,21 +354,24 @@ namespace details
...
 
     private:
         index_type size_;
     };
 
-    template <class ElementType, std::ptrdiff_t Extent, std::ptrdiff_t Offset, std::ptrdiff_t Count>
+    template <class ElementType, std::size_t Extent, std::size_t Offset, std::size_t Count>
     struct calculate_subspan_type
     {
         using type = span<ElementType, Count != dynamic_extent
```
StephanTLavavej: Pre-existing: Consider making this an alias template, which is both simpler and higher throughput.

include/gsl/span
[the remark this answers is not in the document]
CaseyCarter: The `difference_type` of a container is the same as the `difference_type` of its iterators, and the `difference_type` of iterators is always a signed integer type.

include/gsl/span
```diff
@@ -351,19 +381,20 @@ namespace details
...
-template <class ElementType, std::ptrdiff_t Extent>
+template <class ElementType, std::size_t Extent>
 class span
 {
 public:
     // constants and types
     using element_type = ElementType;
     using value_type = std::remove_cv_t<ElementType>;
-    using index_type = std::ptrdiff_t;
+    using index_type = std::size_t;
```
StephanTLavavej: `std::span` has `size_type` but not `index_type`. You should consider adding `size_type` for interface compatibility, and possibly consider marking `index_type` as deprecated.

CaseyCarter: `std::span` has a `size_type` instead of an `index_type` now.

include/gsl/span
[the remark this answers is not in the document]
StephanTLavavej: `difference_type` on a container or view is signed because iterator distances are signed. (Random-access iterators can be subtracted and will return negative if they are transposed.)

include/gsl/span
```diff
@@ -379,11 +410,11 @@ public:
     template <bool Dependent = false,
               // "Dependent" is needed to make "std::enable_if_t<Dependent || Extent <= 0>" SFINAE,
               // since "std::enable_if_t<Extent <= 0>" is ill-formed when Extent is greater than 0.
```
StephanTLavavej: This comment is outdated now.

include/gsl/span
```diff
@@ -584,7 +616,7 @@ private:
         }
```
CaseyCarter: This function seems to be useless now: `index_type` *is* `size_t` so it needn't be narrowed to `size_t`.

include/gsl/span
```diff
@@ -521,23 +551,25 @@ public:
...
+        return this->operator[](idx);
+    }
 
     constexpr pointer data() const noexcept { return storage_.data(); }
 
     // [span.iter], span iterator support
-    constexpr iterator begin() const noexcept { return {this, 0}; }
-    constexpr iterator end() const noexcept { return {this, size()}; }
+    constexpr iterator begin() const noexcept { return {data(), data() + size(), data()}; }
+    constexpr iterator end() const noexcept { return {data(), data() + size(), data() + size()}; }
```
StephanTLavavej: Debug codegen: Consider factoring out the calls to `data()` and `size()` to avoid calling them repeatedly, as `begin()` and `end()` are popular functions. Ditto for `cbegin()` and `cend()` below.

include/gsl/span
JordanMaples: Yep, I was just about to push the removal of this.

include/gsl/span
```diff
@@ -646,52 +678,52 @@ private:
...
     }
 };
 
 #if defined(GSL_USE_STATIC_CONSTEXPR_WORKAROUND)
-template <class ElementType, std::ptrdiff_t Extent>
+template <class ElementType, std::size_t Extent>
 constexpr const typename span<ElementType, Extent>::index_type span<ElementType, Extent>::extent;
 #endif
 
 // [span.comparison], span comparison operators
```
StephanTLavavej: These were removed for C++20 by [P1085R2](https://wg21.link/P1085R2).

include/gsl/span
```diff
@@ -785,65 +814,74 @@ constexpr span<typename Ptr::element_type> make_span(Ptr& cont)
...
     // No bounds checking here because it is done in span::operator[] called below
     return s[i];
 }
 
 // [span.obs] Free observer functions
-template <class ElementType, std::ptrdiff_t Extent>
-constexpr typename span<ElementType, Extent>::index_type ssize(const span<ElementType, Extent> &span) noexcept
+template <class ElementType, std::size_t Extent>
+constexpr typename span<ElementType, Extent>::index_type
+ssize(const span<ElementType, Extent>& span) noexcept
```
StephanTLavavej: `ssize` is now returning an unsigned `index_type` which is contrary to the intent of the function.

include/gsl/span
```diff
@@ -785,65 +814,74 @@ constexpr span<typename Ptr::element_type> make_span(Ptr& cont)
...
 {
     return span.size();
 }
 
 // [span.iter] Free functions for begin/end functions
-template <class ElementType, std::ptrdiff_t Extent>
-constexpr typename span<ElementType, Extent>::iterator begin(const span<ElementType, Extent> &span) noexcept
+template <class ElementType, std::size_t Extent>
+constexpr typename span<ElementType, Extent>::iterator
+begin(const span<ElementType, Extent>& span) noexcept
```
StephanTLavavej: Pre-existing: the type/variable shadowing here is undesirable (to me), ditto below.

include/gsl/span
JordanMaples: will change that.

include/gsl/string_span
```diff
@@ -108,18 +108,20 @@ namespace details
 // Will fail-fast if sentinel cannot be found before max elements are examined.
 //
 template <typename T, const T Sentinel>
-span<T, dynamic_extent> ensure_sentinel(T* seq, std::ptrdiff_t max = PTRDIFF_MAX)
+span<T, dynamic_extent> ensure_sentinel(T* seq,
+                                        std::size_t max = std::numeric_limits<std::size_t>::max())
```
StephanTLavavej: Pre-existing: I don't see `<limits>` included for `numeric_limits`, and this call to `max` will be stomped over by `<Windows.h>` macros. I recommend using `static_cast<std::size_t>(-1)` here. Ditto below.

include/gsl/span
JordanMaples: I will leave it then, thanks!

include/gsl/span
JordanMaples: Why yes... yes they do. Will fix that. Thanks

include/gsl/string_span
```diff
@@ -312,83 +316,84 @@ public:
...
     return {view.data(), narrow_cast<std::size_t>(view.length())};
 }
 
-template <class ElementType, std::ptrdiff_t Extent>
+template <class ElementType, std::size_t Extent>
 basic_string_span<const byte, details::calculate_byte_size<ElementType, Extent>::value>
 as_bytes(basic_string_span<ElementType, Extent> s) noexcept
 {
-    GSL_SUPPRESS(type.1) // NO-FORMAT: attribute
+    GSL_SUPPRESS(type .1) // NO-FORMAT: attribute
```
StephanTLavavej: Is a space acceptable here? Occurs below.

tests/span_tests.cpp
```diff
@@ -158,27 +153,27 @@ TEST(span_test, from_pointer_length_constructor)
...
```
StephanTLavavej: Technically you've included `<stddef.h>` for `size_t` but not `<cstddef>` for `std::size_t`.

include/gsl/span
JordanMaples: Good catch.

include/gsl/span
JordanMaples: Fixed with Casey's comment. About to push update.

include/gsl/span
JordanMaples: Thanks, I missed that.

include/gsl/span
```diff
@@ -262,22 +289,22 @@ namespace details
...
-            Expects(lhs.span_ == rhs.span_        // range spans have to match
-                    && lhs.index_ <= rhs.index_); // range must not be transposed
+            Expects(lhs.begin_ == rhs.begin_    // range spans have to match
+                    && lhs.end_ <= rhs.end_); // range must not be transposed
         }
 
         constexpr void _Verify_offset(const difference_type n) const noexcept
-        { // test that the iterator *this + n is a valid range in an STL
+        { // test that *this + n is within the span of this iterator STL
             // algorithm call
```
StephanTLavavej: "test that *this + n is within the span of this iterator" makes grammatical sense (although using "span" in any sense other than "gsl::span" is somewhat confusing). However, "STL algorithm call" is a relic of the previous comment.

include/gsl/span
JordanMaples: I must have been on autopilot when i wrote that, removing.

include/gsl/span
JordanMaples: fixing.

include/gsl/span
JordanMaples: removing

include/gsl/span
JordanMaples: removed

include/gsl/string_span
JordanMaples: this got hit with my clang-format pass. Thanks for pointing that out. Fixing.

include/gsl/span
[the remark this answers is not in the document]
JordanMaples: ahh, I've corrected my previous comment.

include/gsl/span
JordanMaples: I'll verify the Expects tomorrow morning.

include/gsl/span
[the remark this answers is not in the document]
sunnychatterjee: Thank you Casey and Stephan for your clarification.

include/gsl/span
```diff
@@ -154,7 +155,7 @@ namespace details
         GSL_SUPPRESS(bounds.1) // NO-FORMAT: attribute
         constexpr reference operator*() const
         {
-            Expects(index_ != span_->size());
+            Expects(static_cast<size_t>(index_) != span_->size());
```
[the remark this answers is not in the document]
JordanMaples: Outdated

tests/span_tests.cpp
JordanMaples: I'll switch this to <cstddef> instead of <stddef.h>, it was indirectly being pulled in through <gsl/span.

include/gsl/span
```diff
@@ -451,45 +454,47 @@ public:
     constexpr span& operator=(const span& other) noexcept = default;
 
     // [span.sub], span subviews
-    template <std::ptrdiff_t Count>
+    template <std::size_t Count>
     constexpr span<element_type, Count> first() const noexcept
     {
-        Expects(Count >= 0 && Count <= size());
+        Expects(Count <= size());
```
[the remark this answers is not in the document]
JordanMaples: If my understanding of the template specialization for extent_type is correct, it should never be in a state where size() returns dynamic_extent. However, I added the check just as an added layer of protection.

include/gsl/span
```diff
@@ -451,45 +454,47 @@ public:
...
         return {data(), Count};
     }
 
-    template <std::ptrdiff_t Count>
+    template <std::size_t Count>
     GSL_SUPPRESS(bounds.1) // NO-FORMAT: attribute
     constexpr span<element_type, Count> last() const noexcept
     {
-        Expects(Count >= 0 && size() - Count >= 0);
+        Expects(size() >= Count);
```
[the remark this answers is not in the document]
JordanMaples: See other instance of this comment.

include/gsl/span
```diff
@@ -351,95 +396,100 @@ namespace details
...
+    using const_iterator = details::span_iterator<const ElementType>;
     using reverse_iterator = std::reverse_iterator<iterator>;
     using const_reverse_iterator = std::reverse_iterator<const_iterator>;
 
-    using size_type = index_type;
+    using index_type [[deprecated("use size_type instead of index_type")]] = size_type;
 
 #if defined(GSL_USE_STATIC_CONSTEXPR_WORKAROUND)
-    static constexpr const index_type extent{Extent};
+    static constexpr const size_type extent{Extent};
```
JordanMaples: remove constexpr from the GSL_USE_STATIC_CONSTEXPR_WORKAROUND

include/gsl/span
```diff
@@ -451,62 +501,74 @@ public:
...
         return make_subspan(size() - count, dynamic_extent, subspan_selector<Extent>{});
     }
 
-    constexpr span<element_type, dynamic_extent> subspan(index_type offset,
-                                                         index_type count = dynamic_extent) const noexcept
+    constexpr span<element_type, dynamic_extent> subspan(size_type offset,
+                                                         size_type count = dynamic_extent) const
+        noexcept
     {
         return make_subspan(offset, count, subspan_selector<Extent>{});
```
JordanMaples: can we remove subspan_selector?

include/gsl/span
```diff
@@ -451,62 +501,74 @@ public:
...
     }
+
     constexpr bool empty() const noexcept { return size() == 0; }
 
     // [span.elem], span element access
+    // clang-format off
     GSL_SUPPRESS(bounds.1) // NO-FORMAT: attribute
-    constexpr reference operator[](index_type idx) const noexcept
+    // clang-format on
+    constexpr reference operator[](size_type idx) const noexcept
```
JordanMaples: should [[nodiscard]] be added? (can do in later pr)

include/gsl/span
```diff
@@ -351,95 +396,100 @@ namespace details
...
 
-    using iterator = details::span_iterator<span<ElementType, Extent>, false>;
-    using const_iterator = details::span_iterator<span<ElementType, Extent>, true>;
+    using iterator = details::span_iterator<ElementType>;
+    using const_iterator = details::span_iterator<const ElementType>;
     using reverse_iterator = std::reverse_iterator<iterator>;
     using const_reverse_iterator = std::reverse_iterator<const_iterator>;
 
-    using size_type = index_type;
+    using index_type [[deprecated("use size_type instead of index_type")]] = size_type;
```
JordanMaples: remove this

include/gsl/span
```diff
@@ -627,71 +677,87 @@ private:
 
     // The rest is needed to remove unnecessary null check
     // in subspans and constructors from arrays
-    constexpr span(KnownNotNull ptr, index_type count) : storage_(ptr, count) {}
+    constexpr span(KnownNotNull ptr, size_type count) : storage_(ptr, count) {}
```
JordanMaples: is this needed?

include/gsl/span
```diff
@@ -627,71 +677,87 @@ private:
...
+template <class Type, std::size_t Size>
+span(const std::array<Type, Size>&)->span<const Type, Size>;
+
+#endif // ( defined(__cpp_deduction_guides) && (__cpp_deduction_guides >= 201611L) )
+
 #if defined(GSL_USE_STATIC_CONSTEXPR_WORKAROUND)
-template <class ElementType, std::ptrdiff_t Extent>
-constexpr const typename span<ElementType, Extent>::index_type span<ElementType, Extent>::extent;
+template <class ElementType, std::size_t Extent>
+constexpr const typename span<ElementType, Extent>::size_type span<ElementType, Extent>::extent;
```
JordanMaples: remove constexpr from here

include/gsl/span
```diff
@@ -705,45 +771,57 @@ namespace details
...
+    GSL_SUPPRESS(type.1) // NO-FORMAT: attribute
+    // clang-format on
+    return {reinterpret_cast<byte*>(s.data()), s.size_bytes()};
+}
+
+template <class ElementType, std::size_t Extent,
+          std::enable_if_t<!std::is_const<ElementType>::value, int> = 0>
+[[deprecated(
+    "use as_writable_bytes")]] span<byte, details::calculate_byte_size<ElementType, Extent>::value>
 as_writeable_bytes(span<ElementType, Extent> s) noexcept
```
JordanMaples: just remove it

include/gsl/span
```diff
@@ -785,67 +863,76 @@ constexpr span<typename Ptr::element_type> make_span(Ptr& cont)
...
 
 // Specialization of gsl::at for span
-template <class ElementType, std::ptrdiff_t Extent>
+template <class ElementType, std::size_t Extent>
 constexpr ElementType& at(span<ElementType, Extent> s, index i)
 {
     // No bounds checking here because it is done in span::operator[] called below
-    return s[i];
+    Ensures(i >= 0);
+    return s[narrow_cast<std::size_t>(i)];
```
JordanMaples: call static_cast directly.

tests/span_tests.cpp
```diff
@@ -110,17 +105,17 @@ TEST(span_test, from_nullptr_size_constructor)
         std::abort();
     });
     {
-        span<int> s{nullptr, narrow_cast<span<int>::index_type>(0)};
+        span<int> s{nullptr, narrow_cast<span<int>::size_type>(0)};
```
JordanMaples: can we get rid of the narrow_cast here?

include/gsl/span
JordanMaples: we should probably keep this for perf gains from not checking the nullness of the pointers.

tests/span_tests.cpp
JordanMaples: the cast itself will have to stay, but maybe we should just use static_cast instead of narrow_cast.

JordanMaples: for clarity, I'm going to leave the narrow_cast

include/gsl/span
JordanMaples: When GSL_USE_STATIC_CONSTEXPR_WORKAROUND is defined in MSVC constexpr is pushed and redefined as /*constexpr*/, however that is not the case for clang/apple-clang/or g++.  So I'll leave these in.

include/gsl/span
JordanMaples: We should leave the subspan_selector in. It is being used to determine which version of make_subspan is called. In the known extent case eliminates checks that are performed in the dynamic_extent case.

include/gsl/span
```diff
@@ -287,61 +327,66 @@ namespace details
...
-        static_assert(Ext >= 0, "A fixed-size span must be >= 0 in size.");
+        static_assert(Ext != dynamic_extent, "A fixed-size span must be >= 0 in size.");
 
         constexpr extent_type() noexcept {}
 
-        template <index_type Other>
+        template <size_type Other>
         constexpr extent_type(extent_type<Other> ext)
         {
             static_assert(Other == Ext || Other == dynamic_extent,
```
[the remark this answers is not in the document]
JordanMaples: tests still pass with it removed. Pushing to see if everything passes in Travis and Appveyor.

JordanMaples:
@StephanTLavavej @BillyONeal @CaseyCarter Would you mind taking another look at this? I'm planning on creating another PR after this one closes to move all of the span comparison operators and everything that doesn't align to the standard into another file.

tests/utils_tests.cpp
```diff
@@ -23,6 +23,7 @@
 #include <limits>      // for numeric_limits
 #include <stdint.h>    // for uint32_t, int32_t
 #include <type_traits> // for is_same
+#include <cstddef>     // for std::size_t
```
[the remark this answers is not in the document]
JordanMaples: Thanks, I missed switching that back. (I switched it for a change that I have since removed)

include/gsl/span
```diff
@@ -351,95 +392,98 @@ namespace details
...
-    {
-    }
-
-    constexpr span(const std::array<std::remove_const_t<element_type>, 0>&) noexcept
-        : storage_(static_cast<pointer>(nullptr), details::extent_type<0>())
-    {
-    }
+    {}
 
     // NB: the SFINAE here uses .data() as a incomplete/imperfect proxy for the requirement
```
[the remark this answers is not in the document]
JordanMaples: This codebase has so many typos in it, thanks for pointing it out. Fixed it.

include/gsl/span
```diff
@@ -627,71 +671,87 @@ private:
 
     // The rest is needed to remove unnecessary null check
     // in subspans and constructors from arrays
-    constexpr span(KnownNotNull ptr, index_type count) : storage_(ptr, count) {}
+    constexpr span(KnownNotNull ptr, size_type count) : storage_(ptr, count) {}
```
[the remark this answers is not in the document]
JordanMaples: Good catch, I'll add `noexcept` to these

include/gsl/span
```diff
@@ -351,95 +392,98 @@ namespace details
...
     // NB: the SFINAE here uses .data() as a incomplete/imperfect proxy for the requirement
     // on Container to be a contiguous sequence container.
     template <class Container,
               class = std::enable_if_t<
                   !details::is_span<Container>::value && !details::is_std_array<Container>::value &&
                   std::is_convertible<typename Container::pointer, pointer>::value &&
-                  std::is_convertible<typename Container::pointer,
-                                      decltype(std::declval<Container>().data())>::value>>
-    constexpr span(Container& cont) noexcept : span(cont.data(), narrow<index_type>(cont.size()))
+                  std::is_convertible<pointer, typename Container::pointer>::value &&
```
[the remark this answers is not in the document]
JordanMaples: it doesn't look like it is required. removing it.

include/gsl/span
```diff
@@ -104,156 +105,187 @@ namespace details
...
-            return *(span_->data() + index_);
+            return {begin_, end_, current_};
         }
 
-        constexpr pointer operator->() const
+        constexpr reference operator*() const noexcept
         {
-            Expects(index_ != span_->size());
-            return span_->data() + index_;
+            Expects(begin_ && current_ && end_);
```
BillyONeal: The following test lets you delete the current_ test. (If begin and end are not null and current is in between, then current can't be null either)

Occurs below.

include/gsl/span
```diff
@@ -351,95 +392,97 @@ namespace details
...
-    constexpr span(pointer ptr, index_type count) noexcept: storage_(ptr, count) {}
+    constexpr span(pointer ptr, size_type count) noexcept : storage_(ptr, count)
+    {
+        if (Extent != dynamic_extent) Expects(count == Extent);
+    }
 
     constexpr span(pointer firstElem, pointer lastElem) noexcept
-        : storage_(firstElem, std::distance(firstElem, lastElem))
-    {}
+        : storage_(firstElem, static_cast<std::size_t>(std::distance(firstElem, lastElem)))
```
BillyONeal: I think this call to distance when you know the inputs are pointers is an unnecessary debug pessimization. Use `lastElem - firstElem` instead?

include/gsl/span
```diff
@@ -287,61 +323,66 @@ namespace details
...
             // after _Unwrapped
-            index_ = p - span_->data();
+            current_ = p;
         }
 #endif
 
-    protected:
-        const Span* span_ = nullptr;
-        std::ptrdiff_t index_ = 0;
+        pointer begin_ = nullptr;
```
BillyONeal: Is the intent for these to be public?

include/gsl/span
```diff
@@ -104,156 +105,187 @@ namespace details
...
+        }
 
-        constexpr friend bool operator==(span_iterator lhs, span_iterator rhs) noexcept
+        template <
+            class Type2,
+            std::enable_if_t<std::is_same<std::remove_cv_t<Type2>, value_type>::value, int> = 0>
+        constexpr bool operator==(const span_iterator<Type2>& rhs) const noexcept
         {
-            return lhs.span_ == rhs.span_ && lhs.index_ == rhs.index_;
+            Expects(begin_ == rhs.begin_ && end_ == rhs.end_);
```
BillyONeal: This seems an unreasonable `Expects`? Why are you expecting that the spans are equal in order to test if they are equal?

BillyONeal: Oh because this is the iterator and Bill is dumb. :)

include/gsl/span
BillyONeal: (Probably occurs elsewhere)

include/gsl/span
```diff
@@ -451,62 +494,74 @@ public:
     constexpr span& operator=(const span& other) noexcept = default;
 
     // [span.sub], span subviews
-    template <std::ptrdiff_t Count>
+    template <std::size_t Count>
     constexpr span<element_type, Count> first() const noexcept
     {
-        Expects(Count >= 0 && Count <= size());
+        Expects(Count != dynamic_extent);
+        Expects(Count <= size());
```
BillyONeal: `Count <= size()` already tests that it isn't `dynamic_extent` because `dynamic_excent` is `SIZE_MAX`.

BillyONeal: (Occurs below)

include/gsl/span
```diff
@@ -451,62 +494,74 @@ public:
...
 
     // [span.obs], span observers
-    constexpr index_type size() const noexcept { return storage_.size(); }
-    constexpr index_type size_bytes() const noexcept
+    constexpr size_type size() const noexcept { return storage_.size(); }
+
+    constexpr size_type size_bytes() const noexcept
     {
-        return size() * narrow_cast<index_type>(sizeof(element_type));
+        Expects(size() < dynamic_extent / sizeof(element_type));
```
BillyONeal: If this Expects is an invariant it seems like it should be on the constructors rather than here? After all, if this condition ever happens that the span isn't meeting the "is a range of N objects of type T" invariant.

include/gsl/span
```diff
@@ -521,23 +576,35 @@ public:
...
-    constexpr iterator end() const noexcept { return {this, size()}; }
+    constexpr iterator begin() const noexcept
+    {
+        const auto data = storage_.data();
+        return {data, data + size(), data};
+    }
+
+    constexpr iterator end() const noexcept
+    {
+        const auto data = storage_.data();
```
BillyONeal: Super debug perf nitpick: How about:

```
        const auto data = storage_.data();
        const auto lastData = data + storage_.size();
        return {data, lastData, lastData};
```

JordanMaples: good call, will use this.

include/gsl/span
JordanMaples: makes sense, updating.

include/gsl/span
JordanMaples: will do.

include/gsl/span
```diff
@@ -785,67 +844,76 @@ constexpr span<typename Ptr::element_type> make_span(Ptr& cont)
 }
 
 // Specialization of gsl::at for span
-template <class ElementType, std::ptrdiff_t Extent>
+template <class ElementType, std::size_t Extent>
 constexpr ElementType& at(span<ElementType, Extent> s, index i)
 {
     // No bounds checking here because it is done in span::operator[] called below
-    return s[i];
+    Ensures(i >= 0);
```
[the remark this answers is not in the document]
JordanMaples: No idea, expects makes more sense.

include/gsl/span
```diff
@@ -705,45 +764,45 @@ namespace details
...
     // constexpr
     // and so will fail compilation of the template
-    template <class ElementType, std::ptrdiff_t Extent>
-    struct calculate_byte_size
-        : std::integral_constant<std::ptrdiff_t,
-                                 static_cast<std::ptrdiff_t>(sizeof(ElementType) *
-                                                             static_cast<std::size_t>(Extent))>
+    template <class ElementType, std::size_t Extent>
+    struct calculate_byte_size : std::integral_constant<std::size_t, sizeof(ElementType) * Extent>
     {
```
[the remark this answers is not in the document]
JordanMaples: will add that.

include/gsl/span
```diff
@@ -351,95 +392,97 @@ namespace details
...
+    // NB: the SFINAE here uses .data() as an incomplete/imperfect proxy for the requirement
     // on Container to be a contiguous sequence container.
     template <class Container,
               class = std::enable_if_t<
                   !details::is_span<Container>::value && !details::is_std_array<Container>::value &&
                   std::is_convertible<typename Container::pointer, pointer>::value &&
-                  std::is_convertible<typename Container::pointer,
-                                      decltype(std::declval<Container>().data())>::value>>
-    constexpr span(Container& cont) noexcept : span(cont.data(), narrow<index_type>(cont.size()))
+                  std::is_convertible<decltype(std::declval<Container&>().data()), pointer>::value>>
```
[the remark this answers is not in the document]
JordanMaples: thanks!

include/gsl/span
```diff
@@ -451,62 +494,74 @@ public:
...
         return {data(), Count};
     }
 
-    template <std::ptrdiff_t Count>
+    template <std::size_t Count>
+    // clang-format off
     GSL_SUPPRESS(bounds.1) // NO-FORMAT: attribute
-    constexpr span<element_type, Count> last() const noexcept
+        // clang-format on
+        constexpr span<element_type, Count> last() const noexcept
```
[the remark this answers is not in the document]
JordanMaples: nope, unintended. fixed.

include/gsl/span
JordanMaples: These will be moved in a different pr.

include/gsl/span
```diff
@@ -785,67 +844,76 @@ constexpr span<typename Ptr::element_type> make_span(Ptr& cont)
 }
 
 // Specialization of gsl::at for span
```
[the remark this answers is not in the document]
JordanMaples: removed incorrect comment.

JordanMaples:
The only failure in Travis has to do with XCode 9.1, the error is that it is outdated. This will be addressed in another PR that cleans up the Travis matrix.

include/gsl/span
```diff
@@ -287,61 +323,66 @@ namespace details
...
-    template <std::ptrdiff_t Ext>
+    template <std::size_t Ext>
     class extent_type
     {
     public:
-        using index_type = std::ptrdiff_t;
+        using size_type = std::size_t;
 
-        static_assert(Ext >= 0, "A fixed-size span must be >= 0 in size.");
+        static_assert(Ext != dynamic_extent, "A fixed-size span must not have size == dynamic_extent");
```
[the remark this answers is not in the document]
JordanMaples: I'll address this in my next PR.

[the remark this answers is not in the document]
JordanMaples: (unless there are other changes that need to go in right now)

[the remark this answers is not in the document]
JordanMaples: nvm, fixed already

include/gsl/span
```diff
@@ -121,139 +121,171 @@ namespace details
...
+            if (n < 0) Expects(current_ - begin_ >= -n);
+            current_ += n;
+            return *this;
+        }
+
+        constexpr span_iterator operator+(const difference_type n) const noexcept
         {
-            auto ret = *this;
+            span_iterator ret = *this;
             return ret += n;
```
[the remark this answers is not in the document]
JordanMaples: Addressed it. It should take advantage of NRVO now.

include/gsl/string_span
```diff
@@ -108,18 +108,20 @@ namespace details
...
 template <typename T, const T Sentinel>
-span<T, dynamic_extent> ensure_sentinel(T* seq, std::ptrdiff_t max = PTRDIFF_MAX)
+span<T, dynamic_extent> ensure_sentinel(T* seq,
+                                        std::size_t max = static_cast<std::size_t>(-1))
 {
     Ensures(seq != nullptr);
 
-    GSL_SUPPRESS(f.23) // NO-FORMAT: attribute // TODO: false positive // TODO: suppress does not work
+    GSL_SUPPRESS(
+        f.23) // NO-FORMAT: attribute // TODO: false positive // TODO: suppress does not work
```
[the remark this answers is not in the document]
JordanMaples: Should be fine for now. I need to verify that these are still needed.

KindDragon:
Maybe you can release a new version with these changes?

PR #845, opened 2020-02-07
Fix incorrect package name for vcpkg

cbuchart:
The package indicated in README.md corresponded to the GNU Scientific Library https://www.gnu.org/software/gsl/


JordanMaples:
Thanks for the fix!

PR #847, opened 2020-02-19
span file pruning

JordanMaples:
General cleanup / pruning. When span was standardized the span based comparison operators were removed from the definition. https://wg21.link/P1085R2. 

This PR's content:
- Take the comparison operators, the [cr](begin|end), make_span, ssize, and gsl::at support and put them into a separate file named span_ext. 
- The tests that relate to these functions are being moved to their own file. 
- string_span takes the span_ext dependency to maintain the comparison operator support.


JordanMaples:
Here's the follow up PR that I mentioned in #844 to remove the comparison operators and make_span calls. @stephantlavavej.

include/gsl/span
```diff
@@ -21,13 +21,27 @@
 #include <gsl/gsl_byte>   // for byte
 #include <gsl/gsl_util>   // for narrow_cast, narrow
 
-#include <algorithm> // for lexicographical_compare
```
sunnychatterjee: Looks like this PR is addressing two sets of changes:
1. Moving non-standard free functions outside of the span header into its own separate header.
2. Changing the span implementation itself by forward declaring std::array.

Unless there's a technical justification, could you please create separate PRs for the two sets of changes?

include/gsl/span
```diff
@@ -21,13 +21,27 @@
...
-#include <array>     // for array
```
sunnychatterjee: Great to see the includes removed. Nice work @JordanMaples!

JordanMaples: Thanks!

include/gsl/span
JordanMaples: I saw the elimination of the span header as a necessary aspect of cleaning the span header itself. To me, it made sense to have the efforts be combined into a single PR, however you're right that they're not technically intertwined. So I've reverted the array header removal and have set that as #849

JordanMaples: @sunnychatterjee I've abandoned the array removal option for now. I need to investigate the UB associated with the forward decl a bit more before committing to one solution or another.

PR #849, opened 2020-02-25
forward declare array to eliminate header include in span

JordanMaples:


ericLemanissier:
Isn't this undefined behavior?

JordanMaples:
@ericLemanissier I discussed this with @stephantlavavej and yes, you're absolutely correct. This is UB and is technically forbidden. I was hoping to mimic std::span's forward decl of array to avoid unnecessarily pulling in array.  Instead relying on the end user's to pull in the dependency should it be needed. This goal was to avoid unnecessary header bloat.

I'll abandon this PR.

PR #850, opened 2020-02-28
zstring_span: fix for Expects, simplify functions

beinhaerter:
Fixes for a few issues recognized while answering https://stackoverflow.com/questions/56874532/using-gslzstring-view-with-c-apis.

- `s[s.size() - 1]` is wrong for empty `s`, so `Expects(s.size() > 0)`
- no hard coded `'\0'`but `value_type{}`
- hard code `empty()` to return `false`
- simplify `as_string_span`: can never be `empty`
- clarify comment on `ensure_z`

### DIFF
--- a/include/gsl/string_span
+++ b/include/gsl/string_span
@@ -120,9 +120,8 @@ span<T, dynamic_extent> ensure_sentinel(T* seq,
 }
 
 //
-// ensure_z - creates a span for a zero terminated strings.
-// Will fail fast if a null-terminator cannot be found before
-// the limit of size_type.
+// ensure_z - creates a span for a zero terminated strings. The span will not contain the zero termination.
+// Will fail fast if a null-terminator cannot be found before the limit of size_type.
 //
 template <typename CharT>
 span<CharT, dynamic_extent> ensure_z(CharT* const& sz,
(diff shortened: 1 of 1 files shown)


JordanMaples:
Maintainers' call: The change looks good but would you mind rebasing this so I can merge it?

beinhaerter:
@JordanMaples Did a rebase. When looking again at the code I am am (again) scratching my head. I did not change the result of `empty`, but is `empty` exactly what it should be? When a client of GSL has a `basic_zstring_span s` and calls `if (s.empty())` what did he mean to do? Surely not getting a constant `false`. And probably not trying to check if the `basic_zstring_span` does not contain even a terminating null byte - because that is what the class is about and is checking. He probably meant to check if the underlying string is empty. So in my opinion `empty` should either return `span_.size() > 1` or be removed. Such a behaviour change should be addressed in a later PR.

JordanMaples:
@beinhaerter You're right, `empty` in its current state doesn't really make much sense. We can definitely address `empty` in a later PR though. I'll merge the changes now and I'll let you know what the other maintainers think next week after our next meeting. Thanks!

PR #861, opened 2020-04-14
implement LWG3320

JordanMaples:
[LWG3320](https://cplusplus.github.io/LWG/issue3320 ) removes const_iterator and const_reverse_iterator from span. 

This change removes all references to the removed elements.

### DIFF
--- a/include/gsl/span_ext
+++ b/include/gsl/span_ext
@@ -152,20 +152,6 @@ end(const span<ElementType, Extent>& s) noexcept
 }
 
 template <class ElementType, std::size_t Extent>
-constexpr typename span<ElementType, Extent>::const_iterator
-cbegin(const span<ElementType, Extent>& s) noexcept
-{
-    return s.cbegin();
-}
-
-template <class ElementType, std::size_t Extent>
-constexpr typename span<ElementType, Extent>::const_iterator
-cend(const span<ElementType, Extent>& s) noexcept
-{
-    return s.cend();
-}
-
-template <class ElementType, std::size_t Extent>
 constexpr typename span<ElementType, Extent>::reverse_iterator
 rbegin(const span<ElementType, Extent>& s) noexcept
 {
@@ -177,20 +163,6 @@ constexpr typename span<ElementType, Extent>::reverse_iterator
 rend(const span<ElementType, Extent>& s) noexcept
 {
     return s.rend();
-}
-
-template <class ElementType, std::size_t Extent>
-constexpr typename span<ElementType, Extent>::const_reverse_iterator
-crbegin(const span<ElementType, Extent>& s) noexcept
-{
-    return s.crbegin();
-}
-
-template <class ElementType, std::size_t Extent>
-constexpr typename span<ElementType, Extent>::const_reverse_iterator
-crend(const span<ElementType, Extent>& s) noexcept
-{
-    return s.crend();
 }
 
 } // namespace gsl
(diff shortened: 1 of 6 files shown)


JordanMaples:
@CaseyCarter would you mind taking a quick look at this for me?

include/gsl/span_ext
```diff
@@ -1,198 +1,170 @@
...
-template <class ElementType, std::size_t Extent = dynamic_extent>
-constexpr typename span<ElementType, Extent>::iterator
-end(const span<ElementType, Extent>& s) noexcept
-{
-    return s.end();
-}
-
-template <class ElementType, std::size_t Extent>
-constexpr typename span<ElementType, Extent>::const_iterator
-cbegin(const span<ElementType, Extent>& s) noexcept
```
CaseyCarter, line 156: IIUC, these functions exist to enable `meow(some_span)` syntax in cases where `meow(some_std_span)` would call `std::meow`. If that's the case, this should still exist and should have been returning `s,begin()` (Yes, member and non-member `cbegin` for `std::span` had different behavior. This was much of the motivation for the change.)

include/gsl/span_ext
```diff
@@ -1,198 +1,170 @@
...
-{
-    return s.cbegin();
-}
-
-template <class ElementType, std::size_t Extent>
-constexpr typename span<ElementType, Extent>::const_iterator
-cend(const span<ElementType, Extent>& s) noexcept
-{
-    return s.cend();
```
CaseyCarter, line 165: Ditto "This could stay and return `s.end()`"

include/gsl/span_ext
```diff
@@ -1,198 +1,170 @@
...
-rend(const span<ElementType, Extent>& s) noexcept
-{
-    return s.rend();
-}
-
-template <class ElementType, std::size_t Extent>
-constexpr typename span<ElementType, Extent>::const_reverse_iterator
-crbegin(const span<ElementType, Extent>& s) noexcept
-{
-    return s.crbegin();
```
CaseyCarter, line 186: Ditto "Could stay and return `s.rbegin()`".

include/gsl/span_ext
```diff
@@ -1,198 +1,170 @@
...
-}
-
-template <class ElementType, std::size_t Extent>
-constexpr typename span<ElementType, Extent>::const_reverse_iterator
-crend(const span<ElementType, Extent>& s) noexcept
-{
-    return s.crend();
```
CaseyCarter, line 193: Finally ditto "Could stay and return `s.rend()`".

tests/span_ext_tests.cpp
```diff
@@ -1,360 +1,348 @@
...
-
- TEST(span_ext_test, iterator_free_functions)
- {
-     int a[] = {1, 2, 3, 4};
-     span<int> s{a};
-
-     EXPECT_TRUE((std::is_same<decltype(s.begin()), decltype(begin(s))>::value));
-     EXPECT_TRUE((std::is_same<decltype(s.end()), decltype(end(s))>::value));
-
-     EXPECT_TRUE((std::is_same<decltype(s.cbegin()), decltype(cbegin(s))>::value));
```
CaseyCarter, line 211: If you do bring back the `span_ext` functions as I suggest, these could return in the form `EXPECT_TRUE((std::is_same<decltype(cbegin(s)), decltype(std::cbegin(s))>::value));` etc.

PR #863, opened 2020-04-16
Change install destination CMake Config-file

Farwaykorse:
Removing the final `cmake` from the install destination path for the `Microsoft.GSLConfig.cmake` file.

While updating the vcpkg port https://github.com/microsoft/vcpkg/pull/10831 to use the new functionality added with PR #784 
Vcpkg does not expect the additional `cmake` folder in the export destination path.
`vcpkg install ms-gsl` returns the following hint:
````CMake
The package ms-gsl:x86-windows provides CMake targets:

    find_package(cmake CONFIG REQUIRED)
    target_link_libraries(main PRIVATE Microsoft.GSL::GSL)
````
Note the `cmake`  in `find_package`.

----
This PR solves the issue with vcpkg, but `find_package` works for both situations.
@hanst99, @Adnn Is there a reason or advantage for the additional `cmake` folder?
@ras0219-msft If so, should vcpkg support this?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -89,7 +89,7 @@ install(
     DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
 )
 # Make library importable by other projects
-install(EXPORT Microsoft.GSLConfig NAMESPACE Microsoft.GSL:: DESTINATION share/Microsoft.GSL/cmake)
+install(EXPORT Microsoft.GSLConfig NAMESPACE Microsoft.GSL:: DESTINATION share/Microsoft.GSL)
 export(TARGETS GSL NAMESPACE Microsoft.GSL:: FILE Microsoft.GSLConfig.cmake)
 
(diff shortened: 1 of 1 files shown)


xvitaly:
Your PR is incorrect too. The correct path must be share/cmake/Microsoft.GSL.

Farwaykorse:
@xvitaly Do you have a source for this?
From my experience, CMake works with both the original and my change.
The format `share/Microsoft.GSL` is what is used by vcpkg.

The CMake [documentation for ìnstall()`](https://cmake.org/cmake/help/latest/command/install.html#installing-exports) shows no sample with `share/cmake/*`.

xvitaly:
>  From my experience, CMake works with both the original and my change.

I don't know about vcpkg, but I'm Fedora GNU/Linux distribution maintainer. According to guidelines, all cmake configs must be installed to `$libdir/cmake` (arched) or `$datadir/cmake` (noarch) directories.

Farwaykorse:
Sounds good to me.

PR #864, opened 2020-04-16
adding back free funcs to [c|r|cr]begin/end to span

JordanMaples:


JordanMaples:
This can be re-opened if/when there are complaints that they're missing.

PR #867, opened 2020-04-17
Fixed installation on GNU/Linux distributions

xvitaly:
This PR fixes installation on GNU/Linux distributions (and all other Unixes like BSD).

Cmake will find installed configs from `$libdir/cmake` or `$datadir/cmake` directories.


Farwaykorse:
Confirmed to fix the integration with vcpkg.

JordanMaples:
Thanks for the fix!

PR #868, opened 2020-04-18
Search for GoogleTest via pkg-config first

mymedia2:
I discovered that pkg-config module gtest_main may be used in environments without Internet access instead downloading and building Google Test framework by GSL tests. For example, the [libgtest-dev][1] package contains the needed gtest_main.pc file.

 [1]: https://packages.debian.org/sid/amd64/libgtest-dev/filelist


JordanMaples:
Maintainers' call: I'll verify this offline and merge if everything looks good. Thanks!

Issue #865, opened 2020-04-16
Compilation errors due to unscoped ptrdiff_t


masonboswell:
I get build errors in multi_span due to the 3-4 places where ptrdiff_t is used without the namespace in front of it (std::ptrdiff_t).  It seems like the namespace appears in some uses throughout the file but not others.  Adding these I get no build errors.  This is happening with the 3.0 release from an hour ago as well as before.

JordanMaples:
Our implementation of multi_span and strided_span are deprecated and aren't officially supported anymore. There are a few users who have taken copies of the implementation to maintain themselves. 

However, if the change is as trivial as qualifying the std namespace, please create a PR and I'll merge it (provided the tests all pass).

JordanMaples:
@masonboswell did the change in #874 fix your build issue?

masonboswell:
Yes, thank you!  3.0.1 builds clean for me with no changes.

PR #874, opened 2020-04-22
Minor changes to multi_span

JordanMaples:
1. Fully qualify std::ptrdiff_t to avoid compilation errors as mentioned in #865.
~2. Remove multi_span from <gsl/gsl>.~

~multi_span should need to be explicitly pulled in to the end user's source, rather than being pulled in by the <gsl/gsl> header.~


JordanMaples:
The second item should be removed.

include/gsl/gsl
```diff
@@ -21,7 +21,6 @@
 #include <gsl/gsl_assert>    // Ensures/Expects
 #include <gsl/gsl_byte>      // byte
 #include <gsl/gsl_util>      // finally()/narrow()/narrow_cast()...
```
JordanMaples, line 23:
```suggestion
#include <gsl/gsl_util>      // finally()/narrow()/narrow_cast()...
#include <gsl/multi_span>    // multi_span, strided_span...
```

We should add this back, the deprecation of the types should be sufficient. We can remove this header from <gsl/gsl> with the next major release (v4.0.0)

PR #887, opened 2020-05-19
gsl::span P1976R2 implemenation

JordanMaples:
http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p1976r2.html 

@CaseyCarter would you mind taking a look at this?


JordanMaples:
I'll keep this in draft form until I get clean builds of Appveyor and Travis, I'm guessing there will be some issues that pop up with some of the older compilers.

tests/span_ext_tests.cpp
```diff
@@ -320,7 +320,7 @@ TEST(span_ext_test, make_span_from_array_constructor)
      {
          int arr[] = {1, 2, 3};
 
-         span<int> s1 = {&arr[0], 2}; // shorter
+         span<int> s1 = span<int>(&arr[0], 2); // shorter
```
CaseyCarter, line 323: Yep, some `explicit`s are reversed; this change should be unnecessary.

tests/span_tests.cpp
```diff
@@ -157,15 +157,15 @@ TEST(span_test, from_pointer_length_constructor)
         for (int i = 0; i < 4; ++i)
         {
             {
-                span<int> s = {&arr[0], narrow_cast<std::size_t>(i)};
+                span<int> s = span<int>(&arr[0], narrow_cast<std::size_t>(i));
```
CaseyCarter, line 160: Ditto this should be unnecessary, and on 168 and 681 as well.

include/gsl/span
```diff
@@ -519,14 +540,14 @@ public:
...
-        return {data() + Offset, Count == dynamic_extent ? size() - Offset : Count};
+        using type = typename details::calculate_subspan_type<ElementType, Extent, Offset, Count>::type;
+        return type(data() + Offset, Count == dynamic_extent ? size() - Offset : Count);
     }
 
     constexpr span<element_type, dynamic_extent> first(size_type count) const noexcept
     {
         Expects(count <= size());
-        return {data(), count};
+        return span<element_type>(data(), count);
```
CaseyCarter, line 550: This change should be unnecessary with the `explicit`s fixed.

include/gsl/string_span
```diff
@@ -121,7 +121,7 @@ span<T, dynamic_extent> ensure_sentinel(T* seq,
     GSL_SUPPRESS(bounds.1) // NO-FORMAT: attribute // TODO: suppress does not work
     while (static_cast<std::size_t>(cur - seq) < max && *cur != Sentinel) ++cur;
     Ensures(*cur == Sentinel);
-    return {seq, static_cast<std::size_t>(cur - seq)};
+    return span<T>(seq, static_cast<std::size_t>(cur - seq));
```
CaseyCarter, line 124: Ditto should be unnecessary.

tests/string_span_tests.cpp
```diff
@@ -1,1217 +1,1218 @@
...
-TEST(string_span_tests, as_writable_bytes)
-{
-    wchar_t buf[]{L"qwerty"};
-    wzstring_span<> v(buf);
-    const auto s = v.as_string_span();
-    const auto bs = as_writable_bytes(s);
-    EXPECT_TRUE(static_cast<const void*>(bs.data()) == static_cast<const void*>(s.data()));
-    EXPECT_TRUE(bs.size() == s.size_bytes());
-}
+///////////////////////////////////////////////////////////////////////////////
```
CaseyCarter, line 1: I suspect similar unnecessary changes here, but I'm not going to audit all 1200 lines.

JordanMaples: yeah this file is weird. I'm guessing there's an issue with the whitespaces. I'll revert this and adjust everything as necessary

CaseyCarter: There's a 99.99999% chance that you managed to change the line endings from CRLF to LF or vice versa.

JordanMaples: I think I opened it in VS instead of VSCode and the line endings got borked.

include/gsl/span
```diff
@@ -420,18 +420,28 @@ public:
     constexpr span() noexcept : storage_(nullptr, details::extent_type<0>())
     {}
 
+     template<std::size_t extent = Extent, std::enable_if_t<extent != gsl::dynamic_extent, int> = 0>
```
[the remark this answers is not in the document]
JordanMaples: I fixed the indentation issue and renamed `extent` to `SpanExtent`. I still find it kind of strange that I had to pull in `SpanExtent` and that the `Extent` from the class template arg couldn't be used. Please let me know if there's a more elegant solution that I can use here.

tests/string_span_tests.cpp
JordanMaples: Either way, you're right. The file got switched to CRLF for some reason.

include/gsl/span
```diff
@@ -420,18 +420,28 @@ public:
     constexpr span() noexcept : storage_(nullptr, details::extent_type<0>())
     {}
 
-    constexpr span(pointer ptr, size_type count) noexcept : storage_(ptr, count)
+    template<std::size_t SpanExtent = Extent, std::enable_if_t<SpanExtent != gsl::dynamic_extent, int> = 0>
```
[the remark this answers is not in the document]
JordanMaples: Will do, thanks.

include/gsl/span
```diff
@@ -456,40 +466,74 @@ public:
...
     constexpr span(const span& other) noexcept = default;
 
     template <
-        class OtherElementType, std::size_t OtherExtent,
-        class = std::enable_if_t<
-            details::is_allowed_extent_conversion<OtherExtent, Extent>::value &&
-            details::is_allowed_element_type_conversion<OtherElementType, element_type>::value>>
+        class OtherElementType, std::size_t OtherExtent, std::size_t SpanExtent = Extent,
+        std::enable_if_t<
+            (SpanExtent == dynamic_extent || SpanExtent == OtherExtent) &&
```
StephanTLavavej, line 522: I verified the logic for the implicit and explicit constraints here - looks correct to me.

include/gsl/span
```diff
@@ -519,8 +563,8 @@ public:
         typename details::calculate_subspan_type<ElementType, Extent, Offset, Count>::type
     {
         Expects((size() >= Offset) && (Count == dynamic_extent || (Count <= size() - Offset)));
-
-        return {data() + Offset, Count == dynamic_extent ? size() - Offset : Count};
+        using type = typename details::calculate_subspan_type<ElementType, Extent, Offset, Count>::type;
+        return type(data() + Offset, Count == dynamic_extent ? size() - Offset : Count);
```
StephanTLavavej, line 567: Style: The previous lines used braces to construct the returned `span`, but this uses parentheses. Ditto for the following `span` returns.

include/gsl/span
```diff
@@ -731,7 +775,8 @@ as_bytes(span<ElementType, Extent> s) noexcept
     // clang-format off
     GSL_SUPPRESS(type.1) // NO-FORMAT: attribute
     // clang-format on
-    return {reinterpret_cast<const byte*>(s.data()), s.size_bytes()};
+    return span<const byte, details::calculate_byte_size<ElementType, Extent>::value>(
```
StephanTLavavej, line 778: Style: Since you're paying 2 lines anyways, consider `using type =` to simplify the return statement. Ditto `as_writable_bytes`.

tests/span_tests.cpp
```diff
@@ -1180,7 +1180,7 @@ TEST(span_test, from_array_constructor)
  #endif
      {
          auto f = [&]() {
-             const span<int, 4> _s4 = {arr2, 2};
+             span<int, 4> _s4{arr2, 2};
```
StephanTLavavej, line 1183: Why remove top-level `const`? (Doesn't really matter since the variable is discarded, but `s2` above remains `const`.)

include/gsl/span
```diff
@@ -420,18 +420,28 @@ public:
     constexpr span() noexcept : storage_(nullptr, details::extent_type<0>())
     {}
 
-    constexpr span(pointer ptr, size_type count) noexcept : storage_(ptr, count)
+    template<std::size_t SpanExtent = Extent, std::enable_if_t<SpanExtent != gsl::dynamic_extent, int> = 0>
+    constexpr explicit span(pointer ptr, size_type count) noexcept : storage_(ptr, count)
```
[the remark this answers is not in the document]
JordanMaples: Having everything as `constexpr explicit` makes the most sense to me. I'll change them all to this.

include/gsl/span
JordanMaples: Thanks!

include/gsl/span
```diff
@@ -456,40 +466,74 @@ public:
...
     {}
 
     constexpr span(const span& other) noexcept = default;
 
     template <
-        class OtherElementType, std::size_t OtherExtent,
-        class = std::enable_if_t<
-            details::is_allowed_extent_conversion<OtherExtent, Extent>::value &&
-            details::is_allowed_element_type_conversion<OtherElementType, element_type>::value>>
+        class OtherElementType, std::size_t OtherExtent, std::size_t SpanExtent = Extent,
```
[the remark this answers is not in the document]
JordanMaples: Good call, I'll change them all to be MyExtent to be more consistent. Thanks.

include/gsl/span
JordanMaples: Cleaned that up. Thanks.

include/gsl/span
JordanMaples: Good call, swapping that in. Thanks.

tests/span_tests.cpp
JordanMaples: No real reason, I was experimenting with something and forgot to add it back. It's fixed now.

include/gsl/span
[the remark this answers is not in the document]
JordanMaples: It looks like there's a mix of `dynamic_extent` and `gsl::dynamic_extent` all throughout this file.  I'll file an issue and separate PR to make it more cohesive.

JordanMaples:
Thanks again @CaseyCarter and @StephanTLavavej for the code review!

PR #893, opened 2020-06-06
Added ARCH_INDEPENDENT parameter to write_basic_package_version_file()

xvitaly:
Added `ARCH_INDEPENDENT` parameter to `write_basic_package_version_file()`, because this is a architecture agnostic header-only library.

Regression caused by #879.

This PR fixes the following issue:

```
find_package(Microsoft.GSL CONFIG REQUIRED)
```

Result:
```
CMake Error at cmake/external/gsl/CMakeLists.txt:11 (find_package):
  Could not find a configuration file for package "Microsoft.GSL" that is
  compatible with requested version "".
  The following configuration files were considered but not accepted:
    /usr/share/cmake/Microsoft.GSL/Microsoft.GSLConfig.cmake, version: 3.1.0 (32bit)
-- Configuring incomplete, errors occurred!
```

If the library was built on a 32-bit host and then used on 64-bit machines (which is fine for architecture independent libraries), it cannot be found by Cmake.


xvitaly:
Fixed compatibility with ancient Cmake versions. Now it should work fine.

PR #898, opened 2020-07-15
Remove msvc < 1910 checks.

### LINKED ISSUE #18: Minimize or remove uses of pragma push/pop_macro
A VC extension supported by other compilers for compatibility reasons.

JordanMaples:
remove msvc < 1910 checks as those versions are no longer supported.

These are the last of the pragma push_macro / pop_macro combos and will close #18


JordanMaples:
Whoops, accidentally closed instead of merged. I'll wait for another clean build.

PR #906, opened 2020-08-05
Fix  GSL_SUPPRESS definition on Intel C++ Compiler

### LINKED ISSUE #903: GSL_SUPPRESS raises unknown attribute warnings on Intel C++ Compiler
When using Intel C++ Compiler on Windows, I got the following warnings (handled as errors on my project):

```
gsl/span(301): error #3924: attribute namespace "gsl" is unrecognized
          GSL_SUPPRESS(bounds.1) // NO-FORMAT: attribute
          ^

gsl/span(301): remark #1292: unknown attribute "gsl::suppress"
          GSL_SUPPRESS(bounds.1) // NO-FORMAT: attribute
          ^
```

It seems that the problem lies in the `GSL_SUPPRESS()` macro definition:
```c++
// gsl_assert, l. 45-53 (also in gsl_bype):
#if defined(__clang__)
#define GSL_SUPPRESS(x) [[gsl::suppress("x")]]
#else
#if defined(_MSC_VER) 
#define GSL_SUPPRESS(x) [[gsl::suppress(x)]]
#else
#define GSL_SUPPRESS(x)
#endif // _MSC_VER
#endif // __clang__
```
[Intel C++ compiler defines the `_MSC_VER` macro](https://software.intel.com/content/www/us/en/develop/documentation/cpp-compiler-developer-guide-and-reference/top/compiler-reference/macros/additional-predefined-macros.html) on Windows, so it falls into the second `#define` case.

Perhaps it could be changed to something like
```c++
#if defined(_MSC_VER) && ! defined(__INTEL_COMPILER)
#define GSL_SUPPRESS(x) [[gsl::suppress(x)]]
```
to avoid such warnings.

**My test environment:**
- Intel C++ Compiler 19.1.2
- MSVC 19.26

TManhente:
Intel C++ Compiler defines the `_MSC_VER` macro, but it doesn't support `[[gsl::*]]` attributes.

This fix makes `GSL_SUPPRESS` expand to nothing on Intel Compiler, solving the following warnings:
```
gsl/span(301): error #3924: attribute namespace "gsl" is unrecognized
          GSL_SUPPRESS(bounds.1) // NO-FORMAT: attribute
          ^
```

This closes #903.

### DIFF
--- a/include/gsl/gsl_assert
+++ b/include/gsl/gsl_assert
@@ -45,7 +45,7 @@
 #if defined(__clang__)
 #define GSL_SUPPRESS(x) [[gsl::suppress("x")]]
 #else
-#if defined(_MSC_VER)
+#if defined(_MSC_VER) && ! defined(__INTEL_COMPILER)
 #define GSL_SUPPRESS(x) [[gsl::suppress(x)]]
 #else
 #define GSL_SUPPRESS(x)
--- a/include/gsl/gsl_byte
+++ b/include/gsl/gsl_byte
@@ -24,7 +24,7 @@
 #if defined(__clang__)
 #define GSL_SUPPRESS(x) [[gsl::suppress("x")]]
 #else
-#if defined(_MSC_VER)
+#if defined(_MSC_VER) && ! defined(__INTEL_COMPILER)
 #define GSL_SUPPRESS(x) [[gsl::suppress(x)]]
 #else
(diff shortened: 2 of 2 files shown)


JordanMaples:
Maintainers' call: Looks good, I'll merge it as soon as appveyor completes. Thanks!

PR #907, opened 2020-08-12
Finally [[nodiscard]] - Version 1

JordanMaples:
Adding `[[nodiscard]]` to `gsl::finally`. 

MSVC and Clang incorrectly warn when it doesn't understand an attribute, so this version includes suppression of the warning for those compilers.

### DIFF
--- a/include/gsl/gsl_util
+++ b/include/gsl/gsl_util
@@ -42,7 +42,7 @@
 #elif defined(__clang__)
 
 #pragma clang diagnostic push
-#pragma clang diagnostic ignored "c++17-extensions"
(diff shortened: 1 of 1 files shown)


JordanMaples:
The test run shows the following error for VS2017 `MSVC error: error C2429: attribute 'nodiscard' requires compiler flag '/std:c++17'` however, adding suppression for that gives the following error in VS2019: `#pragma warning: warning number '2429' not a valid compiler warning`. I'll come back to this later.

JordanMaples:
Closing in favor of using the GSL_NODISCARD approach from #908.

PR #908, opened 2020-08-12
Finally [[nodiscard]] - Version 2

JordanMaples:
Adding `[[nodiscard]]` to `gsl::finally`. 

This version uses a macro approach with the __has_cpp_attribute feature test.

### DIFF
--- a/include/gsl/gsl_util
+++ b/include/gsl/gsl_util
@@ -32,6 +32,12 @@
 
 #endif            // _MSC_VER
 
+#if __has_cpp_attribute(nodiscard) >= 201603L
+#define GSL_NODISCARD [[nodiscard]]
+#else
+#define GSL_NODISCARD
+#endif
+
 namespace gsl
 {
 //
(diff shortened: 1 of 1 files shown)


JordanMaples:
Added a macro that wraps [[nodiscard]] for supporting compilers. Closes #805

jwakely:
I've tried to read the various related issues and I didn't find a reason for replacing `__has_cpp_attribute` with `__cplusplus >= 201703L` in afe824490ebae9188be0e96ecff3dbc73b37285c

Using `__has_cpp_attribute` (correctly, i.e. after `#ifdef __has_cpp_attribute`) seems the ideal way to check for the attribute being supported.

For all versions of Clang and GCC that support the attribute (i.e. since GCC 7.1.0 and Clang 3.9.0) it's available for earlier `-std` modes, so you could use `[[nodiscard]]` even when `__cplusplus >= 201103L`. Using `__has_cpp_attribute` would enable the attribute for C++11 and C++14 users, checking `__cplusplus` obviously doesn't.

Also, I don't know if you care about such an old compiler, but with Clang 4.0.1 `__has_cpp_attribute(nodiscard)` is true, but `__cplusplus >= 201703L` is false (it has the value `201406L` for `-std=c++1z`). So by not using `__has_cpp_attribute` you're missing out on using the attribute where it is actually supported, even for C++17 code (or C++1z anyway).

PR #916, opened 2020-08-13
Update nodiscard macro test

JordanMaples:


JordanMaples:
I'll experiment privately to try making it so we can exclusively rely on __has_cpp_attribute

PR #917, opened 2020-08-15
feat: narrow for non totally ordered types

### LINKED ISSUE #900: feat: narrow for non totally ordered types
The C++ Core Guidelines simply [specify `gsl::narrow`](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#gslutil-utilities) as `narrow // narrow<T>(x) is static_cast<T>(x) if static_cast<T>(x) == x or it throws narrowing_error`. The implementation uses relational operators, so I can't do `gsl::narrow<point<unsigned>>(point<int>{})` for my `point` type.

johelegp:
Resolves #900.

In the original overload, `((t < T{}) != (u < U{}))` is only ever reached when `is_different_signedness` is `true`, which is only ever `true` for arithmetic types because the same applies to `std::is_signed<T>::value` which `is_different_signedness` is initialized from. So we use `std::is_arithmetic<T>::value` to add an overload.

### DIFF
--- a/include/gsl/gsl_narrow
+++ b/include/gsl/gsl_narrow
@@ -25,7 +25,7 @@ struct narrowing_error : public std::exception
 };
 
 // narrow() : a checked version of narrow_cast() that throws if the cast changed the value
-template <class T, class U>
+template <class T, class U, typename std::enable_if<std::is_arithmetic<T>::value>::type* = nullptr>
 GSL_SUPPRESS(type.1) // NO-FORMAT: attribute
 GSL_SUPPRESS(f.6) // NO-FORMAT: attribute // TODO: MSVC /analyze does not recognise noexcept(false)
 constexpr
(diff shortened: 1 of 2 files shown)


hsutter:
Maintainers' call: Thanks for the feature suggestion. Because the GSL design and feature set is owned by the C++ Core Guidelines, this is just an implementation and we follow what the Guidelines specify. So we defer all feature suggestions by opening the issue in the Guidelines repo, and in the meantime we'll close the one here which can then be reopened later when there is guidance to follow from the Guidelines' making a design/feature change. Please reactivate this one when there's a conclusion.

JohelEGP:
> Please reactivate this one when there's a conclusion.

Here's the conclusion: https://github.com/isocpp/CppCoreGuidelines/issues/1726#issuecomment-812109347.

Can this PR be reactivated? It's not clear to me whether this PR makes `narrow` more aligned with the updated C++ Core Guidelines' intent.

JordanMaples:
Yes, please reactivate this. Unfortunately, you'll need to rebase your changes and post a new PR. When we migrated the primary branch from `master` to `main`, we lost the ability to reopen PRs made against `master`. Sorry for the inconvenience.

PR #931, opened 2020-10-01
Deprecate string_span

JordanMaples:
isocpp/cppcoreguidelines#1680 removed string_span from the C++ Core Guidelines. 
This change deprecates `basic_string_span`, `basic_zstring_span`, and all of their derived types.

Edit: as with multi_span and strided_span, string_span will remain deprecated in the GSL for a year before it is removed.


JordanMaples:
Travis-CI is clean: https://travis-ci.org/github/microsoft/GSL/builds/732083875
I'll have to investigate the infra issue after I return from my vacation

PR #937, opened 2020-10-14
eliminate warning, use narrow_cast instead of static_cast

beinhaerter:
The original code triggers

> warning C26472: Don't use a static_cast for arithmetic conversions. Use brace initialization, gsl::narrow_cast or gsl::narrow (type.1).

The change eliminates the warning.

### DIFF
--- a/include/gsl/span
+++ b/include/gsl/span
@@ -434,7 +434,7 @@ public:
 
     template <std::size_t MyExtent = Extent, std::enable_if_t<MyExtent != dynamic_extent, int> = 0>
     constexpr explicit span(pointer firstElem, pointer lastElem) noexcept
-        : storage_(firstElem, gsl::narrow_cast<std::size_t>(lastElem - firstElem))
+        : storage_(firstElem, narrow_cast<std::size_t>(lastElem - firstElem))
     {
         Expects(lastElem - firstElem == static_cast<difference_type>(Extent));
     }
--- a/include/gsl/span
+++ b/include/gsl/span
@@ -19,6 +19,7 @@
 
 #include <gsl/gsl_assert> // for Expects
 #include <gsl/gsl_byte>   // for byte
+#include <gsl/gsl_util>   // for narrow_cast
 
 #include <array>       // for array
 #include <cstddef>     // for ptrdiff_t, size_t, nullptr_t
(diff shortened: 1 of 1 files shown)


beinhaerter:
@JordanMaples Are the unit test for iOS broken? I get

>An error was encountered processing the command (domain=FBSOpenApplicationServiceErrorDomain, code=3):
GSL.no_exception_ensure_tests: -1
The request to open "GSL.no_exception_ensure_tests" failed.

That does not look related to my changes. The other open PR also runs into iOS unit test trouble.

include/gsl/span
```diff
@@ -433,14 +434,14 @@ public:
 
     template <std::size_t MyExtent = Extent, std::enable_if_t<MyExtent != dynamic_extent, int> = 0>
     constexpr explicit span(pointer firstElem, pointer lastElem) noexcept
-        : storage_(firstElem, static_cast<std::size_t>(lastElem - firstElem))
+        : storage_(firstElem, gsl::narrow_cast<std::size_t>(lastElem - firstElem))
```
JordanMaples, line 437: The namespace qualification is unnecessary for narrow_cast. 
```suggestion
        : storage_(firstElem, narrow_cast<std::size_t>(lastElem - firstElem))
```

include/gsl/span
```diff
@@ -433,14 +434,14 @@ public:
...
     {
         Expects(lastElem - firstElem == static_cast<difference_type>(Extent));
     }
 
     template <std::size_t MyExtent = Extent, std::enable_if_t<MyExtent == dynamic_extent, int> = 0>
     constexpr span(pointer firstElem, pointer lastElem) noexcept
-        : storage_(firstElem, static_cast<std::size_t>(lastElem - firstElem))
+        : storage_(firstElem, gsl::narrow_cast<std::size_t>(lastElem - firstElem))
```
JordanMaples, line 444: same here.
```suggestion
        : storage_(firstElem, narrow_cast<std::size_t>(lastElem - firstElem))
```

JordanMaples:
Yeah @beinhaerter, the iOS tests have been failing for a while. I haven't had a chance to look into what's causing it to fail. 
I'm need to look into the Travis-CI integration before I get around to that though.

JordanMaples:
I checked the Travis run and it looks good with the exception of the CMake failure which isn't related to your change.

Info about the Travis failure: 
- CMake fails because GoogleTest requires a deprecated required version of CMake. 
Related issue: google/googletest#3040

PR #939, opened 2020-10-20
Include gsl_narrow from gsl and fix comment.

bowie7070:
narrow() is now in gsl_narrow rather than gsl_util. Include gsl_narrow when exceptions are enabled. Ensures that #include <gsl/gsl> includes entire library as advertised in the documentation.


JordanMaples:
Maintainers' call: Looks good. Thanks!

PR #940, opened 2020-10-21
remove ostream from not_null

### LINKED ISSUE #933: Support STL implementations without ios
I'm working with EASTL as STL implementation, which does not implement ios. `<gsl/pointers>` includes `iosfwd` to be able to define `operator<<`  for `not_null`, so I presently cannot use these features. It would be useful to be able to disable the generation of this operator so that the GSL can work without iostreams.

JordanMaples:
Removing the operator<< from not_null to eliminate dependency on iosfwd.
Closes #933

### DIFF
--- a/include/gsl/pointers
+++ b/include/gsl/pointers
@@ -20,7 +20,7 @@
 #include <gsl/gsl_assert>  // for Ensures, Expects
 
 #include <algorithm>    // for forward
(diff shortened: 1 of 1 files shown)


JordanMaples:
I'm closing this issue in the interests of aligning the `not_null` interface with those of std::unique_ptr and std::shared_ptr, which added `operator<<` in C++20.

PR #946, opened 2020-10-29
Header rename

### LINKED ISSUE #700: Why do some header names contain the gsl_ prefix?
Below is the list of include files provided by this library. Could we remove the gsl_ prefix in front of algorithm, assert, byte and util to have a consistent naming? We already have the gsl prefix because we have to include <gsl/header> so <gsl/gsl_util> is redundant.

gsl
gsl_algorithm -> algorithm
gsl_assert -> assert
gsl_byte -> byte
gsl_util -> util
multi_span
pointers
span
string_span

JordanMaples:
- Renaming main logic files, all files prefixed with `gsl_` are having the `gsl_` dropped. EG. `gsl_algorithm` now is just `algorithm`. 
- The old files prefixed with `gsl_` still exist for backwards compatibility, however they only consist of a message about the impending removal and include a passthrough to the new correct file. 
- Renamed includes in the source files. 
- Ran Clang-Format on all files.

Closes #700 

There were a few instances where the clang-format pass caused issues with the `GSL_SUPPRESS` macro. I've added the required clang-format suppression in those areas to prevent further issues.


mskybert:
This changed:
```
-#include <gsl/multi_span>    // multi_span, strided_span...
-#include <gsl/span>          // span
+#include <gsl/span>        // multi_span, strided_span...
+#include <gsl/span>        // span
```
Ie. removed include of multi_span, which is probably wrong.

JordanMaples:
good catch @mskybert, I'll fix that really quickly. 
The plan is to remove multi_span entirely from the GSL repo before the next release. However, this is not going to be the PR to do it.

JordanMaples:
@mskybert fixed in #954. Thanks again.

PR #950, opened 2020-11-11
suppress bounds.1

beinhaerter:
Suppress "warning C26481: Don't use pointer arithmetic. Use span instead (bounds.1)."


JordanMaples:
Thanks!

PR #953, opened 2020-11-15
Use CMAKE_CXX_STANDARD when available

hdf89shfdfs:
If the use has provided the variable CMAKE_CXX_STANDARD use that instead of providing a default cache variable.

### DIFF
--- a/cmake/guidelineSupportLibrary.cmake
+++ b/cmake/guidelineSupportLibrary.cmake
@@ -1,0 +1,61 @@
+# This cmake module is meant to hold helper functions/macros
+# that make maintaining the cmake build system much easier.
+# This is especially helpful since gsl needs to provide coverage
+# for multiple versions of cmake.
+#
+# Any functions/macros should have a gsl_* prefix to avoid problems
+if (DEFINED guideline_support_library_include_guard)
+    return()
+endif()
+set(guideline_support_library_include_guard ON)
+
+function(gsl_set_default_cxx_standard min_cxx_standard)
+    set(GSL_CXX_STANDARD "${min_cxx_standard}" CACHE STRING "Use c++ standard")
+
+    set(GSL_CXX_STD "cxx_std_${GSL_CXX_STANDARD}")
+
+    if (MSVC)
+        set(GSL_CXX_STD_OPT "-std:c++${GSL_CXX_STANDARD}")
+    else()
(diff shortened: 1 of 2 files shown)


hdf89shfdfs:
This change does a couple of things.

It creates a cmake module to assist in writing helper functions/macros.

Creates 2 new function:
- gsl_set_default_cxx_standard
- gsl_client_set_cxx_standard

gsl_set_default_cxx_standard does exactly what the old code did before.

gsl_client_set_cxx_standard checks if the client set the CMAKE_CXX_STANDARD.

If the client has set CMAKE_CXX_STANDARD then we should avoid doing anything else.

1. It's easier for clients.
2. It's less error prone.
3. Avoid extra configuration time.
4. Avoid unnecessary cache variable overriding.

hdf89shfdfs:
Seems like the IOS builds have been failing for a while. And aren't related to this PR.

JordanMaples:
Yeah, the iOS failures are unrelated. I'm not sure why the Appveyor run exploded like that. 
I'll test the changes locally before merging.

Xazax-hun:
These changes look good to me. In general, I prefer refactoring changes and functional changes separated in different commits, but this is just a nit :)

cmake/guidelineSupportLibrary.cmake
```diff
@@ -0,0 +1,61 @@
...
+endif()
+set(guideline_support_library_include_guard ON)
+
+function(gsl_set_default_cxx_standard min_cxx_standard)
+    set(GSL_CXX_STANDARD "${min_cxx_standard}" CACHE STRING "Use c++ standard")
+
+    set(GSL_CXX_STD "cxx_std_${GSL_CXX_STANDARD}")
+
+    if (MSVC)
+        set(GSL_CXX_STD_OPT "-std:c++${GSL_CXX_STANDARD}")
```
Xazax-hun, line 18: I have a hard time spotting the difference between the two branches here. Maybe time to clean this up?

JordanMaples:
I looked into the Appveyor failure since it's happening across all builds. There's no issue with this PR and as far as I can tell your changes work, so I'm going to merge it. 

In case anyone is interested in the Appveyor failure.
 The issue is that the compiler version that Appveyor is using was updated from Visual Studio v16.7.6 to v16.8.1 which defaults from Clang 10.0.0 to Clang 11.0.0. GoogleTest is emitting a -Wsuggest-override warning in Clang 11.

PR #959, opened 2020-12-04
narrow_cast specialization for floating point -> unsigned types

JordanMaples:
Floating point to unsigned values can have unpredictable conversions across various architectures and build configurations when the floating value is negative. To force a consistent behavior the float should first cast to signed and then to unsigned.

Also ran clang-format and adjusted some weird formatting.

### DIFF
--- a/include/gsl/util
+++ b/include/gsl/util
@@ -88,12 +88,13 @@ finally(F&& f) noexcept
         std::forward<F>(f));
 }
 
-namespace details{
+namespace details
+{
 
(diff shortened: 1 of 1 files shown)


JordanMaples:
Holding off on merging this until the CppCoreGuidelines editors can meet in the new year.

hsutter:
Editors' call: On review, the benefit of doing this is unclear and the suggestions we received from floating point experts is that neither the current nor the proposed behavior is likely "ideal", and that other FP conversion rules are under development. We want to wait for those decisions to be made elsewhere and follow them, not innovate in FP conversions here.

Issue #960, opened 2020-12-12
[cmake] GSL " Found Git: C:/Program Files/Git/cmd/git.exe (found version "2.27.0.windows.1")"


hdf89shfdfs:
https://github.com/microsoft/GSL/blob/a6cef6bc6c6daa4710cf0b6e5bcfd40a3908baa1/CMakeLists.txt#L9

GSL looks for Git even though it isn't always necessary.

```
-- Found Git: C:/Program Files/Git/cmd/git.exe (found version "2.27.0.windows.1") 
```

A minor optimization for configure time would be to only look for Git when testing is enabled. As a small quality of life for users.

It's good practice to minimize configure time for cmake projects.

JordanMaples:
Are you suggesting that the `find_package` line isn't necessary or that it can be relocated to help configuration time?

hdf89shfdfs:
Correct the call can be relocated to help configuration time:

Try using FetchContent (https://cmake.org/cmake/help/latest/module/FetchContent.html) or Cmake Package Manger (https://github.com/TheLartians/CPM.cmake)

```cmake
    CPMAddPackage(
        NAME GSL
        GITHUB_REPOSITORY "Microsoft/GSL"
        VERSION "3.1.0"
        GIT_SHALLOW ON
    )
```

Here is example output from my project:
```
-- Building for: Visual Studio 16 2019
-- The CXX compiler identification is MSVC 19.28.29333.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: C:/Program Files (x86)/Microsoft Visual Studio/2019/Community/VC/Tools/MSVC/14.28.29333/bin/Hostx64/x64/cl.exe - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
--  adding package GSL@3.1.0 (D:/Git/cpm_source_cache/gsl/0746477a58301b687104ac097be39737dc1ceef3)
-- Found Git: C:/Program Files/Git/cmd/git.exe (found version "2.27.0.windows.1") <-----------------------------------------------------
--  adding package magicEnum@0.7.0 (D:/Git/cpm_source_cache/magicenum/3b7b0ee113d60d2ec0297d4c8d1f628a818b67cf)
-- Configuring done
-- Generating done
-- Build files have been written to: D:/Git/PitoUtility/build
```

hdf89shfdfs:
[...]

PR #961, opened 2020-12-15
Minor cmake changes

hdf89shfdfs:
Abstract adding natvis file.

Move call to find_package(Git) as well as making it required/quiet.

This PR address the issues I was trying to address:
https://github.com/microsoft/GSL/issues/960

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -1,7 +1,11 @@
 cmake_minimum_required(VERSION 3.0.2)
 
 project(GSLTests CXX)
+
 include(FindPkgConfig)
+include(ExternalProject)
+
+find_package(Git REQUIRED QUIET)
 
 # will make visual studio generated project group files
 set_property(GLOBAL PROPERTY USE_FOLDERS ON)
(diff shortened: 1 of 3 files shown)


JordanMaples:
Looks good to me, thanks!

tests/CMakeLists.txt
```diff
@@ -1,7 +1,11 @@
 cmake_minimum_required(VERSION 3.0.2)
 
 project(GSLTests CXX)
+
 include(FindPkgConfig)
+include(ExternalProject)
+
+find_package(Git REQUIRED QUIET)
```
mymedia2, line 8: Please do not require Gti for build tests. It is still possible to get GTest through APT or other package manager even if Git is not installed.

hdf89shfdfs: The reason I added the requires was because I thought it was required. In a lot of CMake code this is just left out, hence I thought it was just left out here as well.

Also in my CMake workflow I tend to just download things :)

Feel free to make the pull request.

But add a comment explaining why the requires isn't necessary in the new PR. Commenting is very important in CMake.

Issue #936, opened 2020-10-13
Allow to disable MSVC exception workaround


mtnpke:
gsl_assert automaticaly uses `__fastfail` instead of `std::terminate` if it detects that

* the compiler is MSVC and
* exceptions are disabled (either due to compilation in kernel mode or user mode with explicit disabling).

This is fine for the normal case (which I think is MSVC+Dinkumware), but in my case (MSVC+EASTL in kernel mode), it causes problems. I do want custom code to run on contract violation, which I can do by implementing `std::terminate` in a compatible way. However, I cannot achieve this since the autodetection overrules calling terminate. If I could define `GSL_MSVC_USE_STL_NOEXCEPTION_WORKAROUND` to 0 beforehand and thereby skip the autodetection, this would be very helpful.

carsonRadtke:
We do not plan on allowing the user to customize the failure mode at this time.

Issue #930, opened 2020-09-29
strict_not_null performance issues with reference types


Amaroker:
This is a follow up to https://github.com/microsoft/GSL/issues/396, where @hsutter expressed concern of changes needed to the call site and asked for examples.

To understand my situation better, here's some background information:

It's our company's policy to only use not_null for legacy code / existing APIs that we can't break. However, the first choice for new modern code is always strict_not_null, because we much prefer compile-time errors to run-time errors.

So, the only times we do changes to call sites are during internal refactoring when we strengthen the code with strict_not_null. With this approach I believe Herb's concern doesn't apply.

Consider the following function declaration:

`void foo( const strict_not_null<X*>& p )`

(1). Sometimes we already have a not_null pointer ready to use directly...

`foo( px ); `

...and (2) sometimes we have a stack based value and we need to construct a temporary strict_not_null object. 

`foo( strict_not_null<X*>(&x) );`

However, (2) is not as efficient as it could be. Performance could be improved by introducing a new explicit and noexcept constructor that avoids the null check. It wouldn't break any existing code.

NicolBolas:
So just to be clear, the goal would be to make this work:

    foo(strict_not_null<X*>(x));

But because it's `explicit`, this won't work:

    foo(x);

Correct?

Amaroker:
@NicolBolas
Correct. Still, dangling pointers must be avoided, so rvalue references must not be accepted.

`foo( strict_not_null<X*>(X{}) ); // compiler error`

hsutter:
Thanks!

> However, (2) is not as efficient as it could be. Performance could be improved

Do you have a test case that this null check has measurable overhead?

I'm not trying to be pedantic here... The reason I ask is because this summer I spent several weeks doing an extensive survey and measurement of 20 years' worth of processors, and could not measure any difference for a similar check for `if (this != &rhs)` ... for reasons which appear to apply equally to the `delete nullptr` mandated null check and I think equally for this null check.

I didn't put the paper in a mailing yet because I'm working on a related companion paper that I want to publish at the same time, but for this thread I put a copy in my public misc-papers repo here: https://github.com/hsutter/misc-papers/blob/main/2229%20self%20move%20noop%20data.pdf 

The first Conclusion at the end of the paper was this: "Branches with the three characteristics above — with one alternative empty, virtually always taking the nonempty path, and testing a condition that uses only memory locations that will be used anyway by the nonempty path — should be considered cost-free on all known hardware."

[...]

Issue #877, opened 2020-04-24
gsl::not_null can't be anymore zero cost abstraction.


unixod:
Hi,

Some time ago, until the PR #831 was accepted, GSL had 3 configuration options:

* GSL_TERMINATE_ON_CONTRACT_VIOLATION: std::terminate will be called (default)
* GSL_THROW_ON_CONTRACT_VIOLATION: a gsl::fail_fast exception will be thrown
* GSL_UNENFORCED_ON_CONTRACT_VIOLATION: nothing happens

For some unclear reason, in #831 these options were removed. 

Removal of those options led us to the following not good consequences:
* _Expects(cond)_ can't be configured anymore: now its usage in different parts of GSL always implies runtime checks and calling std::terminate in case of failures, _which in turn implies_ =>
* _gsl::not_null_ allways checks in runtime for nullptr, _which in turn implies_ =>
* _gsl::not_null_ can't be configured to be zero cost abstraction anymore for example in release builds, , _which in turn implies_ =>
* using _gsl::not_null_  in hot parts of code may cause performance issues.

So, my proposal: let's return back the GSL_UNENFORCED_ON_CONTRACT_VIOLATION, or something else which will allow clients to make _Expects(cond)_/_Ensures(cond)_ zero cost.

gdr-at-ms:
@unixod Could you elaborate further on what you meant by _Expects_ and _Ensures_ being zero cost?

unixod:
> @unixod Could you elaborate further on what you meant by _Expects_ and _Ensures_ being zero cost?

Sure, 

I think the easier way to explain my point is to show it in example: https://godbolt.org/z/grfjgv

The following is explanations.

We have a function:
```c++
#include <gsl/gsl>

int foo(gsl::not_null<int*> ptr)
{
    return *ptr;
}
```

and output of two compilation processes (both use -O3 and  -DGSL_UNENFORCED_ON_CONTRACT_VIOLATION): 

| GSL 2.0.0 | GSL 3.0.0 |
| ----- | ------- |
|  <pre>foo(gsl::not_null<int*>):<br>        mov     eax, DWORD PTR [rdi]<br>        ret</pre> | <pre>foo(gsl::not_null<int*>):<br>        test    rdi, rdi<br>        je      .L7<br>        mov     eax, DWORD PTR rdi]<br>        ret<br>.L7:<br>        sub     rsp, 8<br>        call    std::terminate() |

[...]

PR #962, opened 2020-12-17
Expects/Ensures revision

hsutter:
This PR does the following:
   - Removes `__builtin_expect` branch prediction assistance, which in #930 was measured to be useless or a small pessimization.
   - Removes `GSL_ASSUME` which was unused and is not mentioned in the C++ Core Guidelines which defines GSL.
   - Reinstates `GSL_UNENFORCED_ON_CONTRACT_VIOLATION` and `GSL_TERMINATE_ON_CONTRACT_VIOLATION` (the latter is still the default)
   - Replaces `GSL_CONTRACT_CHECK` with a general `contract_group`, and parameterizes `Expects` and `Ensures` with a group.

The last is the main change. Each `contract_group` object wraps one handler, "terminate_handler" style. Making each one an object allows controlling violation handlers separately for different groups of contracts, notably `Bounds` and `Null` checks -- now GSL users who want `Bounds` violations to always fail-fast can enforce that only those violations do always fail-fast, and independently can configure whether `Null` violations should fail-fast. The default violation action for each group is currently set to be `terminate`, but can be configured to do nearly anything by customizing `set_handler` -- for example, `gsl.Bounds.set_handler( []{ my_custom_logger.LogAndContinue("out of bounds access detected"); } );`.

The only restriction on the violation handler is that it must be `noexcept`. This intentionally prevents installing throwing violation handlers, because those are not meaningful in general: Whether a given function wants to write `Expects`/`Ensures` contracts for debugging purposes, and/or the same function can never fail and therefore wants to also be `noexcept` at run time, are orthogonal and independent issues, and in the case of a function that does both it is not useful to throw an exception from an `Expects`/`Ensures` violation because it will just hit the function's `noexcept`, so installing a throwing violation handler is just a fancy way of installing a handler that calls `terminate`.

I've done performance tests at `-O1` and have not been able to measure an overhead to doing this `if (!b) call_handler();` test that is not in the range of instruction alignment issues. That is, in the few tests where I could measure a slight overhead (or speedup!) to inserting the `Expects` or `Ensures` check, I was able to erase the difference by adding a nop (`asm(nop);`), for example here: https://quick-bench.com/q/PqA9EPJqUOwp51iql4NKUDEpsUg . So at least so far it appears there is zero cost to the predictable branch and the only overhead is the extra instructions.

More performance testing is welcome -- in particular, @JordanMaples please try this branch vs. the current implementation in any existing `span` performance tests you have and see if there's a measurable overhead on `span` bounds checking now that the `span` bounds check would invoke a user-customizable violation handler on failure.

This PR intends to address several open issues:
   - #604 
   - #877 
   - #930 
   - #936

### DIFF
--- a/include/gsl/assert
+++ b/include/gsl/assert
@@ -109,12 +109,7 @@ public:
     using handler = void (*)();
 #endif
 
-#if defined(__cpp_constexpr) && __cpp_constexpr >= 201806
-    constexpr contract_group  (handler h)            { set_handler(h); }
-#else // VESTIGIAL, remove when no longer needed for downlevel compilers
     constexpr contract_group  (handler h)            : chandler(h ? h : []()noexcept{}) { }
-#endif
-
 #if __cplusplus >= 202002L
     constexpr auto set_handler(handler h) -> handler { return std::exchange(chandler, h ? h : []()noexcept{}); }
 #else // VESTIGIAL, remove when no longer needed for downlevel compilers
--- a/include/gsl/assert
+++ b/include/gsl/assert
@@ -102,9 +102,11 @@ namespace details
 
 class contract_group {
 public:
-    using handler = void (*)() /*noexcept*/;    // TODO: make this noexcept once all supported compilers allow it
+    // TODO: change this back to unconditionally noexcept
+    //       again once all supported compilers allow it
+    using handler = void (*)() /*noexcept*/;
 
-    constexpr contract_group(handler h)    { set_handler(h); }
+    contract_group  (handler h)            { set_handler(h); }
     auto set_handler(handler h) -> handler { if (!h) h = []()noexcept{};
                                              return chandler.exchange(h); }
     auto get_handler()          -> handler { return chandler; }
@@ -117,6 +119,8 @@ private:
     std::atomic<handler> chandler;
 };
 
+// TODO: change this back to "auto static Blah = contract_group{ ... };"
+//       again once all supported compilers allow it
 static contract_group Default { &gsl::details::terminate };
 static contract_group Bounds  { Default.get_handler() };
 static contract_group Null    { Default.get_handler() };
--- a/include/gsl/assert
+++ b/include/gsl/assert
@@ -109,7 +109,7 @@ public:
     using handler = void (*)();
 #endif
 
-    constexpr contract_group  (handler h)            : chandler(sanitize(h)) { }
+    constexpr contract_group  (handler h = nullptr)  : chandler(sanitize(h)) { }
 #if __cplusplus >= 202002L
     constexpr auto set_handler(handler h) -> handler { return std::exchange(chandler, sanitize(h)); }
 #else // VESTIGIAL, remove when no longer needed for downlevel compilers
@@ -125,7 +125,14 @@ private:
     handler chandler;
 };
 
-auto static Default = contract_group( &gsl::details::terminate );
+auto static Default = contract_group( 
+#if defined GSL_TERMINATE_ON_CONTRACT_VIOLATION
+                                      &gsl::details::terminate
+#else // if defined GSL_UNENFORCED_ON_CONTRACT_VIOLATION
+                                      // use default == null handler
+#endif
+                                    );
+
 auto static Bounds  = contract_group( Default.get_handler() );
 auto static Null    = contract_group( Default.get_handler() );
 auto static Testing = contract_group( Default.get_handler() );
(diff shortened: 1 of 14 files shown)


include/gsl/assert
```diff
@@ -120,13 +97,47 @@ namespace details
...
+#endif
+                                    );
+auto GSL_CONTRACT_VIOLATION_GRANULARITY Bounds  = Default;
+auto GSL_CONTRACT_VIOLATION_GRANULARITY Null    = Default;
+auto GSL_CONTRACT_VIOLATION_GRANULARITY Testing = Default;
+
+} // namespace gsl
+
+#define Expects(cond, kind) kind.expects(cond)
+#define Ensures(cond, kind) kind.ensures(cond)
```
JordanMaples, line 140: Do we want to add an `Expects(cond)` to prevent outright breaking users who have Expects/Ensures in their code or are we willing to make this a breaking change?

mbeutel:
Is this extension of `Expects()`/`Ensures()` backed by a change in the Core Guidelines, or are there plans for such a change?

hsutter:
Before this PR is accepted, it needs to be reviewed by the Guidelines editors.

Summarizing the current status and plan: Right now, we (Microsoft) are first doing a performance test in some internal code that heavily uses GSL (particularly but not only gsl::span) to validate that it doesn't cause measurable perf overhead. If that efficiency is not validated, there's no point in bothering the Guidelines editors with a non-starter and we should just reject this PR. If it is validated, then the next step is to have the design reviewed by the Guidelines editors and see if they agree, agree with changes or other feedback, or reject it.

mbeutel:
Thank you, that is helpful context.

Why do you want to make contract enforcement strictly a runtime decision instead of sticking with the traditional preprocessor-based approach (conditional definition of `Expects()`/`Ensures()`)? I understand you want to support different handlers for different contract groups; but that could also be done with the preprocessor.

hsutter:
@mbeutel Sorry for the delay in replying. Re this:

> Why do you want to make contract enforcement strictly a runtime decision instead of sticking with the traditional preprocessor-based approach (conditional definition of `Expects()`/`Ensures()`)?

We have some customers who want one or both of:

1) Be able to use `span` in different parts of the same program, some of which have checking on and checking off, without ODR violations. `span` does it checking via `Expects`, and having macros change the definition of `Expects` in a single `span` implementation creates ODR issues. We considered other approaches, such as having checked and unchecked `span` types, but then that defeats composability for a vocabulary types.

2) Be able to guarantee certain uses of `Expects` are always checked (notably, to guarantee that bounds checks are always performed and violations always fail-fast), regardless of whether the same program contains other uses of `Expects` that may not be checked or may be checked but handled differently.

mbeutel:
@hsutter No problem, thanks for taking the time to explain.

> Be able to use `span` in different parts of the same program, some of which have checking on and checking off

How would that work? How would `gsl::span<>` know that checks are desired in the current part of the program? The only thing that comes to mind is global state, i.e. you install a contract violation handler when entering the part of the program that desires checks and you restore the previous handler upon leaving.

> We considered other approaches, such as having checked and unchecked `span` types

(...e.g. `gsl::span<>` and `std::span<>`...)

> that defeats composability for a vocabulary types

Could you elaborate on that concern? I understand that bifurcation of vocabulary types can be problematic, the prime example being `std::vector<>` and `std::pmr::vector<>`. But spans are mostly used at function interfaces, and different span types *should* be constructible from each other through the contiguous-range constructor.

> Be able to guarantee certain uses of Expects are always checked (notably, to guarantee that bounds checks are always performed and violations always fail-fast), regardless of whether the same program contains other uses of Expects that may not be checked or may be checked but handled differently.

I agree that it would be desirable to have some granularity in the enforcement of precondition/postcondition checks, and I like the idea of contract groups.

Sometimes it's easier to assess how expensive and how useful a check will be than which group it would belong to. Some of the Contracts proposals had floated the idea of audit-level checks, which had inspired me to add different severity-level flavors of `Expects()` to [*gsl-lite*](https://github.com/gsl-lite/gsl-lite):

- `gsl_Expects()`: on by default
- `gsl_ExpectsDebug()`: like `assert()`, this is not evaluated if `NDEBUG` is defined
- `gsl_ExpectsAudit()`: off by default

Contract groups could be orthogonal to severity, and I can see how configuring different violation behaviors for different contract groups might be useful. E.g. one might want to terminate on safety-related violations but only emit a log message for violations of program logic.

PR #963, opened 2020-12-26
Improve build script for standalone tests

mymedia2:
* Do not require Git for build tests

   Cloning via Git is not the only path to obtain Google Test framework. In Linux, using a system package manager is the preferred way, and gtest can be installed through APK, APT, DNF, pacman, or many other.

   See also: https://github.com/microsoft/GSL/pull/961#discussion_r548959056

 * Support standalone tests

    The patch makes possible to run auto-tests against globally installed GSL, ignoring local headers. To do this, run CMake in the tests/ folder.

        cmake -B build -S tests -DGSL_CXX_STANDARD=14

    This feature should not break existing build recipes.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -5,13 +5,14 @@ project(GSLTests CXX)
 include(FindPkgConfig)
 include(ExternalProject)
 
-find_package(Git REQUIRED QUIET)
-
 # will make visual studio generated project group files
 set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 
 pkg_search_module(GTestMain gtest_main)
 if (NOT GTestMain_FOUND)
+    # No pre-installed GTest is available, try to download it using Git.
+    find_package(Git REQUIRED QUIET)
+
     configure_file(CMakeLists.txt.in googletest-download/CMakeLists.txt)
     execute_process(
         COMMAND ${CMAKE_COMMAND} -G "${CMAKE_GENERATOR}" .
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -1,6 +1,7 @@
 cmake_minimum_required(VERSION 3.0.2)
 
 project(GSLTests CXX)
+enable_testing()  # again, for support standalone testing
 
 include(FindPkgConfig)
 include(ExternalProject)
@@ -40,6 +41,13 @@ if (NOT GTestMain_FOUND)
         ${CMAKE_CURRENT_BINARY_DIR}/googletest-build
         EXCLUDE_FROM_ALL
     )
+endif()
+
+if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
+    # CMake has been started independently in this directory with tests. Do
+    # import the globally installed Guidelines Support Library and test it
+    # instead of the current version from the include/ folder.
+    find_package(Microsoft.GSL REQUIRED)
 endif()
 
 if (MSVC AND (GSL_CXX_STANDARD EQUAL 17))
@@ -157,7 +165,7 @@ set_property(TARGET  PROPERTY FOLDER "GSL_tests")
 function(add_gsl_test name)
     add_executable(${name} ${name}.cpp)
     target_link_libraries(${name}
-        GSL
+        Microsoft.GSL::GSL
         gsl_tests_config
         ${GTestMain_LIBRARIES}
     )
@@ -259,7 +267,7 @@ endif(MSVC)
 function(add_gsl_test_noexcept name)
     add_executable(${name} ${name}.cpp)
     target_link_libraries(${name}
-        GSL
+        Microsoft.GSL::GSL
         gsl_tests_config_noexcept
         ${GTestMain_LIBRARIES}
     )


tests/CMakeLists.txt
```diff
@@ -1,11 +1,12 @@
 cmake_minimum_required(VERSION 3.0.2)
 
 project(GSLTests CXX)
+enable_testing()
 
 include(FindPkgConfig)
 include(ExternalProject)
 
-find_package(Git REQUIRED QUIET)
+find_package(Git QUIET)
```
hdf89shfdfs, line 9: Add a comment explaining why the REQUIRED isn't necessary.

mymedia2: In fact, there are conditions under which Git would be needed. One of globally installed GTest or Git command is required. I have an idea to express this clause more precisely and move call to find_package() right below [GTest availability check][1] or into the CMakeLists.txt.in template that [actually uses Git][2].

 [1]: https://github.com/microsoft/GSL/pull/963/files#diff-4461c617ceae0c7e0206622aacdf555aedd62eb129c2efb74c84fa1567bcbe0dR15
 [2]: https://github.com/microsoft/GSL/blob/a47352cb2a05cc2684877d8a55077d092f20574b/tests/CMakeLists.txt.in#L6-L7

tests/CMakeLists.txt
```diff
@@ -1,17 +1,19 @@
 cmake_minimum_required(VERSION 3.0.2)
 
 project(GSLTests CXX)
+enable_testing()
```
hdf89shfdfs, line 4: enable_testing() can be called 2 times now.

Once in this test/CMakeLists.txt and in CMakeLists.txt

I'd rather just move the code in the main cmakelists.txt to this file.

tests/CMakeLists.txt
```diff
@@ -41,6 +43,10 @@ if (NOT GTestMain_FOUND)
     )
 endif()
 
+if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
+    find_package(Microsoft.GSL REQUIRED)
+endif()
```
hdf89shfdfs, line 48: Add a comment explaining why this is necessary.

hdf89shfdfs (changes requested):

tests/CMakeLists.txt
mymedia2: [Documentation says][1] that this command should be called in start directory so CTest can operate. If I remove the call in root CMakeLists.txt, `make test` will fail with "No rule" error.

 [1]: https://cmake.org/cmake/help/v3.19/command/enable_testing.html

hdf89shfdfs: Fair enough. Add a comment instead remarking on why this 2nd enable_testing() call is necessary.

hdf89shfdfs (changes requested):

tests/CMakeLists.txt
```diff
@@ -1,17 +1,19 @@
 cmake_minimum_required(VERSION 3.0.2)
```
hdf89shfdfs, line 1: Not your code but this is inconsistent with the main CMakeLists.txt

cmake_minimum_required(VERSION 3.1.3...3.16)

hdf89shfdfs (changes requested):

mymedia2:
On Вт, 2020-12-29 at 04:21 -0800, hdf89shfdfs wrote:
> @hdf89shfdfs requested changes on this pull request.
> 
> 
> 
> > @@ -1,17 +1,19 @@
>  cmake_minimum_required(VERSION 3.0.2)
> 
> Not your code but this is inconsistent with the main CMakeLists.txt
> 
> cmake_minimum_required(VERSION 3.1.3...3.16)
> 

I think this is intentional. The root CMakeLists.txt and this file use
different features that require different CMake versions.

tests/CMakeLists.txt
```diff
@@ -1,17 +1,19 @@
...
 
 project(GSLTests CXX)
+enable_testing()  # again, for support standalone testing
```
personWhoWritesCmakeExamples, line 4: Only add the enable_testing() if GSLTests is the top level project.

See the main CMakeLists.txt on how to check if a project is top level.

Also the documentation style is inconsistent with your others. Move the comment above enable_testing instead of to the side.

personWhoWritesCmakeExamples (changes requested):

tests/CMakeLists.txt
mymedia2: > Only add the enable_testing() if GSLTests is the top level project.
> See the main CMakeLists.txt on how to check if a project is top level.

Do you suggest to relocate the line under `if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)` condition?

> Also the documentation style is inconsistent with your others. Move the comment above enable_testing instead of to the side.

The comment is short enough. I do not think there is much point in splitting.

hdf89shfdfs: "Do you suggest to relocate the line under if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR) condition?"

Yes that's correct.

hdf89shfdfs: I agree with @personWhoWritesCmakeExamples it's inconsistent.
But it's a minor style issue.

The main thing for me is just making sure the rational is clear.

```cmake
project(GSLTests 
    LANGUAGES CXX
) 

if (${CMAKE_CURRENT_SOURCE_DIR} STREQUAL ${CMAKE_SOURCE_DIR})
    # Enable standalone testing.
    enable_testing()
endif()
```

PR #964, opened 2020-12-29
[cmake] Adding options for INSTALL and TEST

hdf89shfdfs:
Not all consumers of GSL automatically want to have this install logic.

It's good practice to gate install logic behind an option.
For an example look at magic_enum:
https://github.com/Neargye/magic_enum/blob/master/CMakeLists.txt

If the client wants to install GSL they still can. But they should ask
for it by overriding GSL_INSTALL.

PR #966, opened 2021-01-04
Revert "[cmake] Adding options for INSTALL and TEST"

JordanMaples:
Reverts microsoft/GSL#964


hdf89shfdfs:
Was something wrong with the changes I submitted?

JordanMaples:
@hdf89shfdfs The problem was pointed out in the commit discussion: https://github.com/microsoft/GSL/commit/eca0eca6f1ae12c80f73bdfb73bbcf912289fe34 
Essentially, when the install for GSL is created the Microsoft.GSLConfig.cmake file  is missing the INTERFACE_INCLUDE_DIRECTORIES property.

JordanMaples:
I think I found the issue. The missing line reappears when I bring the `include(GNUInstallDirs)` line back to the main file.

PR #967, opened 2021-01-05
Reintroduce CMake changes that were reverted in #966

JordanMaples:
Reintroducing the changes from #964 by @hdf89shfdfs that were reverted in #966. @pr0g found that the change unintentionally removed a line from the Microsoft.GSLConfig.cmake file, which broke `FindPackage`.

I was able to restore the missing line in the config file by moving `include(GNUInstallDirs)` back to the main CMakeLists.txt file.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,10 +3,10 @@ cmake_minimum_required(VERSION 3.1.3...3.16)
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake/")
 include(guidelineSupportLibrary)
 
-project(GSL VERSION 3.1.0 LANGUAGES CXX)
-
-# Use GNUInstallDirs to provide the right locations on all platforms
-include(GNUInstallDirs)
+project(GSL
+    VERSION 3.1.0
+    LANGUAGES CXX
+)
 
 # Creates a library GSL which is an interface (header files only)
 add_library(GSL INTERFACE)
@@ -25,6 +25,10 @@ if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
     set(GSL_STANDALONE_PROJECT ON)
 endif()
 
+### Project options
+option(GSL_INSTALL "Generate and install GSL target" ${GSL_STANDALONE_PROJECT})
+option(GSL_TEST "Build and perform GSL tests" ${GSL_STANDALONE_PROJECT})
+
 # This GSL implementation generally assumes a platform that implements C++14 support.
 set(gsl_min_cxx_standard "14")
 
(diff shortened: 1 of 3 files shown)


pr0g:
Awesome work @JordanMaples! Thanks for fixing the issue 👍😊

JordanMaples:
@pr0g Thanks again for letting me know that the previous change broke your CI.

pr0g:
As the thing that caused the issue was a bit odd (the location of the `include(GNUInstallDirs)` it might be worth dropping a question/comment on the CMake Discourse as it could potentially be a bug in CMake itself.

I don't see anything mentioned about where it needs to be defined here https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html

hdf89shfdfs:
@pr0g + @JordanMaples  thanks for fixing this. Apologies I didn't do enough local testing. Glad the new changes still got in.

pr0g:
@hdf89shfdfs I'm a little curious as to what the rationale is to make installing the GSL library a bit more difficult (by moving it behind the `GSL_INSTALL` switch). A user would have to run `cmake --build build --target install` to actually install it even without the switch. Is it just to make installing 100% explicitly opt-in?

I do think a corresponding note in the README/docs would be appropriate too as it might trip up someone trying to install the library.

hdf89shfdfs:
@pr0g this change is to allow users control of the contents of their install directory. Without this change the user would have to install gsl files into their install folder even if they didn't want to. Which isn't desirable in every circumstance.

This <PROJECT>_INTALL is a convention I've seen in many projects for this exact rational. The top level project should be 100% in control of the install directory contents.

"I do think a corresponding note in the README/docs would be appropriate too as it might trip up someone trying to install the library."

I agree there should be more documentation.

pr0g:
@hdf89shfdfs I'm afraid I don't quite understand _"Without this change the user would have to install gsl files into their install folder even if they didn't want to"_. If a user doesn't run the install command, nothing gets installed right?

Is this if you've cloned gsl or are using it via `add_subdirectory`, and then run the install command from the top level the sub library will also get installed? I don't usually do this so might not have run into this particular issue before.

hdf89shfdfs:
"Is this if you've cloned gsl or are using it via add_subdirectory, and then run the install command from the top level the sub library will also get installed? I don't usually do this so might not have run into this particular issue before."

Yes the use case is adding it via add_subdirectory. Using FetchContent for example.

pr0g:
Ah okay got you, I can see now how this could be a problem with `add_subdirectory`. I'd hope it wouldn't happen with `FetchContent` as the library will end up in a `build/_deps` folder and I thought would be excluded but I'm not actually sure. Thanks for the info, I'll have to update this with some info about it! https://github.com/pr0g/cmake-examples 😝

PR #968, opened 2021-01-06
fixes simctl race in iOS test

dipidoo:


JordanMaples:
Just a couple follow up questions here. Have you had a chance to look into the other failures that we noted in #971? Are the sleeps introduced here necessary to unblock progress on #971?

JordanMaples:
Closing this since #981 addressed and fixed the iOS tests.

PR #969, opened 2021-01-07
Minor cmake nitpicks

hdf89shfdfs:
It's much nicer and less error prone to just use add_subdirectory to establish the include directory.

Hide the GNUInstallDirs module by placing it in the helper module. The intent being that the main CMakeLists.txt should have a little code as possible. So that readers can quickly understand the project.

Use include_guard() when available in cmake 3.10+

### DIFF
--- a/cmake/guidelineSupportLibrary.cmake
+++ b/cmake/guidelineSupportLibrary.cmake
@@ -4,13 +4,20 @@
 # for multiple versions of cmake.
 #
 # Any functions/macros should have a gsl_* prefix to avoid problems
-if (DEFINED guideline_support_library_include_guard)
-    return()
+if (CMAKE_VERSION VERSION_GREATER 3.10 OR CMAKE_VERSION VERSION_EQUAL 3.10)
+    include_guard()
+else()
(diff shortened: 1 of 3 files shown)


cmake/guidelineSupportLibrary.cmake
```diff
@@ -4,14 +4,21 @@
 # for multiple versions of cmake.
 #
 # Any functions/macros should have a gsl_* prefix to avoid problems
-if (DEFINED guideline_support_library_include_guard)
-    return()
+if (CMAKE_VERSION VERSION_GREATER 3.10 OR CMAKE_VERSION VERSION_EQUAL 3.10)
```
JordanMaples, line 7: Why not use `VERSION_GREATER_EQUAL`?

hdf89shfdfs: That's not available until cmake 3.7 and the cmake minimum for this project is 3.1

JordanMaples: I didn't realize that was when it was introduced. Thanks for the clarification.

PR #972, opened 2021-01-12
Fixing cmake developer warning

jpr89:
Here is the warning currently being produced:

CMake Warning (dev) at C:/Program Files/CMake/share/cmake-3.19/Modules/GNUInstallDirs.cmake:223 (message):
  Unable to determine default CMAKE_INSTALL_LIBDIR directory because no
  target architecture is known.  Please enable at least one language before
  including GNUInstallDirs.
Call Stack (most recent call first):
  cmake/guidelineSupportLibrary.cmake:20 (include)
  CMakeLists.txt:4 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.

I noted how I fixed the error. This is caused by GNUInstallDirs automatically executing code just by including it.

I also added -Werror=dev to the CI to ensure this never happens again.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -29,7 +29,7 @@ jobs:
           echo "Emulator starting"
 
       - name: Configure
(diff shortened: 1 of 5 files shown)


JordanMaples:
Thanks for the fix.

Issue #974, opened 2021-01-24
Broken static_assert in gsl::not_null


Leedehai:
```c++
static_assert(std::is_convertible<decltype(std::declval<T>() != nullptr), bool>::value,
  "T cannot be compared to nullptr.");
```
Please see https://stackoverflow.com/a/65867132/8385554.

JordanMaples:
Could you please provide an example that shows that the `static_assert` is broken?

Addressing the answer in the StackOverflow post:
  ```c++
  #include <type_traits>
  
  template<typename T>
  void foo()
  {
      static_assert(
                std::is_convertible<decltype(std::declval<T>() != nullptr), bool>::value,
                "T cannot be compared to nullptr.");
  }
  
  int main()
  {
      foo<int>();
  } 
  ```

`gsl::not_null` is designed to work on pointers. However, the example provided on StackOverflow has a function `foo` that takes an `int` as the template parameter rather than an `int*`, which fails the `static_assert`.

beinhaerter:
@JordanMaples As written in the answer:

> This results in a compiler diagnostic, and not an assertion, using Microsoft's own compiler:

So the code is correct in the sense that it prevents `gsl::not_null<int>` to compile. But it is broken in the sense that not the `static_assert` triggers, but a compiler diagnostic is issued. We would get a better compiler warning with this code:

```
#include <type_traits>

template<typename T, typename = void>
struct is_nullptr_convertible : std::false_type {};

template <typename T>
struct is_nullptr_convertible<T, std::is_convertible<decltype(std::declval<T>() != nullptr), bool>> : std::true_type {};

template<typename T>
void foo()
{
    static_assert(
        is_nullptr_convertible<T>::value,
        "T cannot be compared to nullptr.");
}

int main()
{
    foo<int>();
}
```

So maybe we could adjust the `static_assert`!?

JordanMaples:
Ah, I understand now. I focused on the wrong part of the answer. I thought they were talking about consumption of the type rather than the source of the error. I'm going to blame a lack of coffee.

Would you mind drafting a PR so I can discuss it with the other maintainers this week?

beinhaerter:
PR #975

JordanMaples:
@CaseyCarter or @StephanTLavavej, would one of you mind looking at this `static_assert` from `gsl::not_null`?

As it currently exists:
```c++    
static_assert(std::is_convertible<decltype(std::declval<T>() != nullptr), bool>::value,
[...]

PR #975, opened 2021-01-26
is_comparable_to_nullptr for better static_assert

beinhaerter:
Trying `gsl::not_null<char> p2{ 0 };` on VS2019 the current implementation would trigger

>error C2446 : '!=' : no conversion from 'nullptr' to 'int'
>message: A native nullptr can only be converted to bool or , using reinterpret_cast, to an integral type
>message: see reference to class template instantiation 'gsl::not_null<char>' being compiled
>error C2955 : 'std::is_convertible' : use of class template requires template argument list
>message: see declaration of 'std::is_convertible'
>error C2039 : 'value' : is not a member of 'std::is_convertible<_From,_To>'
>error C2065 : 'value' : undeclared identifier

The new implementation gives much shorter and clearer message and does exactly as the `static_assert` intends to do:

> error C2338: T cannot be compared to nullptr.
> message : see reference to class template instantiation 'gsl::not_null<char *>' being compiled

### DIFF
--- a/include/gsl/pointers
+++ b/include/gsl/pointers
@@ -38,7 +38,7 @@ template<typename T, typename = void>
 struct is_comparable_to_nullptr : std::false_type {};
 
 template <typename T>
-struct is_comparable_to_nullptr<T, std::is_convertible<decltype(std::declval<T>() != nullptr), bool>> : std::true_type {};
+struct is_comparable_to_nullptr<T, std::enable_if_t<std::is_convertible<decltype(std::declval<T>() != nullptr), bool>::value>> : std::true_type {};
 }   // namespace details
 
 //
--- a/include/gsl/pointers
+++ b/include/gsl/pointers
@@ -32,6 +32,15 @@
 namespace gsl
 {
 
+namespace details
+{
+template<typename T, typename = void>
+struct is_comparable_to_nullptr : std::false_type {};
+
+template <typename T>
+struct is_comparable_to_nullptr<T, std::is_convertible<decltype(std::declval<T>() != nullptr), bool>> : std::true_type {};
+}   // namespace details
+
 //
 // GSL.owner: ownership pointers
 //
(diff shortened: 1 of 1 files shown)


beinhaerter:
I don't understand the build failures. For example on MSVC:

>D:\a\1\s\include\gsl/pointers(80,57): error C2338: T cannot be compared to nullptr. [D:\a\1\s\build\tests\notnull_tests.vcxproj]
>D:\a\1\s\tests\notnull_tests.cpp(122): message : see reference to class template instantiation 'gsl::not_null<int *>' being compiled [D:\a\1\s\build\tests\notnull_tests.vcxproj]

Testing
```
#include <gsl/pointers>

bool helper(gsl::not_null<int*> p) { return *p == 12; }

int main() {
	char x;
	gsl::not_null<char*> p1{ &x };
	gsl::not_null<char> p2{ 0 };

	return 0;
}
```

on my local VS2019 16.8.4 I only get 

>1>C:\Users\whenze\source\GSL\include\gsl\pointers(80,57): error C2338: T cannot be compared to nullptr.
>1>D:\tmp\ConsoleApplication1\ConsoleApplication1.cpp(3): message : see reference to class template instantiation 'gsl::not_null<int *>' being compiled

but no error regarding `gsl::not_null<int*>`.

include/gsl/pointers
```diff
@@ -32,6 +32,15 @@
...
 {
 
+namespace details
+{
+template<typename T, typename = void>
+struct is_comparable_to_nullptr : std::false_type {};
+
+template <typename T>
+struct is_comparable_to_nullptr<T, std::is_convertible<decltype(std::declval<T>() != nullptr), bool>> : std::true_type {};
+}   // namespace details
```
CaseyCarter, line 42: This specialization will only ever be matched if someone passes the type `std::is_convertible</*...stuff...*/>` as the second argument to `is_comparable_to_nullptr`. You need to arrange for the second parameter of your specialization to be `void`, but ideally a long and complicated spelling of `void` that's only valid when `std::declval<T>() != nullptr` is a valid expression that's convertible to `bool`. Something like:
```suggestion
template <typename T>
struct is_comparable_to_nullptr<T, std::enable_if_t<std::is_convertible<decltype(std::declval<T>() != nullptr), bool>::value>> : std::true_type {};
}   // namespace details
```
should work, for example.

beinhaerter: Thanks, that works. That was a silly error I made.

JordanMaples:
Maintainers call: Looks good to me. 

Thanks for the PR @beinhaerter and @CaseyCarter for the review and insight.
closes #974

Issue #846, opened 2020-02-13
Improper use `std::forward`


tomilov:
[Here](https://github.com/microsoft/GSL/blob/master/include/gsl/gsl_util#L79-L89) `F` in second overloading has always value (non-reference) type if just deduced. Therefore using of `std::forward` is improper (though works).

ArtemSarmini:
Should be
```c++
template <class F>
final_action<std::decay<F>::type> finally(F&& f) noexcept
{
    return final_action<std::decay<F>::type>(std::forward<F>(f));
}
```
or more specific `remove_reference<remove_const<F>::type>::type` instead of `decay`

johelegp:
You got the reference and const backwards.

ArtemSarmini:
Edited

CaseyCarter:
I came here to make an argument for why the decay is unneeded, but looking at the implementation of `final_action` I'm confused. The use of `std::move` in the member initializers suggests that `final_action::f_` is intended to be a function object. (The unconditional `noexcept` on both constructors is unfortunate, since it will cause the program terminate if the initialization of `f_` throws an exception - perhaps the intent is that `f_` is always a function pointer?) The fact that `final_action` is move constructible seems to contradict the intent of the Core Guidelines that it only be usable in a local scope.

In any case, I agree with statement in the original edition of the first post which suggested that the `finally(const F&)` overload is redundant given existence of the perfect-forwarding `finally(F&&)` overload. Whether that overload should return `final_action<F>` or `final_action<std::decay_t<F>>` is then a consequence of the design intent for `final_action` itself. Is it intended that:

* `final_action` internally stores a function object, which can be moved (along with the responsibility for invocation) into another `final_action` object (always `decay` argument types), or

* `final_action` is usable only within a local scope under the assumption that lvalue arguments passed to `finally` must be valid at the resulting `final_action`'s point of destruction (don't decay so lvalues are captured by reference and rvalues by value)?

Either design can be implemented, but it's not clear to me which design is intended.

johelegp:
The influence on the existence of the move constructor must have some from a discussion I remember in some issue/PR here or at CppCoreGuidelines. It was about being able to return a `final_action` or invalidate it so as to not invoke the final action anymore.

Odept:
Also, is there a reason why the `final_action` ctor doesn't take an rvalue-ref: it's `final_action(F f)` instead of `final_action(F&& f)`? Dosn't it always result in a copy as a result?

[...]

PR #977, opened 2021-02-25
Clean up `final_act` and `finally`, closes #752 and #846

### LINKED ISSUE #752: Finally tries to take ownership of action
The constructor of `final_action` [tries to move](https://github.com/Microsoft/GSL/blob/master/include/gsl/gsl_util#L54) the input action, which (fortunately) [breaks compilation](http://coliru.stacked-crooked.com/a/1d458725d81a28c5) when the input to `finally` is a non-const lvalue:

Replacing `std::move` with `std::forward` fixes the issue.

Related to (but different from) #125.

hsutter:
Somewhere along the way, GSL's implementation of `final_act` and `finally` seems to have become way overthought. This PR is to re-simplify these facilities back to what [C++ Core Guidelines C.30](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c30-define-a-destructor-if-a-class-needs-an-explicit-action-at-object-destruction) said which is simple and clear and works. It just copies the invocable thing, and doesn't bother trying to optimize the copy. This should be fine, because we're typically passing something that's cheap to copy, often a stateless lambda.

The problem in #846 appears to be because `finally` looks like was originally written as a `const&`/`&&` overload (its state at the time that issue was opened)... to eliminate a copy when you invoke it with a temporary. If so, then the `&&` was probably never intended to be a forwarder, but an rvalue reference that tripped over the horrid C++ syntax collision where a `&&` parameter magically instead means a forwarding reference because the type happens to be a template parameter type here. So I suspect the original author was just trying to write an rvalue overload, and the forwarder that's there now was never intended at all.

The problem in #752 just seems like a weird effect of the current complication, and with this PR that bug just disappears -- the reported example code that failed to compile now will compile and run as intended.


include/gsl/util
```diff
@@ -52,40 +52,27 @@ template <class F>
...
 
 // finally() - convenience function to generate a final_action
 template <class F>
-GSL_NODISCARD final_action<typename std::remove_cv<typename std::remove_reference<F>::type>::type>
-finally(F&& f) noexcept
+auto finally(F f) noexcept
 {
-    return final_action<typename std::remove_cv<typename std::remove_reference<F>::type>::type>(
-        std::forward<F>(f));
+    return final_action<F>{f};
```
beinhaerter, line 75: All `final_action` members are `noexcept(false)`, so having `noexcept` here looks wrong. Yet in my opinion it would be better to have `final_action` `noexcept` depending on `F`.

beinhaerter:
If this fixes #752 it would be good to add a test that proves this.

JordanMaples:
@CaseyCarter would you mind taking a look at these changes and letting you know what you think?

include/gsl/util
CaseyCarter, line 75: Copying `f` into the constructor argument here is inconsistent with the move construction on line 55. Either we want to allow move-only functions and better support functions that are more expensive to copy than move - in which case both places should move - or we don't and both places should copy.

CaseyCarter:
> @CaseyCarter would you mind taking a look at these changes and letting us know what you think?

I'm not sure what you want me to comment on. If [Guidelines C.30](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c30-define-a-destructor-if-a-class-needs-an-explicit-action-at-object-destruction) is a specification, then I can point out places where this implementation doesn't agree. However, C.30 says that it is a "simplified" example so I'm not sure what the intended design is.

hsutter:
Maintainers' call: Guidelines editors took a lot and approved, so go ahead and address the comments and merge this.

hsutter:
Sorry for the lag on this...

@CaseyCarter Casey, I've applied your changes, thanks!

@beinhaerter Werner, I've made it `noexcept`-everywhere now. I don't think we want it to be conditionally `noexcept` based on `F` unless IIUC we want to support types `F` whose copy/move could throw? I haven't seen the GSL documentation talk about intending to support such types. And the only time `F` is invoked is in the `final_act` destructor which has to be a `noexcept` call.

include/gsl/util
```diff
@@ -52,40 +52,29 @@ template <class F>
...
+    template <class FF>
+    explicit final_action(FF&& ff) noexcept : f{std::forward<FF>(ff)} { }
 
-    explicit final_action(F f) noexcept : f_(std::move(f)) {}
+    ~final_action() noexcept { if (invoke) f(); }
 
     final_action(final_action&& other) noexcept
-        : f_(std::move(other.f_)), invoke_(std::exchange(other.invoke_, false))
-    {}
+        : f(std::move(other.f)), invoke(std::exchange(other.invoke, false))
```
CaseyCarter, line 61: `F` might be an lvalue reference to non-`const`, in which case the initialization `f(std::move(other.f))` is ill-formed - a non-`const` lvalue reference can't bind to an rvalue. If you truly intend to support `F` as lvalue reference this should be:
```suggestion
        : f(std::forward<F>(other.f)), invoke(std::exchange(other.invoke, false))
```
If, on the other hand, you don't intend `F` to support references, `finally` should return `final_action<std::remove_reference_t<F>>{std::forward<F>(f)}`.

CaseyCarter (changes requested):

include/gsl/util
```diff
@@ -52,40 +52,29 @@ template <class F>
...
-
-    // clang-format off
-    GSL_SUPPRESS(f.6) // NO-FORMAT: attribute // terminate if throws
-    // clang-format on
-    ~final_action() noexcept
-    {
-        if (invoke_) f_();
-    }
+    final_action(const final_action& rhs) = delete;
+    void operator=(const final_action&)   = delete;
```
dmitrykobets-msft, line 65: missing parameter name? (alternatively, remove the other two names)

include/gsl/util
```diff
@@ -52,40 +52,29 @@ template <class F>
 class final_action
 {
 public:
-    static_assert(!std::is_reference<F>::value && !std::is_const<F>::value &&
-                      !std::is_volatile<F>::value,
-                  "Final_action should store its callable by value");
+    template <class FF>
```
dmitrykobets-msft, line 55: what's the purpose of this nested template?

dmitrykobets-msft (changes requested):

dmitrykobets-msft:
@beinhaerter 
> If this fixes #752 it would be good to add a test that proves this.

That issue is now very old and used a version of final/finally that is significantly different from the baseline for this PR; that issue no longer reproduces in the baseline either. So closed that issue.

include/gsl/util
CaseyCarter: I missed that this was a template - my brain autocorrected this to a non-template taking `F&&` as would be consistent with the "`F` is sometimes an lvalue reference type" approach. If the intent is that `F` is sometimes an lvalue reference referring to the lvalue argument to `finally`, then this should become:
```c++
explicit final_action(F&& ff) noexcept : f{std::forward<F>(ff)} { }
```
if the intent is that `F` is always an object type, a template constructor is still not a great idea. Sure, it will work with both `const F&` and `F&&` as we'd like, but it will also compete with the copy and move constructors which we don't want. For that case I'd replace this with two constructors:
```c++
explicit final_action(const F& ff) noexcept : f{ff} { }
explicit final_action(F&& ff) noexcept : f{std::move(ff)} { }
```

hsutter: Thanks, I'll take the overload set and avoid the template.

include/gsl/util
hsutter: There's nothing in the Guidelines about supporting references, so I'll change `finally`. Thanks!

include/gsl/util
hsutter: Fixed, thanks!

hsutter:
@CaseyCarter @dmitrykobets-msft Thanks for the feedback -- I've taken a pass at resolving the things you mentioned. How does this look now?

include/gsl/util
```diff
@@ -72,9 +72,9 @@ private:
 
 // finally() - convenience function to generate a final_action
 template <class F>
-[[nodiscard]] auto finally(F&& f) noexcept
+GSL_NODISCARD auto finally(F&& f) noexcept
 {
-    return final_action<F>{std::forward<F>(f)};
+    return final_action<std::remove_reference_t<F>>{std::forward<F>(f)};
```
CaseyCarter, line 77: Final suggestion (sorry I missed this before) we probably also want to strip cv-qualifiers from `F` so the resulting `final_action` doesn't have a member with `const`- or `volatile`-qualified type:
```suggestion
    return final_action<std::remove_cv_t<std::remove_reference_t<F>>>{std::forward<F>(f)};
```

CaseyCarter:
> @CaseyCarter @dmitrykobets-msft Thanks for the feedback -- I've taken a pass at resolving the things you mentioned. How does this look now?

Found a mistake I made in my original feedback, and then gave this a proper non-rushed review.

include/gsl/util
hsutter: Aha, I was wondering why you hadn't suggested `remove_cvref`! :) I figured you had a reason. OK, I'll make this change, and leave it as two separate calls to keep it C++11 and avoid creating a C++20 dependency yet. Thanks.

dmitrykobets-msft:
@hsutter the changes look good - can you please rebase onto the latest main branch? There is a recent fix to the failing android test that is blocking the merge here.

dmitrykobets-msft (approved):

CaseyCarter (approved):

CaseyCarter:
@dmitrykobets-msft's merge of `main` brought this PR to my attention and I came to do a quick review. I noticed there were no changes from the merge, and it also occurred to me in passing that `finally(an_actual_non_object_function)` is broken. For a reference-to-function type, `remove_cv_t<remove_reference_t<T>>` is a function type. Instantiating `final_action` with a function type for its type parameter is ill-formed. We need to decay function types to pointer-to-function types before instantiating `final_action`. Surprisingly, there was no test coverage for this (IMO) incredibly common use case.

I whipped up a fix for this which I can't push into the PR branch because I'm not a maintainer. Instead I opened https://github.com/microsoft/GSL/pull/1056 to merge the fix into the PR branch. I suggest a maintainer review and accept that small change before merging this PR and breaking a major use case.

PR #981, opened 2021-05-08
Fix iPhone simulator CI

### LINKED ISSUE #971: Investigate iOS failing test
The GitHub Action for iOS is disabled due to failures in `no_exception_ensure_test`. 

The test should have exit code = `0` if the test correctly executes the termination handler. Currently, the test exits with `-1`.

tusharpm:
Closes #971.

The reported failures are correlated with processes terminating extremely quickly and reported as failed to launch.

Change description:
- Added a 1s delay in `no_exception_ensure_tests` guarded by `IOS_PROCESS_DELAY_WORKAROUND`.
- Collected the remaining tests into a single executable to increase the process time and reduce copying overhead.
- Update iOS deployment target to reduce build warnings.

Result:
[Successful iOS build](https://github.com/tusharpm/GSL/runs/2535361506?check_suite_focus=true).


tests/CMakeLists.txt
```diff
@@ -164,36 +164,27 @@ target_include_directories(gsl_tests_config SYSTEM INTERFACE
...
-add_gsl_test(string_span_tests)
-add_gsl_test(at_tests)
-add_gsl_test(notnull_tests)
-add_gsl_test(assertion_tests)
-add_gsl_test(utils_tests)
-add_gsl_test(owner_tests)
-add_gsl_test(byte_tests)
-add_gsl_test(algorithm_tests)
-add_gsl_test(strict_notnull_tests)
+add_executable(gsl_tests
```
JordanMaples, line 167: Please change this section back to the old method. We prefer each test to be compiled into its own executable rather than all tests into a single executable.

tusharpm: In my tests, I [observed a failure](https://github.com/tusharpm/GSL/runs/2535279069?check_suite_focus=true) in `owner_tests` with the same error. We have two options to fix this:
1. add the delay in each process: this can be done with a modified `gtest_main.cc` including the workaround.
2. link the tests to increase the process time.

tusharpm: @JordanMaples I would like to know which of the two options are preferred? With the first one, the delay in all executables adds up.
I chose the second because GTest supports filtering test cases using command line arguments.

JordanMaples: Sorry about the delay. Your current approach is fine. I was hesitant about this approach originally because it changed the behavior of the individual test readout in the Azure Pipeline. My worry was that consolidating the tests into a single executable would make it more difficult to debug failures. However, this is not an issue because the CTest step is decorated w/ `--output-on-failure` so there shouldn't be any difference in quickly identifying failing tests. 

Would you mind removing the function `add_gsl_test_noexcept` below similar to what you did for `add_gsl_test` so the two tests are consistent?

tusharpm: Done.

JordanMaples:
Looks good. Thanks!

tusharpm:
@JordanMaples, the CI_iOS action was not run for the merge. It might need to be enabled manually.

JordanMaples:
Unfortunately I forgot to re-enable it until after I merged the change and there hasn't been any other changes to trigger it since then. I'll need to look for a way to manually trigger the action to run.

Issue #984, opened 2021-05-19
gsl::at behavior with gsl::span and std::span


JordanMaples:
When passing `std::span` to `gsl::at`, the span is passed by reference.
When passing `gsl::span` to `gsl::at`, the span is passed by value, assuming the specialization for `gsl::span` is used.

The way these are passed should be consistent, and so the `gsl::span` specialization of `gsl::at` should take the span by reference instead.

Issue #983, opened 2021-05-19
move gsl::at for gsl::span


JordanMaples:
The version of `gsl::at` that is invoked when calling it with a `gsl::span` changes depending on which headers are included.

`<gsl/span_ext>` has a specialization for `gsl::span` that takes the span by copy: 
```c++
// Specialization of gsl::at for span
template <class ElementType, std::size_t Extent>
constexpr ElementType& at(span<ElementType, Extent> s, index i)
{
    // No bounds checking here because it is done in span::operator[] called below
    Ensures(i >= 0);
    return s[narrow_cast<std::size_t>(i)];
}
```
However, unless you include `<gsl/span_ext>`, `<gsl/string_view>`, or `<gsl/gsl>` the version of at that is compiled is the one from `<gsl/util>`.
```c++
template <class Cont>
// clang-format off
GSL_SUPPRESS(bounds.4) // NO-FORMAT: attribute
GSL_SUPPRESS(bounds.2) // NO-FORMAT: attribute
// clang-format on
    constexpr auto at(Cont& cont, const index i) -> decltype(cont[cont.size()])
{
    Expects(i >= 0 && i < narrow_cast<index>(cont.size()));
    using size_type = decltype(cont.size());
    return cont[narrow_cast<size_type>(i)];
}
``` 

If we intend to continue offering this specialization, it should be moved into `<gsl/span>`.

PR #985, opened 2021-05-19
gsl::at behavior change regarding gsl::span

JordanMaples:
Changed the file location of gsl::span specialization for gsl::at from <gsl/span_ext> to <gsl/span>, addressing #983.
Changed first parameter of gsl::span specialization for gsl::at to take the span by reference, addressing #984.

### DIFF
--- a/include/gsl/assert
+++ b/include/gsl/assert
@@ -22,6 +22,7 @@
 // Currently terminate is a no-op in this mode, so we add termination behavior back
 //
 #if defined(_MSC_VER) && (defined(_KERNEL_MODE) || (defined(_HAS_EXCEPTIONS) && !_HAS_EXCEPTIONS))
+#define GSL_KERNEL_MODE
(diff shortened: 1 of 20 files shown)


JordanMaples:
Changes to be made per maintainer sync:
revert current changes to <span> and <span_ext>
<span> to include <span_ext>
add std::span specialization of gsl::at to <util>

In another PR, deprecate span comparison operators in <span_ext>

JordanMaples:
I compiled this with /kernel in MSVC and found that the tests did not cleanly build/pass. The `set_terminate` call would return nullptr, indicating that the terminate handler wasn't being set. This then lead to the death tests failing because the death string wasn't found. I created a new header and moved the death strings with a new function to determine which regex string to use.

sunnychatterjee (approved):
@JordanMaples Looks good to me.

PR #989, opened 2021-07-12
Allow constructing span<char> from std::string in C++14

### LINKED ISSUE #987: Cannot create gsl::span<char> from std::string
One Compiler Explorer says more than a thousand words: https://godbolt.org/z/j4xnvdq8o

(`string_span` works, but since it's deprecated, I would rather not use it.)

stertingen:
Fix #987


JordanMaples:
Please see response in the issue.

PR #993, opened 2021-09-06
Fix gsl/util for c++20 compilers without <span>

jcelerier:
For instance, clang 10 sets __cplusplus >= 202002L yet does not have span, which causes build errors: 

https://gcc.godbolt.org/z/Yq345zGea


JordanMaples:
Maintainers' call: Thanks for the change. Merged

PR #1019, opened 2021-12-10
Update compiler support

dmitrykobets-msft:
This PR updates the compilers that we are officially supporting, meaning the compilers that we test against in our CI pipeline. The versions are aligned with our philosophy of supporting the "latest" and "latest-1" major releases, with effort to support the latest "minor / revision" numbers within these two major releases.


sunnychatterjee (approved):
Looks good to me. Thanks for updating the list!

PR #1020, opened 2021-12-15
Add tests for VS2022 with LLVM

dmitrykobets-msft:
The Windows-2022 Clang rollout seems to be under way -- the documentation hasn't been updated yet but it looks like the changes to the VMs are in? Let's try them out


dmitrykobets-msft:
It appears that VS Clang is not yet available in windows-2022 VMs :(

dmitrykobets-msft:
There have been other compiler updates since this PR was created; closing in favor of a new PR to tweak compiler support + reduce the frequency with which it is necessary to update the compiler version table in the README.

PR #1021, opened 2022-01-03
Update compiler support

dmitrykobets-msft:
Bump clang support to 12 and 11, and add support for VS with LLVM


sunnychatterjee (approved):
Looks good, thanks Dmitry!

Issue #1013, opened 2021-11-03
subspan creation shouldn't check that size is != dynamic_extent, causes code bloat


JosephBialekMsft:
Currently the subspan code checks that size != dynamic_extent. 

This happens in the following two places:
```
        constexpr explicit extent_type(size_type size) : size_(size)
        {
            Expects(size != dynamic_extent);
        }
```

and 

```
        template <class OtherExtentType>
        constexpr storage_type(KnownNotNull data, OtherExtentType ext)
            : ExtentType(ext), data_(data.p)
        {
            Expects(ExtentType::size() != dynamic_extent);
        }
```

The result is that the following simple test case has an unneeded cmp/branch.

```
#include <windows.h>
#include <iostream>
#include <gsl/gsl>
#include <stdlib.h>

__declspec(noinline)
void
PrintSpan(
    gsl::span<const BYTE> MySpan)
{
    for (auto c : MySpan)
    {
        printf("%i", c);
    }
}

__declspec(noinline)
void
DoStuff(
    gsl::span<const BYTE> MySpan, ULONG Offset)
{
    MySpan = MySpan.subspan(8 + Offset);
    PrintSpan(MySpan);
}

int main()
{
    size_t bufSize = rand() / 1000;
    BYTE* buf = (BYTE*)malloc(bufSize);

    gsl::span<const BYTE> MySpan{ buf, bufSize };

    DoStuff(MySpan, rand() / 50);

}
```

asm:
```
__declspec(noinline)
void
DoStuff(
    gsl::span<const BYTE> MySpan, ULONG Offset)
{
00007FF78B9F10D0  sub         rsp,38h  
    MySpan = MySpan.subspan(8 + Offset);
00007FF78B9F10D4  mov         rax,qword ptr [rcx]                                                 // load size from span
00007FF78B9F10D7  add         edx,8                                                                      // add 8 to Offset
00007FF78B9F10DA  cmp         rax,rdx                                                                   // compare size to the total offset
00007FF78B9F10DD  jb          DoStuff+43h (07FF78B9F1113h)                            // jump to fast-fail code if size is too small
00007FF78B9F10DF  sub         rax,rdx                                                                    // adjust size
00007FF78B9F10E2  mov         qword ptr [rsp+20h],rax                                       // store size to stack temporary span
**00007FF78B9F10E7  cmp         rax,0FFFFFFFFFFFFFFFFh                                       // compare size to dynamic_extent
[...]

PR #1029, opened 2022-01-28
Remove useless check against `dynamic_extent`

dmitrykobets-msft:
Addresses https://github.com/microsoft/GSL/issues/1013


sunnychatterjee (approved):
Dmitry and I walked through the code and this change looks reasonable to me. We'll follow up with Casey for another round of review and then merge.

CaseyCarter (approved):
LGTM - I agree with Dmitry's analysis in #1013.

Issue #1025, opened 2022-01-27
GSL violates SF.12


beinhaerter:
[SF.12: Prefer the quoted form of #include for files relative to the including file and the angle bracket form everywhere else](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rs-incform)

> [...] the guidance is to use the quoted form for including files that exist at a relative path to the file containing the #include statement (from within the same component or project) and to use the angle bracket form everywhere else, where possible. [...]

The GSL includes its own files as `#include <gsl/xxx>`. It would be better to use `#include "xxx"` as SF.12 proposes.

If you agree to SF.12, I would be ready to prepare a PR.

<hr/>

The way it currently is, results in problems in our projects. Following is some background.

We have a build environment which handles all projects and the include pathes. For example

Project not under my control:
```
project1/include/...
project1/src/project1.cpp
```

Project under my control:
```
project2/include/project2/project2.h
project2/src/project2.cpp
```

Project under my control:
```
project3/include/project3/project3.h
project3/src/project3.cpp
```

For project3 the build environment (not under my control) uses gcc with `-Iproject1/include -Iproject2/include`.
`project3.cpp` uses `#include <project2/project2.h>`.

In this structure we want to integrate GSL. GSL cannot be a made global project, so I must make it a subfolder in a project under my control (`project2`). I want to add a new folder `project2/include/project2/gsl` with all the GSL headers like

```
project2/include/project2/gsl/narrow
project2/include/project2/gsl/util
```

The GSL folder is added under `project2/include/project2/` and not under `project2/include/` so that it cannot collide if for example `project1`'s author decides to add the `gsl` folder as `project1/include/gsl`.
<hr/>

In this setup, to use GSL, `project2` needs to write `#include <project2/gsl/narrow>`.

This fails because `gsl/narrow` does `#include <gsl/util>` which fails. It would however work if GSL used `#include "util"` as SF.12 suggests.

The way we currently solve it, is to apply the changes that I ask for, in our local copy of the GSL.

dmitrykobets-msft:
Hi @beinhaerter, thanks for the clear explanation of this issue. Your proposal seems good to me, I'd be happy to take a look at your PR.

PR #1030, opened 2022-01-31
quoted form of #include when GSL includes GSL files

beinhaerter:
[SF.12: Prefer the quoted form of #include for files relative to the including file and the angle bracket form everywhere else](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rs-incform)

Additionally changed #include order in `span` so that all `span_ext` is in the GSL include block and not in the STL include block.

Fixes issues #1025.


dmitrykobets-msft (approved):
Thanks @beinhaerter, looks good

PR #1041, opened 2022-03-21
Suppress reserved identifier warning

dmitrykobets-msft:
The following reserved identifiers are being used specifically to target certain MSVC constructs, so suppress the warning in VS 2022 (LLVM) "... is reserved because it starts with '_' followed by a capital letter":
- _Unchecked_type
- _Verify_range
- _Verify_offset
- _Unwrapped
- _Unwrap_when_unverified
- _Seek_to
- _Unchecked_begin
- _Unchecked_end


sunnychatterjee (approved):
Looks good Dmitry, thanks for the fix!

PR #1042, opened 2022-03-22
Fix C26467(es.46) warning in gsl::narrow implementation

### LINKED ISSUE #1036: es.46 on gsl::narrow/gsl::narrow_cast
When calling `gsl::narrow<double>(size_t)` under VS 16.11.10 I get the warning

> gsl\narrow(42): warning C26467: Converting from floating point to unsigned integral types results in non-portable code if the double/float has a negative value. Use gsl::narrow_cast or gsl::narrow instead to guard against undefined behavior and potential data loss (es.46).

It looks like this warning should be suppressed inside `gsl::narrow` and `gsl::narrow_cast`.

When adding `GSL_SUPPRESS(es.46) // NO-FORMAT: attribute` to both functions directly after the already existing `GSL_SUPPRESS(f.6)` I still get the warning.

Is this something that the VS team should look at?
Is this a warning that we want to suppress in the GSL? If so, with `GSL_SUPPRESS` or with `#pragma warning(disable: ...)` as a workaround?

dmitrykobets-msft:
The static_cast is an intended implementation detail of gsl::narrow. However, MSVC correctly
raises warning C26467 and suggests to use `gsl::narrow_cast`. Use the suggestion, especially since
gsl::narrow_cast is being used to make the initial U -> T conversion. 

Also, while the CoreGuidelines use `static_cast` in their definition of `gsl::narrow`, `gsl::narrow_cast` is officially defined as `static_cast`, so no functional nor semantic regression results from this change.

Fixes #1036

### DIFF
--- a/include/gsl/narrow
+++ b/include/gsl/narrow
@@ -43,7 +43,7 @@ GSL_SUPPRESS(p.2) // NO-FORMAT: attribute // don't rely on undefined behavior
                                    // and cannot fit into the destination integral type), the resultant behavior is benign on the platforms
                                    // that we target (i.e., no hardware trap representations are hit).
 
-    if (static_cast<U>(t) != u || (is_different_signedness && ((t < T{}) != (u < U{}))))
+    if (narrow_cast<U>(t) != u || (is_different_signedness && ((t < T{}) != (u < U{}))))
     {
         throw narrowing_error{};
     }


dmitrykobets-msft:
Update: using gsl::narrow_cast to replace the static_cast is incorrect, since in the implementation of gsl::narrow it is assumed that the static_cast conversion is (most often) widening.

include/gsl/narrow
```diff
@@ -43,7 +43,7 @@ GSL_SUPPRESS(p.2) // NO-FORMAT: attribute // don't rely on undefined behavior
                                    // and cannot fit into the destination integral type), the resultant behavior is benign on the platforms
                                    // that we target (i.e., no hardware trap representations are hit).
 
-    if (static_cast<U>(t) != u || (is_different_signedness && ((t < T{}) != (u < U{}))))
+    if (narrow_cast<U>(t) != u || (is_different_signedness && ((t < T{}) != (u < U{}))))
```
sunnychatterjee, line 46: @dmitrykobets-msft It's better to leave it as static_cast and add a suppression in function scope as we discussed.

PR #1044, opened 2022-04-25
gsl/narrow should include <exception>

TheJCAB:
This file uses std::exception, so it should include the appropriate header.

Normally it gets the STL's "exception" header included via the gsl/assert file, but this is skipped with _HAS_EXCEPTIONS=0. I understand _HAS_EXCEPTIONS is undocumented and unsupported, but regardless, the appropriate header should be included here.

Alternatively, gsl/narrow should be modified to support _HAS_EXCEPTIONS=0, like gsl/assert was. But I'm not proposing that change. "exception" does define std::exception even with _HAS_EXCEPTIONS=0.

### DIFF
--- a/include/gsl/narrow
+++ b/include/gsl/narrow
@@ -16,8 +16,9 @@
 
 #ifndef GSL_NARROW_H
 #define GSL_NARROW_H
-#include "assert" // for Expects
-#include "util"   // for narrow_cast
+#include "assert"    // for Expects
+#include "util"      // for narrow_cast
+#include <exception> // for std::exception
 namespace gsl
 {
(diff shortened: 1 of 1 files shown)


dmitrykobets-msft:
Hi @TheJCAB, thanks for helping clean up the code here, the fix looks good.
I think you are missing a word in the last sentence of your change description. Could you please fix that? Thanks.

TheJCAB:
@dmitrykobets-msft  "missing a word in the last sentence of your change description" - It was complete, but I guess angle brackets are not good here. Too used to ADO's markup, I suppose. Thanx! Fixed.

PR #1049, opened 2022-06-21
Resolve MSVC warning C5260

dmitrykobets-msft:
MSVC warning C5260 says namespace-scope `constexpr` variables should be marked `inline` or `static` for C++20 header units. 
("warning C5260: the constant variable 'X' has internal linkage in an included header file context, but external linkage in imported header unit context; consider declaring it 'inline' as well if it will be shared across translation units, or 'static' to express intent to use it local to this translation unit").

### DIFF
--- a/include/gsl/span_ext
+++ b/include/gsl/span_ext
@@ -41,7 +41,7 @@ namespace gsl
 {
 
 // [span.views.constants], constants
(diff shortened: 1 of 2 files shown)


sunnychatterjee (approved):
Looks good, thanks for the fix.

PR #1052, opened 2022-08-17
solve span compile problem with gcc 5.5.0

beinhaerter:


dmitrykobets-msft:
Hi @beinhaerter, thanks for the fix - I think it's fine to merge because it's so minimal and addresses a bug in GCC (albeit old). Can you rebase / merge on top of the latest main branch? I recently pushed some changes that should fix the failing CI_android run which is blocking the merge for this PR.

PR #1056, opened 2022-09-29
Properly handle finally(actual_function)

CaseyCarter:
`finally` needs to use `decay_t` instead of `remove_cvref_t` so it can properly accept non-object function arguments by decaying to function pointer type. Adds test coverage for this use case which was previously missing.


dmitrykobets-msft:
Thanks for the catch!

PR #1058, opened 2022-10-03
Fix clang-15 warnings

AtariDreams:


dmitrykobets-msft:
Hi @AtariDreams, can you please describe the warning instances you are fixing, and the warning flags you used to enable the warnings? I'm having a hard time reproducing any warnings for this file when compiling with clang 15 with `-Wall -Weverything -pedantic -Wextra`

dmitrykobets-msft:
Thanks @AtariDreams,
The compiler will attempt to mark `default`-ed move constructors as `noexcept` in cases where appropriate. Your modification to always force it to be `noexcept` will change the current exception behavior. For instance, if a custom pointer type may throw on copy, then using the `noexcept` move constructor will call `std::terminate`, whereas currently it will propagate the exception.

The other changes look good to me.

AtariDreams:
I do not see where it can throw on copy or why, but I will take your word on it.

AtariDreams:
Also default copy constructors do not throw on copy

PR #1059, opened 2022-10-10
Prevent inefficient copying when using not_null::get

dmitrykobets-msft:
Issue https://github.com/microsoft/GSL/issues/550 highlighted overhead in `not_null::get` for larger types such as `shared_ptr`. Every call to `get` would return a copy of the contained value. 
This PR implements Herb's suggestion for changing the return type of `not_null::get`. The `not_null`'s value will now only be copied if it is "trivially copyable"; otherwise, it will be returned by const reference. 
Note: this change also forces the returned copy to be `const`.


hsutter (approved):

hsutter:
LGTM, thanks Dmitry!

PR #1061, opened 2022-10-23
Raise CMake min to 3.8

jpr42:
This avoids propogating -std=c++14 as a compile option

Currently if you use less than CMake 3.8 and you install the project
`-std=c++14` gets added to the INTERFACE_COMPILE_OPTIONS. This forces
users to use C++ 14 or remove the property from the imported target.

The solution is to raise the minimum and use `cxx_std_14`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,21 +1,10 @@
-cmake_minimum_required(VERSION 3.1.3...3.16)
+cmake_minimum_required(VERSION 3.8...3.16)
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake/")
 include(gsl_functions)
 
-project(GSL
-    VERSION 4.0.0
-    LANGUAGES CXX
-)
+project(GSL VERSION 4.0.0 LANGUAGES CXX)
 
-# Creates a library GSL which is an interface (header files only)
 add_library(GSL INTERFACE)
-
-# NOTE: If you want to use GSL prefer to link against GSL using this alias target
-# EX:
-#   target_link_libraries(foobar PRIVATE Microsoft.GSL::GSL)
-#
-# Add Microsoft.GSL::GSL alias for GSL so that dependents can be agnostic about
-# whether GSL was added via `add_subdirectory` or `find_package`
 add_library(Microsoft.GSL::GSL ALIAS GSL)
 
 # https://cmake.org/cmake/help/latest/variable/PROJECT_IS_TOP_LEVEL.html
@@ -24,14 +13,8 @@ string(COMPARE EQUAL ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_SOURCE_DIR} PROJECT_IS_
 option(GSL_INSTALL "Generate and install GSL target" ${PROJECT_IS_TOP_LEVEL})
 option(GSL_TEST "Build and perform GSL tests" ${PROJECT_IS_TOP_LEVEL})
 
-# This GSL implementation generally assumes a platform that implements C++14 support.
-set(gsl_min_cxx_standard "14")
-
-if (PROJECT_IS_TOP_LEVEL)
-    gsl_set_default_cxx_standard(${gsl_min_cxx_standard})
-else()
-    gsl_client_set_cxx_standard(${gsl_min_cxx_standard})
-endif()
+# The implementation generally assumes a platform that implements C++14 support
+target_compile_features(GSL INTERFACE "cxx_std_14")
 
 # Setup include directory
 add_subdirectory(include)
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ Note: These CI/CD steps are run with each pull request, however failures in them
 ## Building the tests
 To build the tests, you will require the following:
 
-* [CMake](http://cmake.org), version 3.1.3 (3.2.3 for AppleClang) or later to be installed and in your PATH.
+* [CMake](http://cmake.org), version 3.8 or later to be installed and in your PATH.
 
 These steps assume the source code of this repository has been cloned into a directory named `c:\GSL`.
 
--- a/cmake/gsl_functions.cmake
+++ b/cmake/gsl_functions.cmake
@@ -5,68 +5,16 @@
 #
 # Any functions/macros should have a gsl_* prefix to avoid problems
 
-function(gsl_set_default_cxx_standard min_cxx_standard)
-    set(GSL_CXX_STANDARD "${min_cxx_standard}" CACHE STRING "Use c++ standard")
-
-    # when minimum version required is 3.8.0 remove if below
-    # both branches do exactly the same thing
-    if (CMAKE_VERSION VERSION_LESS 3.7.9)
-        if (MSVC)
-            set(GSL_CXX_STD_OPT "-std:c++${GSL_CXX_STANDARD}")
-        else()
-            set(GSL_CXX_STD_OPT "-std=c++${GSL_CXX_STANDARD}")
-        endif()
-
-        include(CheckCXXCompilerFlag)
-        CHECK_CXX_COMPILER_FLAG("${GSL_CXX_STD_OPT}" COMPILER_SUPPORTS_CXX_STANDARD)
-
-        if(COMPILER_SUPPORTS_CXX_STANDARD)
-            target_compile_options(GSL INTERFACE "${GSL_CXX_STD_OPT}")
-        else()
-            message(FATAL_ERROR "The compiler ${CMAKE_CXX_COMPILER} has no c++${GSL_CXX_STANDARD} support. Please use a different C++ compiler.")
-        endif()
-    else()
-        # Compiler must support at least this standard
-        target_compile_features(GSL INTERFACE "cxx_std_${GSL_CXX_STANDARD}")
-        # on *nix systems force the use of -std=c++XX instead of -std=gnu++XX (default)
-        set(CMAKE_CXX_EXTENSIONS OFF)
-    endif()
-endfunction()
-
-# The best way for a project to specify the GSL's C++ standard is by the client specifying
-# the CMAKE_CXX_STANDARD. However, this isn't always ideal. Since the CMAKE_CXX_STANDARD is
-# tied to the cmake version. And many projects have low cmake minimums.
-#
-# So provide an alternative approach in case that doesn't work.
-function(gsl_client_set_cxx_standard min_cxx_standard)
-    if (DEFINED CMAKE_CXX_STANDARD)
-        if (${CMAKE_CXX_STANDARD} VERSION_LESS ${min_cxx_standard})
-            message(FATAL_ERROR "GSL: Requires at least CXX standard ${min_cxx_standard}, user provided ${CMAKE_CXX_STANDARD}")
-        endif()
-
-        # Set the GSL standard to what the client desires
-        set(GSL_CXX_STANDARD "${CMAKE_CXX_STANDARD}" PARENT_SCOPE)
-
-        # Exit out early to avoid extra unneccessary work
-        return()
-    endif()
-
-    # Otherwise pick a reasonable default
-    gsl_set_default_cxx_standard(${min_cxx_standard})
-endfunction()
-
 # Adding the GSL.natvis files improves the debugging experience for users of this library.
 function(gsl_add_native_visualizer_support)
-    if (CMAKE_VERSION VERSION_GREATER 3.7.8)
-        if (MSVC_IDE)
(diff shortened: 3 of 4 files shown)


cmake/gsl_functions.cmake
```diff
@@ -5,68 +5,16 @@
...
-        endif()
-
-        include(CheckCXXCompilerFlag)
-        CHECK_CXX_COMPILER_FLAG("${GSL_CXX_STD_OPT}" COMPILER_SUPPORTS_CXX_STANDARD)
-
-        if(COMPILER_SUPPORTS_CXX_STANDARD)
-            target_compile_options(GSL INTERFACE "${GSL_CXX_STD_OPT}")
-        else()
-            message(FATAL_ERROR "The compiler ${CMAKE_CXX_COMPILER} has no c++${GSL_CXX_STANDARD} support. Please use a different C++ compiler.")
-        endif()
```
jpr42, line 27: This is the problematic code. It causes the following result in the `Microsoft.GSLConfig.cmake`:

```cmake
set_target_properties(Microsoft.GSL::GSL PROPERTIES
  INTERFACE_COMPILE_OPTIONS "-std=c++14" # <---------- This is problematic
)
```

This causes clients to receive the compiler flag. In the worst case scenario, clients have to filter or zero out the `INTERFACE_COMPILE_OPTIONS` property. Which isn't ideal.

CMakeLists.txt
```diff
@@ -24,14 +13,9 @@ string(COMPARE EQUAL ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_SOURCE_DIR} PROJECT_IS_
...
-set(gsl_min_cxx_standard "14")
-
-if (PROJECT_IS_TOP_LEVEL)
-    gsl_set_default_cxx_standard(${gsl_min_cxx_standard})
-else()
-    gsl_client_set_cxx_standard(${gsl_min_cxx_standard})
-endif()
+# Because GSL is a header only library, that assumes at minimum C++14 support
+# We must require clients to use a compiler that supports at least that.
+target_compile_features(GSL INTERFACE "cxx_std_14")
```
jpr42, line 18: All this does is require at least a C++14 standard compliant compiler.

CMakeLists.txt
```diff
@@ -1,21 +1,10 @@
...
 
-# Creates a library GSL which is an interface (header files only)
 add_library(GSL INTERFACE)
-
-# NOTE: If you want to use GSL prefer to link against GSL using this alias target
-# EX:
-#   target_link_libraries(foobar PRIVATE Microsoft.GSL::GSL)
-#
-# Add Microsoft.GSL::GSL alias for GSL so that dependents can be agnostic about
-# whether GSL was added via `add_subdirectory` or `find_package`
```
jpr42, line 18: This is now covered in the README.md

cmake/gsl_functions.cmake
```diff
@@ -5,68 +5,16 @@
...
-        return()
-    endif()
-
-    # Otherwise pick a reasonable default
-    gsl_set_default_cxx_standard(${min_cxx_standard})
-endfunction()
-
 # Adding the GSL.natvis files improves the debugging experience for users of this library.
 function(gsl_add_native_visualizer_support)
-    if (CMAKE_VERSION VERSION_GREATER 3.7.8)
```
jpr42, line 60: Raising the min also removes this check

tests/CMakeLists.txt
```diff
@@ -1,12 +1,18 @@
...
-project(GSLTests CXX)
+project(GSLTests LANGUAGES CXX)
 enable_testing()  # again, for support standalone testing
 
+set(GSL_CXX_STANDARD "14" CACHE STRING "Use c++ standard")
+
+set(CMAKE_CXX_STANDARD ${GSL_CXX_STANDARD})
+set(CMAKE_CXX_EXTENSIONS OFF)
+set(CXX_STANDARD_REQUIRED ON)
+
```
jpr42, line 11: This is fundamentally where setting the `GSL_CXX_STANDARD` actually matters. Setting it via GSL just complicates things.

cmake/gsl_functions.cmake
jpr42: To be clear this is only a problem for clients using less than 3.8

tests/CMakeLists.txt
jpr42: Also to avoid writing complicated CMake just raise the cmake_minimum, since it won't impact clients at all.

CMakeLists.txt
```diff
@@ -24,13 +13,10 @@ string(COMPARE EQUAL ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_SOURCE_DIR} PROJECT_IS_
...
-
-if (PROJECT_IS_TOP_LEVEL)
-    gsl_set_default_cxx_standard(${gsl_min_cxx_standard})
-else()
-    gsl_client_set_cxx_standard(${gsl_min_cxx_standard})
+# Because GSL is a header only library, that assumes at minimum C++14 support
+# We must require clients to use a compiler that supports at least that.
+if (CMAKE_VERSION VERSION_GREATER "3.8" OR CMAKE_VERSION VERSION_EQUAL "3.8")
+    target_compile_features(GSL INTERFACE "cxx_std_14")
 endif()
```
jpr42, line 20: To summarize the solution.

If CMake less than 3.8 do nothing. This leaves it up to the user which is call we can do.
If CMake greater than or equal to 3.8, use compile feature cxx_std_14. To ensure that users of the library are compiling GSL as expected/required.

CMakeLists.txt
```diff
@@ -24,13 +13,10 @@ string(COMPARE EQUAL ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_SOURCE_DIR} PROJECT_IS_
...
 option(GSL_TEST "Build and perform GSL tests" ${PROJECT_IS_TOP_LEVEL})
 
-# This GSL implementation generally assumes a platform that implements C++14 support.
-set(gsl_min_cxx_standard "14")
-
-if (PROJECT_IS_TOP_LEVEL)
-    gsl_set_default_cxx_standard(${gsl_min_cxx_standard})
-else()
-    gsl_client_set_cxx_standard(${gsl_min_cxx_standard})
+# Because GSL is a header only library, that assumes at minimum C++14 support
```
dmitrykobets-msft, line 16: The way I'm reading this comment: the fact that the GSL is a header-only library implies that C++14 is the minimum-supported version. Is that correct? Why does header-only imply a minimum of C++14?

jpr42: Yeah that's worded weirdly. Apologies.

jpr42: I reworded it to `GSL exposes an interface to clients that requires at minimum C++14` which is more concise.

jpr42: >  Why does header-only imply a minimum of C++14?

It doesn't. My wording was awkward.

As per the READMEmd

"The entire implementation is provided inline in the headers under the [gsl](https://github.com/microsoft/GSL/blob/main/include/gsl) directory. The implementation generally assumes a platform that implements C++14 support."

This is enforceable with CMake via `target_compile_features(GSL INTERFACE "cxx_std_14")`

jpr42: The CMake documentation about this scenario is quite nice:
https://cmake.org/cmake/help/latest/manual/cmake-compile-features.7.html#requiring-language-standards

jpr42: Also CI only tests C++ 14, 17, 20 FWIW

```yml
parameters:
  CXXVersions: [ 14, 17, 20 ]
  buildTypes: [ 'Debug', 'Release' ]
  image: ''
```

dmitrykobets-msft: thanks for clarifying

dmitrykobets-msft:
Hi @jpr42 
> This forces users to use C++ 14 or remove the property from the imported target
Won't the -std=c++14 flag be overridden by whichever compiler version the user project specifies? 

As currently implemented, the logic in `gsl_set_default_cxx_standard` will call `CHECK_CXX_COMPILER_FLAG` will produce a cmake warning if the compiler does not support C++14. With this PR this check is removed.

jpr42:
> Won't the -std=c++14 flag be overridden by whichever c++ version the user project specifies?

I wasn't sure so I just tested this. Here is my summarized output from ninja.

```
/usr/bin/c++ ... -std=c++14 ... -std=c++17 GSL/tests/span_ext_tests.cpp
```

With compilers it's generally the last flag wins approach. And gcc doesn't seem to complain.

But 2 things.

- 1.) It muddies the CLI
- 2.) I'm almost positive MSVC complains about this situation when it happens (https://learn.microsoft.com/en-us/cpp/error-messages/tool-errors/command-line-warning-d9025?view=msvc-170). Although this will only affect Ninja users. Since Visual Studio generated builds don't allow contradictory compiler flags.

jpr42:
> As currently implemented, the logic in `gsl_set_default_cxx_standard` will call `CHECK_CXX_COMPILER_FLAG` will produce a cmake warning if the compiler does not support C++14. With this PR this check is removed.

If you use less the CMake 3.8 that is correct. Otherwise `target_compile_features(GSL INTERFACE "cxx_std_14")` [handles that](https://crascit.com/2015/03/28/enabling-cxx11-in-cmake/).

CMakeLists.txt
```diff
@@ -24,13 +13,9 @@ string(COMPARE EQUAL ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_SOURCE_DIR} PROJECT_IS_
...
-# This GSL implementation generally assumes a platform that implements C++14 support.
-set(gsl_min_cxx_standard "14")
-
-if (PROJECT_IS_TOP_LEVEL)
-    gsl_set_default_cxx_standard(${gsl_min_cxx_standard})
-else()
-    gsl_client_set_cxx_standard(${gsl_min_cxx_standard})
+# The implementation generally assumes a platform that implements C++14 support
+if (CMAKE_VERSION VERSION_GREATER "3.8" OR CMAKE_VERSION VERSION_EQUAL "3.8")
+    target_compile_features(GSL INTERFACE "cxx_std_14")
```
dmitrykobets-msft, line 18:
```suggestion
    target_compile_features(GSL INTERFACE "cxx_std_14")
else()
    include(CheckCXXCompilerFlag)
    CHECK_CXX_COMPILER_FLAG("-std=c++14" COMPILER_SUPPORTS_CXX_STANDARD)
    if(NOT COMPILER_SUPPORTS_CXX_STANDARD)
        message(FATAL_ERROR "The compiler ${CMAKE_CXX_COMPILER} has no c++14 support. Please use a different C++ compiler.")
    endif()
```

README.md
```diff
@@ -123,6 +123,8 @@ To build the tests, you will require the following:
 
 * [CMake](http://cmake.org), version 3.1.3 (3.2.3 for AppleClang) or later to be installed and in your PATH.
 
+NOTE: To run tests you need at least 3.8
```
dmitrykobets-msft, line 126: Here we can just replace this whole "To build the tests, you will require... NOTE: To run the tests" section with a simple
`To build the tests, you will require CMake version 3.8 or higher.`

README.md
```diff
@@ -123,6 +123,8 @@ To build the tests, you will require the following:
...
+
```
dmitrykobets-msft, line 127: Now that DGSL_TEST is off by default, replace the cmake configuration command below with
`cmake -G "Visual Studio 15 2017" c:\GSL -DGSL_TEST=ON`

CMakeLists.txt
```diff
@@ -24,13 +13,9 @@ string(COMPARE EQUAL ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_SOURCE_DIR} PROJECT_IS_
 option(GSL_INSTALL "Generate and install GSL target" ${PROJECT_IS_TOP_LEVEL})
 option(GSL_TEST "Build and perform GSL tests" ${PROJECT_IS_TOP_LEVEL})
```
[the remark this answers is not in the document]
dmitrykobets-msft: Users attempting to do a fresh install with cmake < 3.8 will run into the 3.8 limit error even if they do not intend to run the tests. So make the tests no longer on by default, and document how to turn them on in the README

jpr42:
@dmitrykobets-msft as a test to see if CI would succeed I raised the minimum to `3.8`

It fixes the issues you pointed out in developers workflows very cleanly.

Given that CMake `3.8` was released in April 10, 2017 (https://www.kitware.com/cmake-3-8-0-available-for-download/) and the latest release is `3.25` I think raising the minimum is agreeable.

If this change is unacceptable, I will revert my latest commit and fix the code as you suggested for compatibility with `3.1`.

jpr42:
Updated PR description and squashed commits into 1

PR #1063, opened 2022-10-30
clang-tidy: performance-noexcept-move-constructor

jpr42:
I ran GSL through clang-tidy with the performance-* checks

https://releases.llvm.org/15.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/performance/noexcept-move-constructor.html

### DIFF
--- a/include/gsl/pointers
+++ b/include/gsl/pointers
@@ -267,7 +267,9 @@ public:
     constexpr strict_not_null(const strict_not_null<U>& other) : not_null<T>(other)
     {}
 
-    strict_not_null(strict_not_null&& other) = default;
+    // To avoid invalidating the "not null" invariant, the contained pointer is actually copied
+    // instead of moved. If it is a custom pointer, its constructor could in theory throw exceptions.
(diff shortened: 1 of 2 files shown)


include/gsl/pointers
```diff
@@ -267,7 +267,7 @@ public:
     constexpr strict_not_null(const strict_not_null<U>& other) : not_null<T>(other)
     {}
 
-    strict_not_null(strict_not_null&& other) = default;
+    strict_not_null(strict_not_null&& other) noexcept = default;
```
dmitrykobets-msft, line 270:
```suggestion
    // To avoid invalidating the "not null" invariant, the contained pointer is actually copied
    // instead of moved. If it is a custom pointer, its constructor could in theory throw exceptions.
    strict_not_null(strict_not_null&& other) noexcept(std::is_nothrow_copy_constructible<T>::value) = default;
```

The current behavior is to always copy the contained pointer when invoking this move constructor, to avoid violating the not_null guarantee. For the time being, let's mark this as conditionally no-except, and I'll surface this in the next maintainer's sync.

dmitrykobets-msft:
Looks good, thanks for helping cleanup the codebase!

PR #1067, opened 2022-12-13
Remove null-check Expects assertion inside not_null::get

### LINKED ISSUE #1051: Remove non-essential null-checking from gsl::not_null
We are thinking of revising the gsl::not_null threat-model to reserve all nullness checking to the constructor, and to trust all created instances post-construction:
```cpp
not_null<int*> a = some_pointer; // check that some_pointer is not null in a's constructor
int* b = a.get(); // no need to check that `a.ptr_` is not null; it's already been checked in a's constructor
not_null<int*> other = a; // no need to check that `a.ptr_` is not null; it's already been checked in a's constructor.  
                          // transitively, `other._ptr_` will now also be trusted
```

This change will remove many existing runtime checks that occur inside `gsl::not_null`, notably the one inside `not_null::get()`, and will overall simplify the design of the `gsl::not_null` type.

dmitrykobets-msft:
[Guidelines issue 2006](https://github.com/isocpp/CppCoreGuidelines/issues/2006) removes the null check inside `not_null::get`, since the contained pointer is already guaranteed to be not-null upon construction. 

Resolves #1051

### DIFF
--- a/include/gsl/pointers
+++ b/include/gsl/pointers
@@ -117,7 +117,6 @@ public:
     not_null& operator=(const not_null& other) = default;
     constexpr details::value_or_reference_return_t<T> get() const
     {
-        Ensures(ptr_ != nullptr);
         return ptr_;
     }
 


hsutter (approved):
Looks good!

3 test are failing but that must be for other reasons.

PR #1072, opened 2022-12-23
gsl::at variadic version

JaroPaska:
[SL.Con.3](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rsl-bounds) encourages the use bounds-checked functions such as `.at()` and `gsl::at()` over the use of `operator[]`. However, this becomes inconvenient when trying to access values inside multi-dimensional arrays.

For example, assume we have a 3-dimensional array named memo, which caches the values for some expensive function. To access a value, we would need to write `memo.at(x).at(y).at(z)`, or `at(at(at(memo, x), y), z)` when using the GSL function. By using the new variadic template, we can simply write `at(memo, x, y, z)`.

### DIFF
--- a/include/gsl/util
+++ b/include/gsl/util
@@ -144,6 +144,13 @@ constexpr auto at(std::span<T, extent> sp, const index i) -> decltype(sp[sp.size
     return sp[gsl::narrow_cast<size_t>(i)];
 }
 #endif // __cpp_lib_span >= 202002L
+
+template <class Cont, class... Args>
+constexpr decltype(auto) at(Cont&& cont, const index i, Args&&... args)
+{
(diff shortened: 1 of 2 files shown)


JaroPaska:
@microsoft-github-policy-service agree

beinhaerter:
Related: https://github.com/isocpp/CppCoreGuidelines/issues/2013

dmitrykobets-msft:
Hi @JaroPaska, thanks for the PR. I'll wait for the decision on the related CoreGuidelines issue.

dmitrykobets-msft:
closing as per the coreguidelines issue.

Issue #1075, opened 2022-12-27
Wrong Expects in gsl::at?


beinhaerter:
I was just looking at the `gsl::at` overloads and don't understand the `Expects` check. For example for the C style array version.
```
template <class T, std::size_t N>
    constexpr T& at(T (&arr)[N], const index i)
{
    Expects(i >= 0 && i < narrow_cast<index>(N));
    return arr[narrow_cast<std::size_t>(i)];
}
```

Checking that `i` is non-negative makes sense. But why is then `i` checked against `narrow_cast<index>(N)`? For example on a 32 bit platform when having an array with 0xff'ff'ff'ff elements and when accessing element `i==0` why would I want to check that `i<narrow_cast<index>(0xff'ff'ff'ff)` which is `i<-1`? Wouldn't it make more sense to check `narrow_cast<std::size_t>(i) < N`?
The same question applies to the other overloads of `gsl::at`.

hsutter:
Off the top of my head while on vacation:

I thought that having more elements than half the bytes of the address space generally isn't legal or practical anyway (IIRC of iterator requirements, pointer/iterator subtraction, I've paged out the exact reasons). So how about instead adding `static_assert(N < std::numeric_limits<std::size_t>::max()/2);`?

dmitrykobets-msft:
Hi @hsutter,

Is this your intended change?
```cpp
static_assert(N < std::numeric_limits<std::size_t>::max()/2-1);
Expects(i >= 0 && i < narrow_cast<index>(N));
return arr[narrow_cast<std::size_t>(i)];
```

Is the static assert therefore effectively checking that `N < std::numeric_limits<std::ptrdiff_t>::max()`, which therefore means that `i < narrow_cast<index>(N)` will never wrap?
And since `std::ptrdiff_t` holds a smaller positive value than `std::size_t`, the cast in the container access `return arr[narrow_cast<std::size_t>(i)];` is also fine?

If so, I wonder if it makes sense to simply take the suggested implementation of:
```cpp
Expects(i >= 0 && narrow_cast<std::size_t>(i) < N);
return arr[narrow_cast<std::size_t>(i)];
```
Since this is more straightforward and does not impose a static assertion which can (probably?) never be violated anyways.

hsutter:
> Is the static assert therefore effectively checking that `N < std::numeric_limits<std::ptrdiff_t>::max()`, which therefore means that `i < narrow_cast<index>(N)` will never wrap?

[...]

PR #1076, opened 2022-12-29
1075 Wrong Expects in gsl::at?

beinhaerter:
https://github.com/microsoft/GSL/issues/1075
- Add `static_assert` because we only support C style array `at` for up to half of the address space.
- Add `std::` before `size_t`.
- tests:
  - Add `#include <exception>`
  - Implement `span_tests` `interop_with_gsl_at` like `at_tests` `std_span`

### DIFF
--- a/include/gsl/util
+++ b/include/gsl/util
@@ -111,8 +111,8 @@ GSL_SUPPRESS(bounds.2) // NO-FORMAT: attribute
     // clang-format on
     constexpr T& at(T (&arr)[N], const index i)
 {
-    static_assert(N <= std::numeric_limits<std::size_t>::max() / 2, "We only support arrays up to half the bytes of the address space.");
-    Expects(i >= 0 && narrow_cast<std::size_t>(i) < N);
+    static_assert(N <= static_cast<std::size_t>(std::numeric_limits<std::ptrdiff_t>::max()), "We only support arrays up to half the bytes of the address space.");
+    Expects(i >= 0 && i < narrow_cast<index>(N));
     return arr[narrow_cast<std::size_t>(i)];
 }
 
@@ -123,8 +123,8 @@ GSL_SUPPRESS(bounds.2) // NO-FORMAT: attribute
     // clang-format on
     constexpr auto at(Cont& cont, const index i) -> decltype(cont[cont.size()])
 {
+    Expects(i >= 0 && i < narrow_cast<index>(cont.size()));
     using size_type = decltype(cont.size());
-    Expects(i >= 0 && narrow_cast<size_type>(i) < cont.size());
     return cont[narrow_cast<size_type>(i)];
 }
 
@@ -134,8 +134,7 @@ GSL_SUPPRESS(bounds.1) // NO-FORMAT: attribute
     // clang-format on
     constexpr T at(const std::initializer_list<T> cont, const index i)
 {
-    using size_type = decltype(cont.size());
-    Expects(i >= 0 && narrow_cast<size_type>(i) < cont.size());
+    Expects(i >= 0 && i < narrow_cast<index>(cont.size()));
     return *(cont.begin() + i);
 }
 
@@ -143,7 +142,7 @@ GSL_SUPPRESS(bounds.1) // NO-FORMAT: attribute
 template <class T, std::size_t extent = std::dynamic_extent>
 constexpr auto at(std::span<T, extent> sp, const index i) -> decltype(sp[sp.size()])
 {
-    Expects(i >= 0 && narrow_cast<std::size_t>(i) < sp.size());
+    Expects(i >= 0 && i < narrow_cast<index>(sp.size()));
     return sp[gsl::narrow_cast<std::size_t>(i)];
 }
 #endif // __cpp_lib_span >= 202002L
--- a/include/gsl/util
+++ b/include/gsl/util
@@ -111,7 +111,7 @@ GSL_SUPPRESS(bounds.2) // NO-FORMAT: attribute
     // clang-format on
     constexpr T& at(T (&arr)[N], const index i)
 {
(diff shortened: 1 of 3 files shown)


include/gsl/util
```diff
@@ -121,8 +123,8 @@ GSL_SUPPRESS(bounds.2) // NO-FORMAT: attribute
     // clang-format on
     constexpr auto at(Cont& cont, const index i) -> decltype(cont[cont.size()])
 {
-    Expects(i >= 0 && i < narrow_cast<index>(cont.size()));
     using size_type = decltype(cont.size());
+    Expects(i >= 0 && narrow_cast<size_type>(i) < cont.size());
```
dmitrykobets-msft, line 127: the way I interpret the assertion `static_assert(N <= std::numeric_limits<std::size_t>::max() / 2, "We only support arrays up to half the bytes of the address space.");` is that because `index` is defined as 
```cpp
// index type for all container indexes/subscripts/sizes
using index = std::ptrdiff_t;
```
then:
  a) we should not be supporting container access beyond this size. This is enforced by making the type of `i` be `index`.
  b) we should not be supporting containers that are larger than `std::ptrdiff_t` in size.
  
As per the discussion in https://github.com/microsoft/GSL/issues/1075, adding the `static_assert` for the array case is a helpful way of validating condition (b) at compile-time (in fact, I personally am more in favor of `static_assert(N <= static_cast<size_t>(std::numeric_limits<std::ptrdiff_t>::max())`). In the more general cases such as for `Cont`, I think it would make more sense to keep the `narrow_cast<index>(cont.size())` as a way of ensuring (b) at runtime. 

Imo reverting this change at the array case as well would help the code be more uniform, although not necessary due to the static_assert.

include/gsl/util
```diff
@@ -110,7 +111,8 @@ GSL_SUPPRESS(bounds.2) // NO-FORMAT: attribute
     // clang-format on
     constexpr T& at(T (&arr)[N], const index i)
 {
-    Expects(i >= 0 && i < narrow_cast<index>(N));
+    static_assert(N <= std::numeric_limits<std::size_t>::max() / 2, "We only support arrays up to half the bytes of the address space.");
+    Expects(i >= 0 && narrow_cast<std::size_t>(i) < N);
```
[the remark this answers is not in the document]
beinhaerter: I now reverted most of the changes. Do you still want the `static_assert` or shall the old `narrow_cast` persist?

tests/span_ext_tests.cpp
```diff
@@ -193,10 +194,18 @@ TEST(span_ext_test, make_span_from_container_constructor)
...
+        std::abort();
+    });
+    const auto expected = GetExpectedDeathString(terminateHandler);
+
     int arr[5] = {1, 2, 3, 4, 5};
     gsl::span<int> s{arr};
     EXPECT_TRUE(at(s, 0) == 1);
     EXPECT_TRUE(at(s, 1) == 2);
+    EXPECT_DEATH(at(s, 5), expected);
+    EXPECT_DEATH(at(s, 6), expected);
```
dmitrykobets-msft, line 208: adding a `-1` index test-case would add coverage to the `>= 0` condition

include/gsl/util
```diff
@@ -110,6 +111,7 @@ GSL_SUPPRESS(bounds.2) // NO-FORMAT: attribute
     // clang-format on
     constexpr T& at(T (&arr)[N], const index i)
 {
+    static_assert(N <= static_cast<std::size_t>(std::numeric_limits<std::ptrdiff_t>::max()), "We only support arrays up to half the bytes of the address space.");
```
dmitrykobets-msft, line 114: with the latest changes, the `static_assert` message is less clear - perhaps instead of "half the bytes..." it should just be "`PTRDIFF_MAX` elements" (or something similar)

tests/span_ext_tests.cpp
beinhaerter: I saw there `span_tests` `interop_with_gsl_at` and `at_tests` `std_span` differ. I think they should just do the same thing except one using `std::span` and the other using `gsl::span`.

include/gsl/util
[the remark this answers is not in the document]
beinhaerter: The original code used a `narrow_cast` so I did not change that.
When changing to `static_cast` I get a narrowing warning

> `gsl\util(116): warning C26472: Don't use a static_cast for arithmetic conversions. Use brace initialization, gsl::narrow_cast or gsl::narrow (type.1).

So that would require a `GSL_SUPPRESS(type.1)`. Are you sure you want to change to `static_cast` with `GSL_SUPPRESS` and not keep the `narrow_cast`?
If you want the `static_cast`, what about the `narrow_cast` in the other `at` overloads?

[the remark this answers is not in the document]
dmitrykobets-msft: ah that's unfortunate, it seems like the C26472 warning doesn't use the `>=0` to know that the cast is safe. In this case let's leave it as-is.

dmitrykobets-msft:
Looks good, thanks for being patient with the review.

diamante0018:
Hello, following the merging of this PR I get the following warning with latest MSVC (using GitHub's workflows) which causes a compilation error as all warnings are treated as errors in my workflow.
```
<my project>\deps\GSL\include\gsl\util(114,86): warning C4003: not enough arguments for function-like macro invocation 'max'
```

I can work around this by suppressing this warning specifically but please consider looking into this. Thanks.

dmitrykobets-msft:
@diamante0018 ah this seems to be a known issue with windows header users https://github.com/skypjack/entt/issues/343. I'll make a PR implementing the parentheses fix, but defining `NOMINMAX` as illustrated here (https://stackoverflow.com/questions/6884093/warning-c4003-not-enough-actual-parameters-for-macro-max-visual-studio-2010) should also work locally for you.

PR #1077, opened 2022-12-29
cleanup natvis

beinhaerter:
- `fail_fast`,`basic_string_span` and `basic_zstring_span` are not declared in the headers any more.
- Name the exporting header file for all types in `GSL.natvis`.


dmitrykobets-msft:
Looks good, thanks!

PR #1081, opened 2023-01-19
Fix `max` macro collision

dmitrykobets-msft:
PR https://github.com/microsoft/GSL/pull/1076 introduced a usage of `numeric_limits::max()` function, which seems to have [issues interacting with some Windows headers](https://github.com/skypjack/entt/wiki/Frequently-Asked-Questions#warning-c4003-the-min-the-max-and-the-macro). This PR implements [a simple fix](https://stackoverflow.com/questions/1394132/macro-and-member-function-conflict), which is to wrap the invocations in parentheses. This offloads the fix from the users of the library.


beinhaerter:
Thanks for fixing!

PR #1083, opened 2023-01-28
Raise CMake min to 3.14

jpr42:
- Explicitly test package manager support with CMake 3.14 and Ninja
- Remove gsl_install.cmake
- Simplify installation code


CMakeLists.txt
```diff
@@ -22,11 +22,28 @@ add_subdirectory(include)
...
+    set(namespace "Microsoft.GSL::")
+    set(cmake_files_install_dir ${CMAKE_INSTALL_DATADIR}/cmake/Microsoft.GSL)
+
+    install(TARGETS GSL EXPORT ${export_name} INCLUDES DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
+    install(EXPORT ${export_name} NAMESPACE ${namespace} DESTINATION ${cmake_files_install_dir})
+    export(TARGETS GSL NAMESPACE ${namespace} FILE ${export_name}.cmake)
+
+    set(gls_config_version "${CMAKE_CURRENT_BINARY_DIR}/Microsoft.GSLConfigVersion.cmake")
+
+    write_basic_package_version_file(${gls_config_version} COMPATIBILITY SameMajorVersion ARCH_INDEPENDENT)
```
jpr42, line 46: Reason for raising to 3.14
- To use ARCH_INDEPENDENT without condition (less for package managers to worry about)
- https://github.com/microsoft/GSL/issues/1084
- Simplifies writing github actions due to modern CLI (see cmake_find_package.yml)

cmake/gsl_install.cmake
```diff
@@ -1,26 +0,0 @@
...
-# Install header files
-install(DIRECTORY "${PROJECT_SOURCE_DIR}/include/gsl" DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
-
-set(export_name "Microsoft.GSLConfig")
-set(namespace "Microsoft.GSL::")
-set(cmake_files_install_dir ${CMAKE_INSTALL_DATADIR}/cmake/Microsoft.GSL)
-
-# Add find_package() support
-target_include_directories(GSL INTERFACE $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
-install(TARGETS GSL EXPORT ${export_name})
```
jpr42, line 13: This had been updated to

```
install(TARGETS GSL EXPORT ${export_name} INCLUDES DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
```

tests/CMakeLists.txt
```diff
@@ -1,13 +1,12 @@
...
 
 project(GSLTests LANGUAGES CXX)
-enable_testing()  # again, for support standalone testing
 
 set(GSL_CXX_STANDARD "14" CACHE STRING "Use c++ standard")
 
 set(CMAKE_CXX_STANDARD ${GSL_CXX_STANDARD})
 set(CMAKE_CXX_EXTENSIONS OFF)
-set(CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
```
jpr42, line 9: minor mistake

tests/CMakeLists.txt
```diff
@@ -57,11 +56,17 @@ if (NOT GTestMain_FOUND)
...
-    # import the globally installed Guidelines Support Library and test it
-    # instead of the current version from the include/ folder.
-    find_package(Microsoft.GSL REQUIRED)
+    find_package(Microsoft.GSL CONFIG REQUIRED)
+    enable_testing()
+
+    # Test version number support
+    if (NOT DEFINED Microsoft.GSL_VERSION)
+        message(FATAL_ERROR "Microsoft.GSL_VERSION not defined!")
+    endif()
```
jpr42, line 67: Test the version number produced by `write_basic_package_version_file`

.github/workflows/cmake_find_package.yml
```diff
@@ -0,0 +1,32 @@
...
+    name: Build ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ ubuntu-20.04, ubuntu-22.04, macos-11, macos-12 ]
+    steps:
+      - uses: actions/checkout@v3
+      - uses: lukka/get-cmake@latest
+        with:
+          cmakeVersion: 3.14.0
```
jpr42, line 20: Check your cmake_minimum_required with `get-cmake`
https://discourse.cmake.org/t/check-your-cmake-minimum-required-with-get-cmake/7257

.github/workflows/cmake_find_package.yml
```diff
@@ -0,0 +1,32 @@
+name: cmake_find_package
+
+on:
+    push:
+    pull_request:
+        branches:
+            - main
+
+jobs:
+  cmake-find-package:
```
dmitrykobets-msft, line 10: is this testing just `find_package` specifically, or the entire process of installing, building, and testing GSL?

tests/CMakeLists.txt
```diff
@@ -57,11 +56,17 @@ if (NOT GTestMain_FOUND)
     )
 endif()
 
+# cmake_find_package.yml explicitly tests find_package support used by package managers
```
dmitrykobets-msft, line 59: move comment to right above the `find_package` call?

tests/CMakeLists.txt
```diff
@@ -57,11 +56,17 @@ if (NOT GTestMain_FOUND)
...
 if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
-    # CMake has been started independently in this directory with tests. Do
-    # import the globally installed Guidelines Support Library and test it
-    # instead of the current version from the include/ folder.
-    find_package(Microsoft.GSL REQUIRED)
+    find_package(Microsoft.GSL CONFIG REQUIRED)
+    enable_testing()
+
+    # Test version number support
+    if (NOT DEFINED Microsoft.GSL_VERSION)
```
dmitrykobets-msft, line 65: this is coming from `find_package`, correct?

jpr42: Correct.

The find_package command will create the variable `<PackageName>_VERSION`.
In this case <PackageName> is `Microsoft.GSL`
https://cmake.org/cmake/help/latest/command/find_package.html?highlight=find_package#config-mode-version-selection

.github/workflows/cmake_find_package.yml
jpr42: 1.) It first tests installation.
2.) Tests the find_package support.
3.) It then builds/runs the tests to ensure the installation is 100% valid.

Part 3 is generally overkill but still helpful I think.

Also `actions` unlike `azure-pipelines` can be used in forks which can be helpful when working on a PR and not wanting to publish the PR until you are 100% ready.

.github/workflows/cmake_find_package.yml
```diff
@@ -0,0 +1,32 @@
+name: cmake_find_package
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
```
jpr42, line 7: Updated to match `android.yml` and `ios.yml`

.github/workflows/cmake_find_package.yml
dmitrykobets-msft: In that case, can this be renamed to something more general than `cmake_find_package.yaml`? This test seems to be performing an end-to-end test of installing GSL as a standalone project and running tests. Then

The comment `# cmake_find_package.yml explicitly tests find_package support used by package managers` inside `tests/CMakeLists.txt` emphasizes that this test is "explicitly" testing the `find_package` call but that just seems to be a by-product of the installation step.

Also out of curiosity, will replacing generator with something other than "Ninja" test anything different?

dmitrykobets-msft: (linking above comment to review)

jpr42: I've revised the script to just test `find_package` functionality.

> Also out of curiosity, will replacing generator with something other than "Ninja" test anything different?

For find_package testing I don't see a benefit.

dmitrykobets-msft:
Great, thanks for the fix @jpr42

PR #1085, opened 2023-02-06
add not_null and strict_not_null tests for const

beinhaerter:
While writing the docs I found a `remove_cv_t` in `make_null`/`make_strict_not_null` which made me wonder if we want to make sure that the code works well with const pointers.


dmitrykobets-msft (approved):
thanks, looks good

PR #1087, opened 2023-02-07
Remove include/CMakeLists.txt

jpr42:


include/CMakeLists.txt
```diff
@@ -1,13 +0,0 @@
-# Add include folders to the library and targets that consume it
-# the SYSTEM keyword suppresses warnings for users of the library
-#
-# By adding this directory as an include directory the user gets a
-# namespace effect.
-#
-# IE:
-#   #include <gsl/gsl>
```
jpr42, line 8: CI testing ensures this namespace effect is tested.

include/CMakeLists.txt
```diff
@@ -1,13 +0,0 @@
...
-if(PROJECT_IS_TOP_LEVEL)
-    target_include_directories(GSL INTERFACE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>)
-else()
-    target_include_directories(GSL SYSTEM INTERFACE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>)
```
jpr42, line 12: Using the SYSTEM keyword is not necessary anymore.

It was never needed for find_package support.

It only benefited `add_subdirectory` but now with CMake 3.25 it's not needed anymore.

```
New in version 3.25: If the SYSTEM argument is provided, the [SYSTEM](https://cmake.org/cmake/help/latest/prop_dir/SYSTEM.html#prop_dir:SYSTEM) directory property of the subdirectory will be set to true. This property is used to initialize the [SYSTEM](https://cmake.org/cmake/help/latest/prop_tgt/SYSTEM.html#prop_tgt:SYSTEM) property of each non-imported target created in that subdirectory.
```

jpr42: To be clear CMake 3.25 also lets you choose whether you even need the SYSTEM keyword anyway.

jpr42: Also using the SYSTEM keyword makes local GSL development more difficult since it ignores warnings from header files.

dmitrykobets-msft: the minimum-supported cmake version is 3.14 though, no?

jpr42: > the minimum-supported cmake version is 3.14 though, no?

Not sure where my head was with this PR... apologies. See new PR.

PR #1090, opened 2023-02-14
simplify to_byte

beinhaerter:
- to_byte_impl is not necessary, the same can be achieved with shorter code
- add test code for things that should not compile


dmitrykobets-msft:
thanks, looks good

PR #1093, opened 2023-02-17
add some span test for constructor from other span

beinhaerter:
More combinations with dynamic_extent or fixed extent for higher coverage.


dmitrykobets-msft:
thanks, looks good.

PR #1094, opened 2023-02-18
Better links to CoreGuidelines

beinhaerter:
The `#i13-do-not-pass-an-array-as-a-single-pointer` anchor seems to be auto generated based on the title of the section. The `#Ri-array` anchor is explicitely written in the source, so this should be a better link.


dmitrykobets-msft:
thanks, looks good

PR #1096, opened 2023-02-18
Extend docs and tests for span ctor

beinhaerter:


beinhaerter:
While documenting and thinking about the span ctor I wondered about this special case, so I wrote a test case for it and added the missing constraint to the docs.

dmitrykobets-msft:
thanks, looks good

Issue #1095, opened 2023-02-18
Purpose of `<span_ext>`


beinhaerter:
While writing the documentation for GSL I saw that it is not clear to me what the exact purpose of `<span_ext>` is. It cannot be used alone, you must `#include <span>` first. But `<span>` already does `#include <span_ext>`. What is the intent of the separation into two files? Do we still need/want two files or should we merge `<span_ext>` into `<span>`?

PR #1098, opened 2023-02-22
Update documentation of span_ext

dmitrykobets-msft:
`<gsl/span>` is intended to mirror `<span>` in the STL. `<span_ext>` provides convenience features that aren't part of `<span>`.
Address #1095


dmitrykobets-msft:
@beinhaerter does this address your concern in issue #1095?

beinhaerter:
Thanks, that makes it more clear.

PR #1099, opened 2023-03-13
Suppress warning C26481

beinhaerter:
Suppress "warning C26481: Don't use pointer arithmetic. Use span instead (bounds.1)." in the code that impements `span`.


dmitrykobets-msft:
thanks for the fix!

PR #1109, opened 2023-05-04
Only enable std::hash<gsl::not_null<T>> if std::hash<T> is enabled.

### LINKED ISSUE #914: std::hash<gsl::not_null<T>> should really be constrained
Ideally, `std::hash<gsl::not_null<T>>` would only be "enabled" (see [unord.hash]) if `std::hash<T>` is enabled.

This can be achieved very easily in C++20:

```diff
--- a/include/gsl/pointers
+++ b/include/gsl/pointers
@@ -175,6 +175,9 @@ not_null<T> operator+(std::ptrdiff_t, const not_null<T>&) = delete;
 namespace std
 {
 template <class T>
+#if __cpp_concepts
+  requires std::is_invocable_v<std::hash<T>, T>
+#endif
 struct hash<gsl::not_null<T>>
 {
     std::size_t operator()(const gsl::not_null<T>& value) const { return hash<T>{}(value); }
@@ -278,6 +281,9 @@ template <class T> strict_not_null(T) -> strict_not_null<T>;
 namespace std
 {
 template <class T>
+#if __cpp_concepts
+  requires std::is_invocable_v<std::hash<T>, T>
+#endif
 struct hash<gsl::strict_not_null<T>>
 {
     std::size_t operator()(const gsl::strict_not_null<T>& value) const { return hash<T>{}(value); }
```

For portability to older standards something like this works:

```diff
--- a/include/gsl/pointers
+++ b/include/gsl/pointers
@@ -170,14 +170,27 @@ not_null<T> operator+(const not_null<T>&, std::ptrdiff_t) = delete;
 template <class T>
 not_null<T> operator+(std::ptrdiff_t, const not_null<T>&) = delete;
 
+template <class T, class U = decltype(std::declval<T>().get()), bool enabled = std::is_invocable_v<std::hash<U>, U>>
+struct not_null_hash
+{
+    std::size_t operator()(const T& value) const { return std::hash<U>{}(value.get()); }
+};
+
+template <class T, class U>
+struct not_null_hash<T, U, false>
+{
+  not_null_hash() = delete;
+  not_null_hash(const not_null_hash&) = delete;
+  not_null_hash& operator=(const not_null_hash&) = delete;
+};
+
 } // namespace gsl
 
 namespace std
 {
 template <class T>
-struct hash<gsl::not_null<T>>
+struct hash<gsl::not_null<T>> : gsl::not_null_hash<gsl::not_null<T>>
 {
-    std::size_t operator()(const gsl::not_null<T>& value) const { return hash<T>{}(value); }
 };
 
 } // namespace std
@@ -278,9 +291,8 @@ template <class T> strict_not_null(T) -> strict_not_null<T>;
 namespace std
 {
 template <class T>
-struct hash<gsl::strict_not_null<T>>
+struct hash<gsl::strict_not_null<T>> : gsl::not_null_hash<gsl::strict_not_null<T>>
 {
-    std::size_t operator()(const gsl::strict_not_null<T>& value) const { return hash<T>{}(value); }
 };
 
 } // namespace std
```

dmitrykobets-msft:
Disable `std::hash<gsl::not_null<T>>` if `std::hash<T>` is not enabled.
Resolves #914

### DIFF
--- a/include/gsl/pointers
+++ b/include/gsl/pointers
@@ -216,7 +216,7 @@ template <class T>
 not_null<T> operator+(std::ptrdiff_t, const not_null<T>&) = delete;
 
 
-template <class T, class U = decltype(std::declval<T>().get()), typename = std::enable_if_t<std::is_default_constructible_v<std::hash<U>>>>
+template <class T, class U = decltype(std::declval<T>().get()), typename = std::enable_if_t<std::is_default_constructible<std::hash<U>>::value>>
 struct not_null_hash
 {
     std::size_t operator()(const T& value) const { return std::hash<U>{}(value.get()); }
--- a/include/gsl/pointers
+++ b/include/gsl/pointers
@@ -223,7 +223,7 @@ struct not_null_hash
 };
 
 template <class T, class U>
-struct not_null_hash<T, U, std::false_type>
+struct not_null_hash<T, U, false>
 {
     not_null_hash() = delete;
     not_null_hash(const not_null_hash&) = delete;
--- a/include/gsl/pointers
+++ b/include/gsl/pointers
@@ -216,7 +216,7 @@ template <class T>
 not_null<T> operator+(std::ptrdiff_t, const not_null<T>&) = delete;
 
 
-template <class T, class U = decltype(std::declval<T>().get()), typename = std::enable_if_t<std::is_default_constructible<std::hash<U>>::value>>
+template <class T, class U = decltype(std::declval<const T&>().get()), bool = std::is_default_constructible<std::hash<U>>::value>
 struct not_null_hash
 {
     std::size_t operator()(const T& value) const { return std::hash<U>{}(value.get()); }
--- a/include/gsl/pointers
+++ b/include/gsl/pointers
@@ -215,14 +215,28 @@ not_null<T> operator+(const not_null<T>&, std::ptrdiff_t) = delete;
 template <class T>
 not_null<T> operator+(std::ptrdiff_t, const not_null<T>&) = delete;
 
+
+template <class T, class U = decltype(std::declval<T>().get()), typename = std::enable_if_t<std::is_default_constructible<std::hash<U>>::value>>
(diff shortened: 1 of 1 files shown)


include/gsl/pointers
```diff
@@ -215,14 +215,28 @@ not_null<T> operator+(const not_null<T>&, std::ptrdiff_t) = delete;
 template <class T>
 not_null<T> operator+(std::ptrdiff_t, const not_null<T>&) = delete;
 
+
+template <class T, class U = decltype(std::declval<T>().get()), typename = std::enable_if_t<std::is_default_constructible<std::hash<U>>::value>>
```
CaseyCarter, line 219: `std::enable_if_t<std::is_default_constructible<std::hash<U>>::value>` is `void` when `std::is_default_constructible<std::hash<U>>::value` is `true` and ill-formed when `std::is_default_constructible<std::hash<U>>::value` is `false`. It is notably never `std::false_type` to match the partial specialization on line 226. I think you want this parameter to be `bool = std::is_default_constructible<std::hash<U>>::value`, and you want to match the value `false` on line 226.

Nitpick: `std::declval<T>()` here disagrees with the actual usage on line 22 that calls `.get()` on a `const T` lvalue.

```suggestion
template <class T, class U = decltype(std::declval<const T&>().get()), bool = std::is_default_constructible<std::hash<U>>::value>
```

include/gsl/pointers
```diff
@@ -215,14 +215,28 @@ not_null<T> operator+(const not_null<T>&, std::ptrdiff_t) = delete;
...
+struct not_null_hash
+{
+    std::size_t operator()(const T& value) const { return std::hash<U>{}(value.get()); }
+};
+
+template <class T, class U>
+struct not_null_hash<T, U, std::false_type>
```
CaseyCarter, line 226: (See comment above)
```suggestion
struct not_null_hash<T, U, false>
```

PR #1112, opened 2023-05-10
Install GSL.natvis

### LINKED ISSUE #1084: [cmake] GSL.natvis issues
Minor problems with the GLS.natvis CMake support

- [ ] find_package isn't supported (IE vpckg, conan, system package managers) don't benefit from GSL.natvis
  - GSL.natvis isn't installed
- [x] Ninja on MSVC isn't supported 
  - https://discourse.cmake.org/t/does-adding-natvis-files-via-target-sources-work-on-ninja/7316
  - This can be solved elegantly with target_link_options (3.13+)

jpr42:
Turns out supporting GSL.natvis perfectly is quite difficult.

There is no universal way to consume .natvis files that would satisfy everyone.

I thought the solution was to use the /NATVIS linker option. But it turns out that actually embeds information into the PDB. So I'm not sure how to properly support the Ninja generator...

That's not even accounting for the fact target_link_options doesn't play nicely with /NATVIS

When you just add the file via target_sources the visual studio solution will just pick it up and recognize it without adding it to the PDB file. Which won't affect the binary and is what most devs want.

This all comes down to the fact that /NATVIS files have native integration with visual studio that makes it difficult to use with non-visual studio solutions. /NATVIS almost works but embeds itself into the PDB which not everyone wants, and not everyone generates PDBs either.

Docs for natvis files and /NATVIS
- https://learn.microsoft.com/en-us/visualstudio/debugger/create-custom-views-of-native-objects?view=vs-2022
- https://learn.microsoft.com/en-us/cpp/build/reference/natvis-add-natvis-to-pdb?view=msvc-170

So my current solution is to just simplify the existing CMake code, and install the natvis so the user can decide.

closes #1084


CMakeLists.txt
```diff
@@ -19,8 +17,7 @@ target_compile_features(GSL INTERFACE "cxx_std_14")
 # Setup include directory
 add_subdirectory(include)
 
-# Add natvis file
-gsl_add_native_visualizer_support()
+target_sources(GSL INTERFACE $<BUILD_INTERFACE:${GSL_SOURCE_DIR}/GSL.natvis>)
```
jpr42, line 20: This does nothing on all platforms except for when generating a visual studio project.

In which case you get the .natvis experience that doesn't affect the PDB.

CMakeLists.txt
```diff
@@ -46,4 +43,6 @@ if (GSL_INSTALL)
     write_basic_package_version_file(${gls_config_version} COMPATIBILITY SameMajorVersion ARCH_INDEPENDENT)
 
     install(FILES ${gls_config_version} DESTINATION ${cmake_files_install_dir})
+
+    install(FILES GSL.natvis DESTINATION ${cmake_files_install_dir})
```
jpr42, line 47: Installing the file at least gives find_package users a chance to decide what they want to do. Versus nothing which is what they currently get.

cmake/gsl_functions.cmake
```diff
@@ -1,20 +0,0 @@
...
-# Adding the GSL.natvis files improves the debugging experience for users of this library.
-function(gsl_add_native_visualizer_support)
-    if (MSVC_IDE)
-        option(GSL_VS_ADD_NATIVE_VISUALIZERS "Configure project to use Visual Studio native visualizers" TRUE)
-    else()
-        set(GSL_VS_ADD_NATIVE_VISUALIZERS FALSE CACHE INTERNAL "Native visualizers are Visual Studio extension" FORCE)
-    endif()
-
-    # add natvis file to the library so it will automatically be loaded into Visual Studio
-    if(GSL_VS_ADD_NATIVE_VISUALIZERS)
```
jpr42, line 17: This if statement isn't really needed. Just complicates the code IMO. I don't see a reason to keep it.

dmitrykobets-msft:
@jpr42 Thanks a lot for looking into this! Unfortunate that natvis support is so limited, but the cleanup you've done to work around it looks good.

PR #1123, opened 2023-06-27
make value_or_reference_return_t non-const T or const T&

beinhaerter:
Returning a `const T` is not better than returning a `T`. In fact it is worse because it prevents moving the returned temporary. So let `value_or_reference_return_t` be either `T` or `const T&`, and not `const T` or `const T&` (which would be a good match for an in parameter, but not for a return value).

### DIFF
--- a/include/gsl/pointers
+++ b/include/gsl/pointers
@@ -47,14 +47,15 @@ namespace details
     {
     };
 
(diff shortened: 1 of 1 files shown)


beinhaerter:
I just saw https://github.com/microsoft/GSL/issues/550#issuecomment-1267448863, so I'll close this PR.

PR #1126, opened 2023-07-03
add missing include

beinhaerter:
The header file uses `std::declval`, so it needs to `#include <utility>`.


dmitrykobets-msft:
Thanks for the cleanup, merging

PR #1127, opened 2023-07-03
get back gcc 8.4 compatibility

beinhaerter:
Before my PR #1122 `gsl/pointers` was gcc 8.4 compatible. Now it is not. This commit makes it compatible with gcc 8.4 again.

### DIFF
--- a/include/gsl/pointers
+++ b/include/gsl/pointers
@@ -117,7 +117,7 @@ public:
     not_null(const not_null& other) = default;
     not_null& operator=(const not_null& other) = default;
     constexpr details::value_or_reference_return_t<T> get() const
-        noexcept(noexcept(details::value_or_reference_return_t<T>{ptr_}))
+        noexcept(noexcept(details::value_or_reference_return_t<T>{std::declval<T&>()}))
     {
         return ptr_;
     }


include/gsl/pointers
```diff
@@ -117,7 +117,7 @@ public:
     not_null(const not_null& other) = default;
     not_null& operator=(const not_null& other) = default;
     constexpr details::value_or_reference_return_t<T> get() const
-        noexcept(noexcept(details::value_or_reference_return_t<T>{ptr_}))
+        noexcept(noexcept(details::value_or_reference_return_t<T>{std::declval<T>()}))
```
dmitrykobets-msft, line 120:
```suggestion
        noexcept(noexcept(details::value_or_reference_return_t<T>{std::declval<T&>()}))
```
As implemented this will check against T's move-ctor. `T&` should force the check on the copy-ctor instead, which is what would be invoked in the case that `value_or_reference_return_t` gives the non-reference `const T`

beinhaerter: Thanks for pointing that out. I already wondered, because `std::declval` is returning an rvalue reference.
Regarding your change I do not see why that would make a difference. https://en.cppreference.com/w/cpp/utility/declval says

> template< class T >
> typename std::add_rvalue_reference<T>::type declval() noexcept;

Aren't `std::add_rvalue_reference<T>::type` and `std::add_rvalue_reference<T&>::type` identical?
Wouldn't we need something like the following?

    noexcept(noexcept(details::value_or_reference_return_t<T>{static_cast<T>(std::declval<T>())}))

dmitrykobets-msft: `std::add_rvalue_reference` seems to be more nuanced (https://en.cppreference.com/w/cpp/types/add_reference):
> 2) If T is a function type that has no cv- or ref- qualifier or an object type, provides a member typedef type which is T&&, otherwise type is T.

So in `std::add_rvalue_reference<T&>::type`, since `T&` is not an object type (nor a function type), this would result in just `T&`,
whereas `std::add_rvalue_reference<T>::type` would give back `T&&`.

beinhaerter: Thanks for clarifying. I just believed in what the name of `add_rvalue_reference` promises, I did not read what exactly it does.

dmitrykobets-msft:
thanks for the patch, merging

PR #1131, opened 2023-08-24
fix and optimize documentation

beinhaerter:
- Add anchor for `finally` in `headers.md` so that the link in `README.md` can work
- In `README.md` add code formatting for the character types of the `zstring` and `string_span` types
- In `README.md` change code formatting to links for GSL types
- Vertical alignment


dmitrykobets-msft:
Thanks for the fix!

PR #1133, opened 2023-09-02
Update Clang GSL_SUPPRESS to stringize parameter instead of using fixed string literal.

### LINKED ISSUE #1130: `GSL_SUPPRESS(y)` expands to `[[gsl::suppress("x")]]` for clang
Here's the definition of `GSL_SUPPRESS`:

https://github.com/microsoft/GSL/blob/4300304ef24c247b3db0255763f46b9f95c3a83d/include/gsl/assert#L48-L50

This always expands to `[[gsl::suppress("x")]]` because the `"x"` string literal won't be replaced.

We probably want something like this instead:
```c++
#define GSL_SUPPRESS(x) [[gsl::suppress( #x )]]
```

edgchen1:
Fix #1130.

### DIFF
--- a/include/gsl/assert
+++ b/include/gsl/assert
@@ -46,7 +46,7 @@
 // Hopefully temporary until suppression standardization occurs
 //
 #if defined(__clang__)
-#define GSL_SUPPRESS(x) [[gsl::suppress("x")]]
+#define GSL_SUPPRESS(x) [[gsl::suppress(#x)]]
 #else
 #if defined(_MSC_VER) && !defined(__INTEL_COMPILER) && !defined(__NVCC__)
 #define GSL_SUPPRESS(x) [[gsl::suppress(x)]]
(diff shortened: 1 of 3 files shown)


dmitrykobets-msft:
@edgchen1 it looks like `gsl/byte` also had a definition of `GSL_SUPPRESS`. After digging into it, the redefinition doesn't look necessary. I've created a pull request into your fork with the necessary changes to hopefully fix the failing tests.

edgchen1:
> @edgchen1 it looks like `gsl/byte` also had a definition of `GSL_SUPPRESS`. After digging into it, the redefinition doesn't look necessary. I've created a pull request [edgchen1#1](https://github.com/edgchen1/GSL/pull/1) into your fork with the necessary changes to hopefully fix the failing tests.

Thanks! Merged your fix.

dmitrykobets-msft:
Hi @edgchen1 sorry in the delay in getting this merged; our CI was having some issues that are now resolved.
Thanks for the fix, merging.

Issue #1134, opened 2023-09-03
gsl::not_null and std::variant are not fully compatible


mymedia2:
I suppose it is worth highlighting a lack of interoperability between gsl::not_null and std::variant from libstdc++. Consider this code.

```c++
#include <variant>
#include <gsl/pointers>

int main()
{
  int i = 0;
  std::variant< std::monostate, gsl::not_null<int*> > v;
  v.emplace< gsl::not_null<int*> >(&i);
}
```
https://godbolt.org/z/KTP9EdMzf &rarr; run the code above in the playground

It segfaults inside std::variant::emplace because of <https://gcc.gnu.org/bugzilla/show_bug.cgi?id=106547>. However, a workaround of that libstdc++'s bug may be added in GSL. Say if the `gsl::not_null<int*>` class had a noexcept move constructor, an optimized path without the bug could be chosen.

PR #1135, opened 2023-09-03
Mark not_null constructors as noexcept when underlying type can be moved with no exception

mymedia2:
This enables possible optimisations for trivial types. This also avoids a bug in std::variant::emplace from GNU's libstdc++.
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=106547

Closes: #1134


dmitrykobets-msft:
Hi @mymedia2, with a quick inspection this looks reasonable, but I'll defer this for the next maintainers' sync. Thanks for the PR.

hsutter:
Thanks!

beinhaerter:
The changes only apply to `not_null`, they are not applied to `strict_not_null`. Is there any reason for that? Or was this only overseen/forgotten?

PR #1142, opened 2023-12-23
ci: Update to `actions/checkout@v4`.

waywardmonkeys:
This fixes annotations on task runs like the following:

    The following actions uses node12 which is deprecated and will be forced
    to run on node16: actions/checkout@v2. For more info:
    https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/


hsutter:
Thanks!

PR #1154, opened 2024-07-16
Added the IWYU pragma export

hkroeg:
clang-tidy has added the check misc-include-cleaner with version 17. It checks for unused and missing includes. For instance, the following code will be flagged since the file gsl does not directly supply gsl::span
```c++
#include "gsl/include/gsl/gsl"

gsl::span<int>
```
The check misc-include-cleaner respects the IWYU pragmas. So, this pull request adds the IWYU pragmas.


carsonRadtke:
@hkroeg Thanks for the contribution!

PR #1158, opened 2024-10-11
[ci]: fix android job

carsonRadtke:
Job has been failing for months and is preventing other PRs from getting
merged. This PR fixes the pipeline - it changed the following:
 - runs on macos-latest-large (which is a non-arm image)
 - starts the emulator under java8 (see
   https://learn.microsoft.com/en-us/java/openjdk/transition-from-java-8-to-java-11)
 - runs emulator with -no-audio

The above were requirements to get the pipeline running. Other changes
include the following:
 - The job will auto-cancel when a newer commit is pushed to the
 PR.
 - The 'Wait for emulator ready' step now has a 2 minute timeout. This
   is necessary because we get no indication on whether or not the
   command that started the emulator suceeded. I believe there is room
   for improvement here, but it is outside the scope of the PR.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -1,4 +1,9 @@
 name: CI_Android
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 on:
   push:
     branches: [ main ]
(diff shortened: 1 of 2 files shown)


carsonRadtke:
@microsoft-github-policy-service agree company="Microsoft"

d-winsor (approved):

PR #1164, opened 2024-10-23
Simplify CMake test code

jpr42:
- Use FetchContent to grab googletest
- Simplify find_package testing so it's done as a part of ctest
- Use the GTest::gtest_main target directly


.github/workflows/cmake_find_package.yml
jpr42, line 1: This is now done when ctest is run. Which IMO is much more elegant.

Issue #1165, opened 2024-11-08
gsl::span performance is a bottleneck for adoption


tiagomacarios:
Office is trying to adopt gsl::span more broadly, but we are seeing some unexpected performance regressions when compiling with clang:
A - range-for loops
B - std algorithms

On a mail thread @StephanTLavavej suggested:

> I'd recommend reaching out to the libc++ devs on either the LLVM Discord server or their Discourse forum, about whether they have an ["unwrapping"](https://github.com/search?q=repo%3Amicrosoft%2FSTL%20_Unchecked_begin&type=code) system similar to what MSVC's STL has. (Presumably you'd want to do the same for libstdc++.)

> We use libc++'s portable subset of tests, and we're aligned on a few implementation things (backporting import std; to C++20, names of compiler builtins, etc.), but our internal mechanisms are wildly different - there is no commonality between our unwrapping mechanisms.

> In C++20 we can detect arbitrary contiguous iterators, which lets us extract raw pointers, but in C++14/17 that's not physically possible so iterators really need to opt in to unwrapping.

Could one of the maintainers please follow up?

carsonRadtke:
@tiagomacarios Thank you for reporting this issue.

After a brief look into libcxx's `std::span`, it looks like they call "special" iterators when [Safe Buffers](https://clang.llvm.org/docs/SafeBuffers.html) are enabled - these seem to make `std::span` analogous to `gsl::span`, but the [bounds checking gets optimized away with `-O2`](https://godbolt.org/z/z1WcxGhnv). I will start a discussion with clang folks to determine their thoughts on the matter and see if there is a protocol that `gsl::span` can use.

In the meantime, could you share what version of clang you are using?

tiagomacarios:
We use 3 different version of clang:

- general tooling - the latest release (19.1.0)
- [android ndk r27 LTS](https://github.com/android/ndk/wiki/Changelog-r27) - [3c92011b600bdf70424e2547594dd461fe411a41](https://github.com/llvm/llvm-project/commits/3c92011b600bdf70424e2547594dd461fe411a41) which seems to be 17.x plus some arbitrary commits
- [apple xcode 16.0](https://en.wikipedia.org/wiki/Xcode#:~:text=5.10.0.13%20clang%2D1500.3.9.4)-,16.0,-1021.4) which maps to 17.0.6

We usually update compilers as soon as new LTS versions are available.

tiagomacarios:
[...]

PR #1166, opened 2024-11-09
improve performance of span_iterator w/ clang

carsonRadtke:
Issue: #1165

Before this PR, the range-for loop was ~3300x slower. After this PR, it is ~1.005x slower

The clang optimizer is very good at optimizing `current != end`, so we changed to this idiom. This moves the Expects assertion into the constructor instead of on the hot-path which is called whenever either operator++ or operator* is called.

Note: The codegen for the assertion is still a missed optimization, but less worrisome as it only happens once per iterator.

### DIFF
--- a/include/gsl/span
+++ b/include/gsl/span
@@ -141,8 +141,8 @@ namespace details
         constexpr span_iterator(pointer begin, pointer end, pointer current)
             : begin_(begin), end_(end), current_(current)
         {
-            Expects(begin && current && end);
-            Expects(begin <= current && current <= end);
+            Expects(begin_ && current_ && end_);
+            Expects(begin_ <= current_ && current <= end_);
         }
 
         constexpr operator span_iterator<const Type>() const noexcept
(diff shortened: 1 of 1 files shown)


carsonRadtke:
Benchmark | CPU Before (ns) | CPU After (ns) | Ratio
-- | -- | -- | --
BM_IsSortedStdSpan | 325 | 323 | 1.0
BM_IsSortedGslSpan | 1573 | 322 | 4.9 
BM_IsSortedRangesStdSpan | 325 | 323 | 1.0
BM_IsSortedRangesGslSpan | 1583 | 475 | 3.3
BM_IsSortedCustomStdSpan | 442 | 435 | 1.0
BM_IsSortedCustomGslSpan | 1420 | 629 | 2.3
BM_MinElementAlgorithmStdSpan | 637 | 631 | 1.0
BM_MinElementAlgorithmGslSpan | 1913 | 1852 | 1.0
BM_MinElementRangeForStdSpan | 33.8 | 33.6 | 1.0
BM_MinElementRangeForGslSpan | 635 | 33.2 | 19.1

More benchmarks (thanks to @galenelias). I'd like to see an improvement in `std::min_element`, but I will open an issue for it instead of blocking this PR. 

It would also be nice to have performance regression tests (see https://github.com/microsoft/GSL/issues/1165#issuecomment-2469310972). I will open an issue for this as well.

PR #1171, opened 2024-11-15
fix failing macos test

carsonRadtke:
Looks like https://github.com/actions/runner-images/issues/10703 took away support for xcode 14. That is why the latest action failed. There an active PR adding support for xcode 16 to the macos runner (https://github.com/actions/runner-images/pull/10967). This commit will fail until that has been merged, but it is probably worth waiting for.


carsonRadtke:
merged into #1172

PR #1174, opened 2024-12-12
Fix SFINAE on `gsl::owner`.

lord-pando:
`std::enable_if_t` must not be used as a default template argument, otherwise the instantiator will be able to override it freely with something that doesn't fail substitution. Instead, `std::enable_if_t` itself must be the type of the template argument.

More information in the examples here: https://en.cppreference.com/w/cpp/types/enable_if


carsonRadtke:
Thanks for the PR!

This looks like something we overlooked that has been wrong for a while. Much appreciated :)

Issue #1162, opened 2024-10-21
`not_null<std::unique_ptr<>>` not working in v4.1.0 with g++-10


asartori86:
With the 4.1.0 release, the following snippet does not compile with g++-10

```c++
// filename: minimal.cc

#include "gsl/gsl"
#include <iostream>
#include <memory>

int main() {
  gsl::not_null<std::unique_ptr<int>> ptr = std::make_unique<int>(42);
  std::cout << *ptr << std::endl;
}
```
```bash
$ g++ minimal.cc -I GSL-4.1.0/include -std=c++20
In file included from GSL-4.1.0/include/gsl/gsl:24,
                 from minimal.cc:1:
GSL-4.1.0/include/gsl/pointers: In instantiation of ‘constexpr gsl::details::value_or_reference_return_t<T> gsl::not_null<T>::get() const [with T = std::unique_ptr<int>; gsl::details::value_or_reference_return_t<T> = const std::unique_ptr<int>&]’:
GSL-4.1.0/include/gsl/pointers:127:58:   required from ‘constexpr decltype(auto) gsl::not_null<T>::operator*() const [with T = std::unique_ptr<int>]’
minimal.cc:7:17:   required from here
GSL-4.1.0/include/gsl/pointers:120:18: error: use of deleted function ‘std::unique_ptr<_Tp, _Dp>::unique_ptr(const std::unique_ptr<_Tp, _Dp>&) [with _Tp = int; _Dp = std::default_delete<int>]’
  120 |         noexcept(noexcept(details::value_or_reference_return_t<T>{std::declval<T&>()}))
      |                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /usr/include/c++/10/memory:83,
                 from GSL-4.1.0/include/gsl/span:28,
                 from GSL-4.1.0/include/gsl/algorithm:21,
                 from GSL-4.1.0/include/gsl/gsl:21,
                 from minimal.cc:1:
/usr/include/c++/10/bits/unique_ptr.h:468:7: note: declared here
  468 |       unique_ptr(const unique_ptr&) = delete;
      |       ^~~~~~~~~~
```
```bash
$ g++ --version
g++ (Debian 10.2.1-6) 10.2.1 20210110
Copyright (C) 2020 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
[...]

PR #1175, opened 2024-12-13
fix: direct-init const ref instead of list-init

carsonRadtke:
fixes: https://github.com/microsoft/GSL/issues/1162

gcc has a bug that generates a call to the copy constructor when list initializing a const reference. This PR offers a workaround which is to direct initialize the value.

see: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=117900


carsonRadtke:
This seems to work locally. I will rebase and merge https://github.com/microsoft/GSL/pull/1160 which adds a test for this.

PR #1176, opened 2024-12-18
Force unit tests for byte to use GSL implementation

beinhaerter:
When setting `GSL_USE_STD_BYTE` to zero, then the unit test does not compile because `byte` is ambiguous (could be `std::byte` or `gsl::byte`). So `gsl::` prefix is needed for `byte`.
It looks like the unit tests never ran on a platform where `gsl::byte` is not based on `std::byte`.
It does not make much sense to test `std::byte` for compliance, so make the unit tests based on the GSL implementation of `gsl::byte` (`#define GSL_USE_STD_BYTE 0`).

### DIFF
--- a/tests/byte_tests.cpp
+++ b/tests/byte_tests.cpp
@@ -16,6 +16,7 @@
 
 #include <gtest/gtest.h>
 
+#define GSL_USE_STD_BYTE 0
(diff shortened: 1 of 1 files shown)


carsonRadtke:
Good find! Thanks for the fix :)

beinhaerter:
Wow, that was a quick merge. It landed before I completed my other pull request. Thanks!

PR #1177, opened 2024-12-18
Better use of std::enable_if

beinhaerter:
Replace all occurances of `class = std::enable_if_t<Cond>` and `typename = std::enable_if_t<Cond>` with `std::enable_if_t<Cond, bool> = true`.

This commit is inspired by #1174, which changed one occurance in the owner header. This commit is aimed to fix all remaining occurances.

### DIFF
--- a/docs/headers.md
+++ b/docs/headers.md
@@ -91,16 +91,16 @@ See [SL.str.5: Use `std::byte` to refer to byte values that do not necessarily r
 ### Non-member functions
 
 ```cpp
-template <class IntegerType, class = std::enable_if_t<std::is_integral<IntegerType>::value>>
+template <class IntegerType, std::enable_if_t<std::is_integral<IntegerType>::value, bool> = true>
 constexpr byte& operator<<=(byte& b, IntegerType shift) noexcept;
 
-template <class IntegerType, class = std::enable_if_t<std::is_integral<IntegerType>::value>>
+template <class IntegerType, std::enable_if_t<std::is_integral<IntegerType>::value, bool> = true>
 constexpr byte operator<<(byte b, IntegerType shift) noexcept;
 
-template <class IntegerType, class = std::enable_if_t<std::is_integral<IntegerType>::value>>
+template <class IntegerType, std::enable_if_t<std::is_integral<IntegerType>::value, bool> = true>
 constexpr byte& operator>>=(byte& b, IntegerType shift) noexcept;
 
-template <class IntegerType, class = std::enable_if_t<std::is_integral<IntegerType>::value>>
+template <class IntegerType, std::enable_if_t<std::is_integral<IntegerType>::value, bool> = true>
 constexpr byte operator>>(byte b, IntegerType shift) noexcept;
 ```
 
@@ -134,7 +134,7 @@ constexpr byte operator~(byte b) noexcept;
 Bitwise negation of a `byte`. Flips all bits. Zeroes become ones, ones become zeroes.
 
 ```cpp
-template <class IntegerType, class = std::enable_if_t<std::is_integral<IntegerType>::value>>
+template <class IntegerType, std::enable_if_t<std::is_integral<IntegerType>::value, bool> = true>
 constexpr IntegerType to_integer(byte b) noexcept;
 ```
 
@@ -310,7 +310,7 @@ auto make_not_null(T&& t) noexcept;
 Creates a `gsl::not_null` object, deducing the target type from the type of the argument.
 
 ```cpp
-template <typename T, typename = std::enable_if_t<std::is_move_assignable<T>::value && std::is_move_constructible<T>::value>>
+template <typename T, std::enable_if_t<std::is_move_assignable<T>::value && std::is_move_constructible<T>::value, bool> = true>
 void swap(not_null<T>& a, not_null<T>& b);
 ```
 
--- a/include/gsl/byte
+++ b/include/gsl/byte
@@ -91,25 +91,25 @@ enum class byte_may_alias byte : unsigned char
 {
 };
 
-template <class IntegerType, class = std::enable_if_t<std::is_integral<IntegerType>::value>>
+template <class IntegerType, std::enable_if_t<std::is_integral<IntegerType>::value, bool> = true>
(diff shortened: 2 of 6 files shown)


docs/headers.md
```diff
@@ -91,16 +91,16 @@ See [SL.str.5: Use `std::byte` to refer to byte values that do not necessarily r
...
-template <class IntegerType, class = std::enable_if_t<std::is_integral<IntegerType>::value>>
+template <class IntegerType, std::enable_if_t<std::is_integral<IntegerType>::value, bool> = true>
 constexpr byte operator<<(byte b, IntegerType shift) noexcept;
 
-template <class IntegerType, class = std::enable_if_t<std::is_integral<IntegerType>::value>>
+template <class IntegerType, std::enable_if_t<std::is_integral<IntegerType>::value, bool> = true>
 constexpr byte& operator>>=(byte& b, IntegerType shift) noexcept;
 
-template <class IntegerType, class = std::enable_if_t<std::is_integral<IntegerType>::value>>
+template <class IntegerType, std::enable_if_t<std::is_integral<IntegerType>::value, bool> = true>
```
carsonRadtke, line 103: I am hesitant to accept these operator changes because I don't believe they are "dangerous". From my experiments it is impossible to override the `std::enable_if_t` with an arbitrary type.

If I'm wrong, could you give me an example that compiles where the shift is a non-integer type?

docs/headers.md
```diff
@@ -134,7 +134,7 @@ constexpr byte operator~(byte b) noexcept;
 Bitwise negation of a `byte`. Flips all bits. Zeroes become ones, ones become zeroes.
 
 ```cpp
-template <class IntegerType, class = std::enable_if_t<std::is_integral<IntegerType>::value>>
+template <class IntegerType, std::enable_if_t<std::is_integral<IntegerType>::value, bool> = true>
```
carsonRadtke, line 137: This is a change I am willing to accept because it makes the following example not compile:
```c++
#define GSL_USE_STD_BYTE 0
#include <gsl/gsl>
static_assert(std::is_floating_point_v<decltype(to_integer<float, float>(gsl::byte(0)))>, "oof");
```
(which [currently compiles](https://godbolt.org/z/1efd7sxsa) on `main`)

docs/headers.md
```diff
@@ -229,17 +229,17 @@ See [F.23: Use a `not_null<T>` to indicate that “null” is not a valid value]
...
-template <typename = std::enable_if_t<!std::is_same<std::nullptr_t, T>::value>>
+template <std::enable_if_t<!std::is_same<std::nullptr_t, T>::value, bool> = true>
 constexpr not_null(T u);
 ```
 
 Constructs a `gsl_owner<T>` from a pointer that is convertible to `T` or that is a `T`. It [`Expects`](#user-content-H-assert-expects) that the provided pointer is not `== nullptr`.
 
 ```cpp
-template <typename U, typename = std::enable_if_t<std::is_convertible<U, T>::value>>
+template <typename U, std::enable_if_t<std::is_convertible<U, T>::value, bool> = true>
```
carsonRadtke, line 242: I can't figure out how to bypass any of these `enable_if_t`s. (Though I have little confidence my attempts were exhaustive).

Could you provide an example of why this change is necessary?

docs/headers.md
```diff
@@ -310,7 +310,7 @@ auto make_not_null(T&& t) noexcept;
 Creates a `gsl::not_null` object, deducing the target type from the type of the argument.
 
 ```cpp
-template <typename T, typename = std::enable_if_t<std::is_move_assignable<T>::value && std::is_move_constructible<T>::value>>
+template <typename T, std::enable_if_t<std::is_move_assignable<T>::value && std::is_move_constructible<T>::value, bool> = true>
```
carsonRadtke, line 313: This is a change I am willing to accept. I missed this with my latest PR.

carsonRadtke (changes requested):
I had originally considered asking for this change as part of #1174, but I was worried there may be subtle differences that need to be considered for each change.

From my point of view, I would like justification for each change (i.e. an example of accepted code that shouldn't be accepted). If the `std::enable_if_t` is impossible to override, I don't think the change is necessary. That being said, I will bring this PR up during the monthly maintainers sync (which happens to be this evening) to get a second opinion.

docs/headers.md
beinhaerter: I started my changes with byte and added tests in `TEST(byte_tests, bitwise_operations )` like this:
```
operator<< <float,void>(a, float{1});
operator>> <float,void>(a, float{1});
operator<<= <float,void>(a, float{1});
operator>>= <float,void>(a, float{1});
```
These look totally strange, and they do trigger a compile error, but the error is inside the operator in byte. With the change the compile error is at the call site (no matching operator found).

carsonRadtke: Ah! I didn't consider that. If these changes make compiler errors easier to understand, then I am all for it. I will re-evaluate with this point in mind.

docs/headers.md
beinhaerter: I'll try to come up with one example per change.

beinhaerter:
I did some investigation. It seems not be possible to explicitely choose the template instantiation for a templated constructor in a class template, so my changes for `not_null` and `span` were not useful, I removed them.

For the other changes I split my PR into three commits:
1. add the tests that should fail but that do not fail
2. do the code changes that make the tests fail as expected
3. revert all changes from the first commit

I would have preferred to keep the tests from the first commits, but my SFINAE skills are too low. I tried it the way `span_compatibility_tests` is doing it, but I did not get it working.

beinhaerter:
I am now happy with the PR. I was able to write tests like `static_assert(XXCompilesFor<YY>, "");` that work for the code before the fix and where negating the condition like `static_assert(!XXCompilesFor<YY>, "");` works after the fix. So we can now keep the tests in place and get rid of the third commit that deleted the tests.

carsonRadtke:
Thanks for adding examples and it is awesome that they are now part of the tests. I will take another look once the tests are green for all compilers.

FYI - I made a change in the repo settings; the actions should now automatically run for your PRs.

carsonRadtke:
This looks great! Thanks for making all those changes.

PR #1179, opened 2024-12-23
strict_not_null for unique_ptr

beinhaerter:
- `strict_not_null<std::unique_ptr<int>>{ std::make_unique<int>()}` failed to compile
  - `strict_not_null` ctor needs to move the passed `unique_ptr`, not copy
  - Copied `not_null` `TestNotNullConstructors` for `strict_not_null`
- The `noexcept` specifiers on the `strict_not_null` and `not_null` constructors were out of sync.
- Added unit test for `not_null<unique_ptr<T>>` and for `strict_not_null<unique_ptr<T>>`
- Added unit test for `gsl::swap` for two `strict_not_null`
- Added unit test for `gsl::swap` for `not_null` and `strict_not_null`

### DIFF
--- a/tests/strict_notnull_tests.cpp
+++ b/tests/strict_notnull_tests.cpp
@@ -21,6 +21,15 @@
 
 using namespace gsl;
 
+// stand-in for a user-defined ref-counted class
+template <typename T>
+struct RefCounted
+{
+    RefCounted(T* p) : p_(p) {}
+    operator T*() { return p_; }
+    T* p_;
+};
+
 namespace
 {
 // clang-format off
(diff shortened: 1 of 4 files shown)


tests/strict_notnull_tests.cpp
```diff
@@ -43,12 +52,116 @@ GSL_SUPPRESS(f.4) // NO-FORMAT: attribute
 // clang-format on
 bool strict_helper_const(strict_not_null<const int*> p) { return *p == 12; }
 
-#ifdef CONFIRM_COMPILATION_ERRORS
 int* return_pointer() { return nullptr; }
+#ifdef CONFIRM_COMPILATION_ERRORS
```
carsonRadtke, line 56: This is unfamiliar and I can't figure out who/what sets it. Can you explain `CONFIRM_COMPILATION_ERRORS` to me?

carsonRadtke (approved):
Thanks for fixing the issue and adding well-written tests. Everything looks good!

I will merge once the conflicts are resolved.

tests/strict_notnull_tests.cpp
carsonRadtke: [nevermind](https://github.com/microsoft/GSL/issues/1130#issuecomment-1699857287)

beinhaerter: If you want, I can write test that check for non-compiling constructs like the ones written in the enable_if-PR. That could be done step by step (whenever I have time) for all the stuff that is right now hidden in those `#ifdef CONFIRM_COMPILATION_ERRORS`.

carsonRadtke: It would be good to have regression tests that make sure we reject invalid uses where possible.

beinhaerter: First shot: https://github.com/microsoft/GSL/pull/1180

PR #1180, opened 2024-12-28
More checks for non-compilable code, plus fix for span

beinhaerter:
- Bugfix: templated versions of span::first(), span::last() and span::subspan() now fail to compile if more elements than available are being extracted from a static sized span; previously it compiled and `Expects` failed during runtime
- Remove some of the checks in `#ifdef CONFIRM_COMPILATION_ERRORS` and add compile time checks for these tests.
- Acitvate some commented out checks.
- Activate some deactivated checks depending on the platform
- Make some death checks simpler to only check the function call that is needed to trigger the death.
- clang-format

### DIFF
--- a/include/gsl/span
+++ b/include/gsl/span
@@ -582,7 +582,7 @@ public:
     template <std::size_t Count>
     constexpr span<element_type, Count> first() const noexcept
     {
-        static_assert(Count <= Extent,
+        static_assert(Extent == dynamic_extent || Count <= Extent,
                       "first() cannot extract more elements from a span than it contains.");
         Expects(Count <= size());
         return span<element_type, Count>{data(), Count};
@@ -594,7 +594,7 @@ public:
         // clang-format on
         constexpr span<element_type, Count> last() const noexcept
     {
-        static_assert(Count <= Extent,
+        static_assert(Extent == dynamic_extent || Count <= Extent,
                       "last() cannot extract more elements from a span than it contains.");
         Expects(Count <= size());
         return span<element_type, Count>{data() + (size() - Count), Count};
@@ -607,7 +607,8 @@ public:
         constexpr auto subspan() const noexcept ->
         typename details::calculate_subspan_type<ElementType, Extent, Offset, Count>::type
     {
-        static_assert(Extent >= Offset && (Count == dynamic_extent || Count <= Extent - Offset),
+        static_assert(Extent == dynamic_extent || (Extent >= Offset && (Count == dynamic_extent ||
+                                                                        Count <= Extent - Offset)),
                       "subspan() cannot extract more elements from a span than it contains.");
         Expects((size() >= Offset) && (Count == dynamic_extent || (Count <= size() - Offset)));
         using type =
--- a/include/gsl/span
+++ b/include/gsl/span
@@ -582,6 +582,8 @@ public:
     template <std::size_t Count>
     constexpr span<element_type, Count> first() const noexcept
     {
+        static_assert(Count <= Extent,
+                      "first() cannot extract more elements from a span than it contains.");
         Expects(Count <= size());
         return span<element_type, Count>{data(), Count};
     }
@@ -592,6 +594,8 @@ public:
         // clang-format on
         constexpr span<element_type, Count> last() const noexcept
     {
+        static_assert(Count <= Extent,
+                      "last() cannot extract more elements from a span than it contains.");
         Expects(Count <= size());
         return span<element_type, Count>{data() + (size() - Count), Count};
     }
@@ -603,6 +607,8 @@ public:
         constexpr auto subspan() const noexcept ->
         typename details::calculate_subspan_type<ElementType, Extent, Offset, Count>::type
     {
+        static_assert(Extent >= Offset && (Count == dynamic_extent || Count <= Extent - Offset),
+                      "subspan() cannot extract more elements from a span than it contains.");
         Expects((size() >= Offset) && (Count == dynamic_extent || (Count <= size() - Offset)));
         using type =
             typename details::calculate_subspan_type<ElementType, Extent, Offset, Count>::type;
--- a/tests/notnull_tests.cpp
+++ b/tests/notnull_tests.cpp
@@ -30,6 +30,13 @@
 #include "deathTestCommon.h"
 using namespace gsl;
 
+#if __cplusplus >= 201703l
+using std::void_t;
+#else  // __cplusplus >= 201703l
+template <class...>
+using void_t = void;
+#endif // __cplusplus < 201703l
+
 struct MyBase
 {
 };
@@ -142,9 +149,38 @@ bool helper_const(not_null<const int*> p) { return *p == 12; }
 int* return_pointer() { return nullptr; }
 } // namespace
 
+template <typename U, typename = void>
+static constexpr bool CtorCompilesFor_A = false;
+template <typename U>
+static constexpr bool
+    CtorCompilesFor_A<U, void_t<decltype(gsl::not_null<void*>{std::declval<U>()})>> = true;
+
+template <typename U, int N, typename = void>
+static constexpr bool CtorCompilesFor_B = false;
+template <typename U, int N>
+static constexpr bool CtorCompilesFor_B<U, N, void_t<decltype(gsl::not_null<U>{N})>> = true;
+
+template <typename U, typename = void>
+static constexpr bool DefaultCtorCompilesFor = false;
+template <typename U>
+static constexpr bool DefaultCtorCompilesFor<U, void_t<decltype(gsl::not_null<U>{})>> = true;
+
+template <typename U, typename = void>
+static constexpr bool CtorCompilesFor_C = false;
+template <typename U>
+static constexpr bool
+    CtorCompilesFor_C<U, void_t<decltype(gsl::not_null<U*>{std::declval<std::unique_ptr<U>>()})>> =
+        true;
+
 TEST(notnull_tests, TestNotNullConstructors)
 {
     {
+        static_assert(CtorCompilesFor_A<void*>, "CtorCompilesFor_A<void*>");
+        static_assert(!CtorCompilesFor_A<std::nullptr_t>, "!CtorCompilesFor_A<std::nullptr_t>");
+        static_assert(!CtorCompilesFor_B<void*, 0>, "!CtorCompilesFor_B<void*, 0>");
+        static_assert(!DefaultCtorCompilesFor<void*>, "!DefaultCtorCompilesFor<void*>");
+        static_assert(!CtorCompilesFor_C<int>, "CtorCompilesFor_C<int>");
+
 #ifdef CONFIRM_COMPILATION_ERRORS
         // Forbid non-nullptr assignable types
         not_null<std::vector<int>> f(std::vector<int>{1});
@@ -271,6 +307,27 @@ TEST(notnull_tests, TestNotNullostream)
     ostream_helper<std::string>("string");
 }
 
+template <typename U, typename V, typename = void>
+static constexpr bool AssignmentCompilesFor = false;
+template <typename U, typename V>
+static constexpr bool
+    AssignmentCompilesFor<U, V,
+                          void_t<decltype(std::declval<gsl::not_null<U*>&>().operator=(
+                              std::declval<gsl::not_null<V*>&>()))>> = true;
+
+template <typename U, typename V, typename = void>
+static constexpr bool SCastCompilesFor = false;
+template <typename U, typename V>
+static constexpr bool
+    SCastCompilesFor<U, V, void_t<decltype(static_cast<U*>(std::declval<gsl::not_null<V*>&>()))>> =
+        true;
+
+template <typename U, typename V, typename = void>
+static constexpr bool RCastCompilesFor = false;
+template <typename U, typename V>
+static constexpr bool RCastCompilesFor<
+    U, V, void_t<decltype(reinterpret_cast<U*>(std::declval<gsl::not_null<V*>&>()))>> = true;
+
 TEST(notnull_tests, TestNotNullCasting)
 {
     MyBase base;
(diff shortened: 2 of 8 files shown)


beinhaerter:
I was not able to replace all checks in the touched files.

`decltype(gsl::to_byte(char{})) x;` compiles, it seems to ignore the `static_assert(std::is_same<T, unsigned char>::value, ...);`. This means I can't use this expression in SFINAE context.

`not_null<int`, even in SFINAE context triggers the class wird `static_assert(details::is_comparable_to_nullptr<T>::value, "T cannot be compared to nullptr.");`.

`algorithm_tests.cpp` and `span_test.cpp` will come later.

tests/span_tests.cpp
```diff
@@ -577,73 +447,38 @@ TEST(span_test, from_container_constructor)
...
         span<const char> cs{cstr};
         EXPECT_TRUE(cs.size() == cstr.size());
         EXPECT_TRUE(cs.data() == cstr.data());
     }
 
-    {
-#ifdef CONFIRM_COMPILATION_ERRORS
-        auto get_temp_vector = []() -> std::vector<int> { return {}; };
-        auto use_span = [](span<int> s) { static_cast<void>(s); };
-        use_span(get_temp_vector());
```
beinhaerter, line 603: Only this line fails to compile. The check could have been made easier with a `std::move`. In the new tests I just pass an rvalue ref to the ctor.

tests/span_tests.cpp
```diff
@@ -577,73 +447,38 @@ TEST(span_test, from_container_constructor)
...
         auto get_temp_string = []() -> std::string { return {}; };
         auto use_span = [](span<const char> s) { static_cast<void>(s); };
         use_span(get_temp_string());
     }
 
-    {
-#ifdef CONFIRM_COMPILATION_ERRORS
-        auto get_temp_vector = []() -> const std::vector<int> { return {}; };
-        auto use_span = [](span<const char> s) { static_cast<void>(s); };
-        use_span(get_temp_vector());
```
beinhaerter, line 631: This fails to compile, but because of mismatched int <-> const char. I guess `const std::string` instead of `const std::vector<int>` was intended.

tests/span_tests.cpp
```diff
@@ -298,70 +297,14 @@ TEST(span_test, from_array_constructor)
...
 
-#ifdef CONFIRM_COMPILATION_ERRORS
-    {
-        span<int, 6> s{arr};
-    }
-
-    {
-        span<int, 0> s{arr};
-        EXPECT_TRUE(s.size() == 0);
-        EXPECT_TRUE(s.data() == &arr[0]);
```
beinhaerter, line 309: These checks don't make sense when the variable isn't even there because of a compilation error.

tests/span_tests.cpp
```diff
@@ -577,73 +447,38 @@ TEST(span_test, from_container_constructor)
     const std::string cstr = "hello";
 
     {
-#ifdef CONFIRM_COMPILATION_ERRORS
```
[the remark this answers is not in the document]
carsonRadtke: What language mode are you using? `basic_string<CharT>::data()` exclusively returned `const charT*` until C++17. I would expect this to fail in C++14 mode and pass in later modes.

tests/span_tests.cpp
carsonRadtke: I wonder if the original author of `CONFIRM_COMPILATION_ERRORS` meant "This doesn't compile, confirm whether that is correct." rather than "This should not compile". Unfortunately, we have no way of determining. Maybe comment it out and see if any older compilers (or C++14) protest?

That being said - Line 303 shouldn't compile because the static extent exceeds the backing array. All the other look fine, I think.

tests/span_tests.cpp
```diff
@@ -577,73 +447,40 @@ TEST(span_test, from_container_constructor)
...
+#ifdef FAIL_ON_SOME_PLATFORMS
```
beinhaerter, line 450: I introduced this (not set macro) check because these checks sometimes succeed and sometimes fail.

tests/span_tests.cpp
beinhaerter: What exactly do you mean with
> Maybe comment it out and see if any older compilers (or C++14) protest?

`#ifdef CONFIRM_COMPILATION_ERRORS` => `#if 1`?

All variable declarations in this `CONFIRM_COMPILATION_ERRORS` block fail. With the new `CtorCompilesFor` checks we are sure that the variable definitions are invalid on all platforms. So from my understanding on no single platform the `EXPECT_TRUE(...)` test statements can compile.

tests/span_tests.cpp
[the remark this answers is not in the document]
beinhaerter: That was the reason -> fixed.

tests/span_tests.cpp
beinhaerter: Correction: `U{std::move(...)}` would still be the explicit constructor. This here checks implicit conversions. I will add additional tests for them.

beinhaerter:
If desired, I can also move the `static_assert`s into the functions where the old `#ifdef CONFIRM-COMPILATION_ERRORS` were. That might be better for getting a good overview and structuring.

tests/span_tests.cpp
```diff
@@ -589,10 +589,7 @@ TEST(span_test, first)
 
     {
         span<int, 5> av = arr;
-#ifdef FAIL_ON_SOME_PLATFORMS
-        EXPECT_TRUE(av.first<6>().size() == 6);
-        EXPECT_TRUE(av.first<-1>().size() == -1);
-#endif
+        EXPECT_DEATH(av.first<6>().size(), expected);
```
beinhaerter: This looks like a regression. I would expect this to be checked at compile time, not at runtime.

tests/span_tests.cpp
carsonRadtke: > What exactly do you mean with <...>

Yeah, `#if 1` is in spirit of what I meant - I just wanted to enable the body to see what fails.

> So from my understanding on no single platform the EXPECT_TRUE(...) test statements can compile.

Do any of these examples compile when using `std::span` instead of `gsl::span`? If so, I'd like to understand why. Otherwise, let's delete the tests.

beinhaerter: I just tested in a small C++20 project: none of the variable declarations compiles.
So as of right now we do check the declarations don't compile, but we remove the `EXPECT_TRUE()` that could never have been effective.
```
#include <span>

int main() {
    int arr[5] = { 1, 2, 3, 4, 5 };
    {
        std::span<int, 6> s{ arr };
    }

    {
        std::span<int, 0> s{ arr };
    }

    int arr2d[2][3] = { 1, 2, 3, 4, 5, 6 };
    {
        std::span<int> s{ arr2d };
    }

    {
        std::span<int, 0> s{ arr2d };
    }

    {
        std::span<int, 6> s{ arr2d };
    }
    return 0;
}
```

tests/span_tests.cpp
carsonRadtke: I agree. `span<T, Extent>::first<N>()` should ensure that `N` and `Extent` play nicely together.

carsonRadtke: That being said - it isn't a PR blocker. If you want to fix it, go ahead. If not, I'm sure another PR will take care of it.

tests/span_tests.cpp
beinhaerter: OK, I changed it. It looks like VS16 and VS17 fail these tests in C++14 mode.

My preprocessor check is `#if !defined(_MSC_VER) || (_MSC_VER > 1942) || (__cplusplus >= 201703L)`. The check `(_MSC_VER > 1942)` is intended to activate the `static_assert` for the next VS release, so that we automatically keep an eye on it when a newer VS version is used.

tests/span_tests.cpp
beinhaerter: I prefer to only update the tests in this PR. Let's do the bugfix in a separate PR.

beinhaerter:
> If desired, I can also move the `static_assert`s into the functions where the old `#ifdef CONFIRM-COMPILATION_ERRORS` were. That might be better for getting a good overview and structuring.

Done.

tests/notnull_tests.cpp
```diff
@@ -515,12 +591,15 @@ TEST(notnull_tests, TestNotNullConstructorTypeDeduction)
...
 
-#ifdef CONFIRM_COMPILATION_ERRORS
-    {
-        not_null x{nullptr};
-        helper(not_null{nullptr});
-        helper_const(not_null{nullptr});
-    }
+    static_assert(TypeDeductionCtorCompilesFor<void*>, "TypeDeductionCtorCompilesFor<void*>");
+#if defined(_MSC_VER)
+    // Fails on gcc, clang, xcode, VS clang with
```
beinhaerter, line 596: The pipeline stops at the first error. I did not check if this compiles for any of the gcc/clang/xcode versions.

tests/span_tests.cpp
```diff
@@ -216,6 +222,12 @@ TEST(span_test, from_pointer_length_constructor)
 
 TEST(span_test, from_pointer_pointer_construction)
```
beinhaerter, line 223: This function has three commented out checks with comments "// this will fail the std::distance() precondition, which asserts on MSVC debug builds". I tried to activate the tests on different platforms, but then the tests failed. I could not match the comment with the death behaviour of the code.

tests/span_tests.cpp
beinhaerter: Our assumption is right. https://godbolt.org/z/3TMsh7sjs shows that gcc, clang and vs all agree that this fails to compile for a `std::span`.

beinhaerter: I changed my mind and fixed it in this PR.

tests/span_tests.cpp
carsonRadtke: > I could not match the comment with the death behaviour of the code.

How does it fail today? I don't see a `std::distance` call in `gsl/span`, so I wonder if the comment is just out of date.

Probably also worth noting that the `&arr[1]` expression is UB so if compilers treat it differently, it may not be a compiler bug.

tests/span_tests.cpp
carsonRadtke: > I guess const std::string instead of const std::vector<int> was intended.

It was introduced by https://github.com/microsoft/GSL/pull/215, but I can't figure out what the intent was. Either delete it or change to `const std::string`; up to you.

tests/span_tests.cpp
```diff
@@ -800,10 +702,9 @@ TEST(span_test, first)
     {
         span<int, 5> av = arr;
 #ifdef CONFIRM_COMPILATION_ERRORS
-        EXPECT_TRUE(av.first<6>().size() == 6);
-        EXPECT_TRUE(av.first<-1>().size() == -1);
```
carsonRadtke, line 804: Why did `first<-1>` disappear?

carsonRadtke (changes requested):
Changing macro'd out code for `static_assert`s is a great change, thanks!

Just a couple comments/questions/suggestions. Once those are resolved, it will be ready to merge.

tests/span_tests.cpp
beinhaerter: -1 is an `int`, the template parameter is a `std::size_t`. clang unit tests fail because of a warning for this narrowing conversion. I could have written `av.first<static_cast<std::size_t>(-1)>()`, but I thought that this does not add to the test coverage. If you want that test, I can reintroduce it.

carsonRadtke: Ah! I missed that. I agree with you removing it.

include/gsl/span
```diff
@@ -582,6 +582,8 @@ public:
     template <std::size_t Count>
     constexpr span<element_type, Count> first() const noexcept
     {
+        static_assert(Count <= Extent,
```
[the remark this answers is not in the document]
beinhaerter: It is added now.

I just copied the checks from `std::span`, that the compiler outputs as a warning, and I thought that dynamic_extent always is the maximum possible value, so we don't need the check. But I missed that one line above the `static_assert`s in `std::span` there are `if constexpr (Extent != dynamic_extent)`.

carsonRadtke:
I opened https://github.com/microsoft/GSL/pull/1181 to address the failing checks.

I will merge ASAP.

tests/span_tests.cpp
beinhaerter: I know that two of the three `EXPECT_DEATH` failed because no death occured. The pipeline said line 296 and line 273 failed.

Because of ammending commits I cannot match failed pipelines to still existing commits. But I think these were the test 2 and 3, not the first one.

tests/span_tests.cpp
beinhaerter: I kept both: the original and the probably intended version.
```
    static_assert(!ConversionCompilesFor<span<const char>, const std::vector<int>>,
                  "!ConversionCompilesFor<span<const char>, const std::vector<int>>");
    static_assert(!ConversionCompilesFor<span<char>, const std::string>,
                  "!ConversionCompilesFor<span<char>, const std::string>");
```

tests/span_tests.cpp
beinhaerter: For `std::span` all three tests are UB.

This is the implementation of the `gsl::span` ctor.
```
    template <std::size_t MyExtent = Extent, std::enable_if_t<MyExtent != dynamic_extent, int> = 0>
    constexpr explicit span(pointer firstElem, pointer lastElem) noexcept
        : storage_(firstElem, narrow_cast<std::size_t>(lastElem - firstElem))
    {
        Expects(lastElem - firstElem == static_cast<difference_type>(Extent));
    }
```

I think for GSL the relevant point is `lastElem - firstElem`. For the first case (&arr[0] - &arr[1]) this is OK. The length store in `storage_` is not fine, but the `Expects` triggers.
For the checks with a nullptr the pointer subtraction should be a problem (the pointers do not point into the same object -> UB).

If I don't miss anything then test one is a test that we can activate, but tests two and three should be deleted or left commented out.

carsonRadtke: I think all tests should be deleted. It will be confusing to have tests that may start to fail with a compiler update.

If you see value in keeping them commented out for reference, that's fine. But the false comments about `std::distance` should be removed.

beinhaerter: > Tests 2 and 3 are identical? Let me know if I'm missing something. Test 1 is [definitely UB](https://godbolt.org/z/b7TGq88q6).

Your godbolt test is range nullptr[1]..nullptr[0]. Test 1 is arr[1]..arr[0] where arr is a C style array.

I delete the tests 2 and 3. For test 1 you can decide. I think we do not need to assure this behaviour, but if we want to, it does not hurt.

carsonRadtke:
Looks good! Thanks for the PR and thanks for making all the requested changes :)

PR #1184, opened 2025-01-07
Add "gsl" to #includes

tiagomacarios:
Office is seeing build breaks due to `#include "span"` including C++20 span instead of gsl/span. Most likely we want all headers includes qualified with "gsl/" to avoid similar issues.

See https://office.visualstudio.com/Office/_git/Office/pullrequest/3657798

### DIFF
--- a/include/gsl/algorithm
+++ b/include/gsl/algorithm
@@ -17,8 +17,8 @@
 #ifndef GSL_ALGORITHM_H
 #define GSL_ALGORITHM_H
 
-#include "assert" // for Expects
-#include "span"   // for dynamic_extent, span
+#include "gsl/assert" // for Expects
+#include "gsl/span"   // for dynamic_extent, span
 
 #include <algorithm>   // for copy_n
 #include <cstddef>     // for ptrdiff_t
--- a/include/gsl/gsl
+++ b/include/gsl/gsl
@@ -18,16 +18,16 @@
 #define GSL_GSL_H
 
 // IWYU pragma: begin_exports
-#include "algorithm"   // copy
-#include "assert"      // Ensures/Expects
-#include "byte"        // byte
-#include "pointers"    // owner, not_null
-#include "span"        // span
-#include "zstring"     // zstring
-#include "util"        // finally()/narrow_cast()...
+#include "gsl/algorithm"   // copy
+#include "gsl/assert"      // Ensures/Expects
+#include "gsl/byte"        // byte
+#include "gsl/pointers"    // owner, not_null
+#include "gsl/span"        // span
+#include "gsl/zstring"     // zstring
+#include "gsl/util"        // finally()/narrow_cast()...
 
 #ifdef __cpp_exceptions
-#include "narrow" // narrow()
+#include "gsl/narrow" // narrow()
 #endif
 // IWYU pragma: end_exports
 
--- a/include/gsl/narrow
+++ b/include/gsl/narrow
@@ -16,8 +16,8 @@
 
 #ifndef GSL_NARROW_H
 #define GSL_NARROW_H
-#include "assert"    // for GSL_SUPPRESS
-#include "util"      // for narrow_cast
+#include "gsl/assert"    // for GSL_SUPPRESS
+#include "gsl/util"      // for narrow_cast
 #include <exception> // for std::exception
 namespace gsl
 {
--- a/include/gsl/pointers
+++ b/include/gsl/pointers
@@ -17,7 +17,7 @@
 #ifndef GSL_POINTERS_H
 #define GSL_POINTERS_H
 
-#include "assert" // for Ensures, Expects
+#include "gsl/assert" // for Ensures, Expects
 
 #include <cstddef>      // for ptrdiff_t, nullptr_t, size_t
 #include <functional>   // for less, greater
--- a/include/gsl/span
+++ b/include/gsl/span
@@ -17,10 +17,10 @@
 #ifndef GSL_SPAN_H
 #define GSL_SPAN_H
 
-#include "assert"   // for Expects
-#include "byte"     // for byte
-#include "span_ext" // for span specialization of gsl::at and other span-related extensions
-#include "util"     // for narrow_cast
+#include "gsl/assert"   // for Expects
+#include "gsl/byte"     // for byte
+#include "gsl/span_ext" // for span specialization of gsl::at and other span-related extensions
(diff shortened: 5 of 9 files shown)


include/gsl/algorithm
```diff
@@ -17,8 +17,8 @@
 #ifndef GSL_ALGORITHM_H
 #define GSL_ALGORITHM_H
 
-#include "assert" // for Expects
-#include "span"   // for dynamic_extent, span
+#include "gsl/assert" // for Expects
```
beinhaerter, line 20: I know, include search order is implementation defined. But from what I see from current compilers:

Includes with "" are searched in the current path first. In the current path here there is no subfolder "gsl/".

If at all then it should be <gsl/...>, but <> is for system/external includes, not for include from the same project.

In my opinion this is a change that makes things worse, not better. I bet it's a problem with the Office build setttings and it should be fixed there, not here.

beinhaerter: https://learn.microsoft.com/en-us/cpp/build/reference/i-additional-include-directories?view=msvc-170

> If the #include directive is specified using double-quote form, it first searches local directories. The search begins in the same directory as the file that contains the #include directive. If it fails to find the file, it searches next in the directories of the currently opened include files, in the reverse order in which they were opened. The search begins in the directory of the parent include file and continues upward through the directories of any grandparent include files.

So the Office code needs to `#include <gsl/...>`, but the GSL itself can `#include "..."`.

tiagomacarios: What we were observing is that gsl/algorithm@21 and gsl/gsl@25 where including the C++20 std::span headers. Our code does use the fully qualified include.

carsonRadtke: I am inclined to revert this PR. I fear we may have blocked customers from installing `gsl` under a different name (for example, they might `#include <ms-gsl/gsl>`). 

> What we were observing is that gsl/algorithm@21 and gsl/gsl@25 where including the C++20 std::span headers. Our code does use the fully qualified include.

I don't want reverting this PR to mean that Office needs to apply a patch every time y'all consume GSL. @tiagomacarios Can you link to a pipeline where the wrong `span` was included? Maybe there are some clues there that will help us understand why.

tiagomacarios: @beinhaerter Office uses header units. When header units are enabled the compiler prefers a header map (not in the docs you linked). It is unclear if this is by design or not. We are following up with the compiler team.
Anyway, path should be fully qualified. See boost or any other large opensource project for examples.

beinhaerter: > Anyway, path should be fully qualified. See boost or any other large opensource project for examples.

First of all Boost and Qt as an example don't use `#include "folder/file"`, they use `#include <folder/file>`.

Second, why should the pathes be fully qualified? Can you please give a reason?

CppCoreGuidelines say [SF.12: Prefer the quoted form of #include for files relative to the including file and the angle bracket form everywhere else](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf12-prefer-the-quoted-form-of-include-for-files-relative-to-the-including-file-and-the-angle-bracket-form-everywhere-else). I think this best practice perfectly matches with the search order described in the MSVC docs that I linked.

Does Office build when you are not using header units? (To check if it is a GSL problem or a compiler problem.)

carsonRadtke: > I am inclined to revert this PR.

I take this back; the PR won't be reverted.

> I fear we may have blocked customers from installing gsl under a different name

Supporting installing GSL under a different name is not a goal.

> If GSL itself does #include "gsl/..." then this fails and what is found then depends on the parent include files and if they have a gsl subfolder.

As per the rules for `#include`:
"""
If it fails to find the file, it searches next in the directories of the currently opened include files, in the reverse order in which they were opened.
"""
The key is the last sentence - it will look for "gsl/gsl/..." and then since it won't find it, it will look for "gsl/../gsl/..." and will then find it.

> Does Office build when you are not using header units?

Yes, but that does not indicate the compiler is at fault (see my response to the next point). 

> Second, why should the pathes be fully qualified? Can you please give a reason?

A couple things are going on with Office that require this. These are not misconfigurations:
1. Office is using [`/translateInclude`](https://learn.microsoft.com/en-us/cpp/build/reference/translateinclude?view=msvc-170). This treats any `#include "abc"` as `import "abc"`. 
2. Office is using  [`/headerUnit:angle span=<...>`](https://learn.microsoft.com/en-us/cpp/build/reference/headerunit?view=msvc-170) which will lookup `import "span"` using the same rules `#include <span>`.

Ultimately, this means that any header that shares a name with a standard library header needs to either be fully-qualified or start with a "./". Otherwise (under `/translateInclude /headerUnit:angle <...>`) the standard library header will get picked up.

tiagomacarios: @carsonRadtke thoughts on replacing quotes with angle brackets? I presume this might make @beinhaerter more comfortable with the change.

beinhaerter: > The key is the last sentence - it will look for "gsl/gsl/..." and then since it won't find it, it will look for "gsl/../gsl/..." and will then find it.

My reading of the docs is different. The docs don't say that the parent directory of the current include is searched. It says that the directory of the parent include is searched. Given
- PROJ-path/a.cpp including <gsl/algorithm>
- GSL-path/gsl/algorithm including "span"
- GSL-path/gsl/span including "gsl/assert"

gsl/assert will be searched as:
1. GSL-path/gsl/gsl/assert (relative to gsl/span)
2. GSL-path/gsl/gsl/assert (relative to gsl/span)
3. PROJ-path/gsl/assert (relative to a.cpp) => this can cause problems!
4. GSL-path/gsl/assert (because of INCLUDE environment variable)

> Office is using [/headerUnit:angle span=<...>](https://learn.microsoft.com/en-us/cpp/build/reference/headerunit?view=msvc-170) which will lookup import "span" using the same rules #include <span>.

So shouldn't this work?
```
/headerUnit:angle span=<...>
/headerUnit:angle gsl/span=<...>
import <span>;
import <gsl/span>;
```

Do you have a minimal compilable example where the wrong behaviour can be observed? Otherwise I will try to reproduce it myself.

carsonRadtke: > If you want to wait with reverting the change until the Office team is ready, then an #include "./span" looks appealing because it follows the guidelines (even if the "./" looks a bit strange) and it works for the Office team.

I will create a PR that reverts this PR and takes this approach. @tiagomacarios Any objections?

Issue #1186, opened 2025-01-27
Add element_type to not_null


daltairwalter:
element_type is a standard concept in smart points and can be useful for specializations.  It would be nice if not_null supported this concept too.

carsonRadtke:
Thanks for noticing this! It sounds like a helpful extension to `gsl::not_null`.

PR #1196, opened 2025-02-11
introduce gsl::not_null<T>::element_type

carsonRadtke:
fixes: https://github.com/microsoft/GSL/issues/1186

### DIFF
--- a/tests/pointers_tests.cpp
+++ b/tests/pointers_tests.cpp
@@ -90,7 +90,7 @@ TEST(pointers_test, swap)
 
 TEST(pointers_test, member_types)
 {
-    static_assert(std::is_same<gsl::not_null<int>::element_type, int>::value,
+    static_assert(std::is_same<gsl::not_null<int*>::element_type, int*>::value,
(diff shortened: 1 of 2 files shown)


beinhaerter:
The commit message contains
"* use std::is_same instead of is_same_v
* fix: cannot put a non-pointer in a gsl::not_null"

Both is nothing that this merge request changed.

docs/headers.md
```diff
@@ -224,6 +224,14 @@ When a nullptr check fails, `std::terminate` is called.
 
 See [F.23: Use a `not_null<T>` to indicate that “null” is not a valid value](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-nullptr)
 
+#### Member Types
+
+```cpp
+using element_type = T;
+```
+
+The type of the pointed-to object.
```
beinhaerter, line 233: This is wrong. T is a raw pointer or a smart pointer, but not the type of the pointed to object.

PR #1198, opened 2025-02-14
deprecate features adopted into C++

carsonRadtke:
1) Mark the following GSL features as deprecated:
 - gsl::unique_ptr (always)
 - gsl::shared_ptr (always)
 - gsl::byte (since c++17)
 - gsl::joining_thread (never implemented)
2) Refactor existing deprecations to use the new GSL_DEPRECATED(msg) macro.
3) Create a section in the README for deprecated features in the
   standard.

### DIFF
--- a/include/gsl/byte
+++ b/include/gsl/byte
@@ -17,6 +17,8 @@
 #ifndef GSL_BYTE_H
 #define GSL_BYTE_H
 
+#include "./util" // for GSL_DEPRECATED
+
 #include <type_traits>
 
 #ifdef _MSC_VER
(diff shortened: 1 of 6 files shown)


include/gsl/byte
```diff
@@ -72,6 +74,12 @@
 #define byte_may_alias
 #endif // defined __clang__ || defined __GNUC__
 
+#if GSL_USE_STD_BYTE
+#define BYTE_TYPE std::byte
+#else // !GSL_USE_STD_BYTE
+#define BYTE_TYPE byte
+#endif // GSL_USE_STD_BYTE
+
```
beinhaerter, line 82: Why is that macro introduced? It was not needed before. Did some code not compile without this macro?

And why is this macro exported?

A macro with a generic name like `BYTE_TYPE` looks like it has potential to break existing code.

include/gsl/span
```diff
@@ -18,7 +18,7 @@
 #define GSL_SPAN_H
 
 #include "./assert"   // for Expects
-#include "./byte"     // for byte
+#include "./byte"     // for BYTE_TYPE
```
beinhaerter, line 21: It worries me if that change is necessary. That would mean that users using GSL might also need to use `BYTE_TYPE` instead of `gsl::byte`, which in my opinion is a change for the worse.

PR #1202, opened 2025-03-26
Fix version info

beinhaerter:
Just tagging as 4.2.0 is not enough, the version number must be incremented for cmake and in the readme. Compare https://github.com/microsoft/GSL/pull/1163.


carsonRadtke:
Thanks for catching this! I'll find a way to automate this, so it isn't a problem (and we don't have to remember this extra step).

Issue #1204, opened 2025-04-25
`gsl::not_null` should allow returning by value for the types whose size equals to the size of two pointers


ozaktash:
In `gsl::not_null`, objects of types smaller than two pointers can be returned by value, but when the size equals to the size of two pointers, returning by reference is forced:
```cpp
    template<typename T>
    using value_or_reference_return_t = std::conditional_t<
                                            sizeof(T) < 2*sizeof(void*) && std::is_trivially_copy_constructible<T>::value,
                                            const T,
                                            const T&>;
```

I think returning by value should also be allowed for types whose size equals to the size of two pointers:
```cpp
    template<typename T>
    using value_or_reference_return_t = std::conditional_t<
                                            sizeof(T) <= 2*sizeof(void*) && std::is_trivially_copy_constructible<T>::value,
                                            const T,
                                            const T&>;
```

This is the threshold used in the current [cppfront](https://github.com/hsutter/cppfront/blob/b2d40209e694f74458ffaf15d8b884300fc405cc/include/cpp2util.h#L1445) and is also compliant with [C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-in).

carsonRadtke:
Thanks for reporting this issue! I agree, it looks like this should be `<=`. From the guidelines:

> (Simple) ((Foundation)) Warn when a parameter passed by reference to const has a size less or equal than 2 * sizeof(void*). Suggest passing by value instead.

@ozaktash, do you want to submit a PR?

PR #1205, opened 2025-05-04
Update return by value threshold in gsl::not_null

ozaktash:
Fixes: #1204 
- Allow returning by value for types that are not greater than two pointers in size


carsonRadtke:
Thanks for identifying this issue and posting a fix!

PR #1218, opened 2025-09-28
Guard includes of standard library headers so users can use import std

### LINKED ISSUE #1217: Allow users to include GSL headers and use import of the standard library module in their code
This proposal is to allow users of GSL to migrate their code to use `import std;` which at the moment is very hard, almost impossible, without using a lot of tricks. I believe there is a growing desire to port code to using the standard library module. However at present projects that rely on third party libraries find this impossible.

About this proposal

1. Any existing users should see no change, they have to opt in by defining GSL_BUILD_USING_STD_MODULE or equivalent macro
2. It does not assume anything about how consumers are building their code, it does not require them to use CMake or a package manager. It does not propose allowing the GSL project, including tests, itself to be build using `import std;`, I would be happy to do this work if people wanted
3. It is not a proposal to create a GSL module.

Currently the header files of GSL include headers from the standard library, for instance
```
#include <vector>
```
If a user I want to migrate my code to use `import std;` they can get compiler failures or at the very least a polluted global namespace when I also include the current GSL headers.

In effect we end up with the following
```
    import std; // User code
    #include <vector> // Via #include of a GSL header
```
I believe that doesn't compile on any compiler as of today and have been told it is very difficult to do for "reasons". So how about
```
    #include <vector> // Via #include of a GSL header
    import std; // User code
```
Well that does compile, mostly, but it throws away the main reason for modules, which is interface encapsulation. So on MSVC the following will compile, similar stuff happens for other compilers and standard libraries
```
    #include <vector>
    import std;
    std::_Iterator_base x;  // Magic internal symbol that leaks
```
So all the internal symbols and macros of the standard library will still pollute any user. You are also paying the compile costs of both the `#include` and the `import std;`. In general from experience in a more complex code base using multiple third party libraries it is extremely difficult to avoid arbitrary ordering of `#include` and `import std;` Even if compilers do manage to allow arbitrary mixing of `#include` and `import` doing so, I believe will still be fragile and poor practice.
Note there are a few cases where you do need to include a header, so
```
#include <version>   // For feature test macros
#include <stddef.h> // For offsetof
import std;
```
Generally this is most to access things only available as macros or C functions not in the standard library. For more info see https://wg21.link/p2654

So the proposal in the issue is to macro guard all #include of the standard library in the GSL public headers. For instance
```
#if defined(GSL_BUILD_USING_STD_MODULE)
    import std;
#else
    #include <string>
#endif
```
Note the macro name can be anything I just picked this as an example.
Also note there is an `import std;`. This is to make sure the headers can stand alone. It's not actually required as the users could always do a `import std;` before the `#include` of the GSL header as all exported standard library names would then be visible, with the exception of macros. This would look like
```
#if !defined(GSL_DONT_INCLUDE_STANDARD_LIBRARY_HEADERS)
    #include <string>
#endif
```

davidhunter22:
fixes #1217 

This is an example of how issue #1217 could be fixed

### DIFF
--- a/include/gsl/algorithm
+++ b/include/gsl/algorithm
@@ -20,9 +20,13 @@
 #include "./assert" // for Expects
 #include "./span"   // for dynamic_extent, span
 
-#include <algorithm>   // for copy_n
-#include <cstddef>     // for ptrdiff_t
-#include <type_traits> // for is_assignable
+#if defined(GSL_BUILD_USING_STD_MODULE)
+    import std;
+#else
+    #include <algorithm>   // for copy_n
+    #include <cstddef>     // for ptrdiff_t
+    #include <type_traits> // for is_assignable
+#endif
 
 #ifdef _MSC_VER
 #pragma warning(push)
--- a/include/gsl/assert
+++ b/include/gsl/assert
@@ -36,7 +36,11 @@
 #else // defined(_MSC_VER) && (defined(_KERNEL_MODE) || (defined(_HAS_EXCEPTIONS) &&
       // !_HAS_EXCEPTIONS))
 
-#include <exception>
+#if defined(GSL_BUILD_USING_STD_MODULE)
+    import std;
+#else
+    #include <exception>
+#endif
 
 #endif // defined(_MSC_VER) && (defined(_KERNEL_MODE) || (defined(_HAS_EXCEPTIONS) &&
        // !_HAS_EXCEPTIONS))
--- a/include/gsl/byte
+++ b/include/gsl/byte
@@ -19,7 +19,11 @@
 
 #include "./util" // for GSL_DEPRECATED
 
-#include <type_traits>
+#if defined(GSL_BUILD_USING_STD_MODULE)
+    import std;
+#else
+    #include <type_traits>
+#endif
 
 #ifdef _MSC_VER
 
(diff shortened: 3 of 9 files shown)


carsonRadtke:
Thanks for the change! Adding modules to a project is new to me, so it will take a little bit of time for me to learn the things necessary to properly review this PR. 

In the meantime, can you add a cmake variable that controls `GSL_BUILD_USING_STD_MODULE` as well as a test case to verify your change behaves as expected?

davidhunter22:
I will add  can you add a cmake variable that controls GSL_BUILD_USING_STD_MODULE and a test, I was just very wary about changing CMake stuff.

I can do this in two ways.
1. Add a CMake option. Add a single project using this option that builds using `import std;` and leave the rest alone. So a minimal set of changes.
2. Add a CMake option. Modify all existing test so the work with `import std;'. This is I think all macro guarding `#include` of standard library headers.

Let me know which you would prefer.
I would suggest I submit this as a separate PR unless you want all the changes together.

davidhunter22:
I did notice one issue with using GoogleTest. On MSVC it compiles with Unicode defined and with the "/J" and "UTF-8" flags set. The GSL tests compile with MultiByteCharacterSet and do not set "/J" and "UTF-8". I'm not sure this is compatible. It certainly complains about it when I use `import std;` which is more picky about these sorts of things.

davidhunter22:
On the "/J" and "UTF-8" flags I'm not sure I'm qualified to make an opinion.
Is it Ok to mix them, I'm not sure. All I do know is when I tried building using `import std;` if was getting warning about incompatible flags.
Given that GSL is a MicroSoft supported thing do you have an in with the compiler devs to ask them if mixing these flags is OK?

carsonRadtke:
> Given that GSL is a MicroSoft supported thing do you have an in with the compiler devs to ask them if mixing these flags is OK?

Yes, I'm happy to link with compiler team to get clarification on this, I would just like a separate issue tracking the task (preferably with the diagnostic you get when mixing these flags when using modules).

PR #1220, opened 2025-09-29
fix: update conditional static assertion

carsonRadtke:
There is a static assertion that spuriously fails on MSVC that was version checked. Unfortunately it fires every time there is a new compiler update, so for now we will remove the version check and add it back once the compiler bug is fixed.


carsonRadtke:
@copilot, could you take a stab at fixing the pipeline failure?

PR #1222, opened 2025-10-06
cmake: surface headers in Visual Studio solution

### LINKED ISSUE #1216: Make files visible in generated Visual Studio solution
If you generate a Visual studio solution from the current CMake the GSL headers like `algorithm` and `assert` don't appear. The only way you seem to be able to see them is to open a test like `algorithm_tests.cpp` and then do a "Go to document" on a line like `#include <gsl/algorithm>`.
In my local copy I added the following
```
add_custom_target(include SOURCES
    gsl/algorithm
    gsl/assert
    gsl/byte
    gsl/gsl
    gsl/narrow
    gsl/pointers	
    gsl/span
    gsl/span_ext
    gsl/util	
    gsl/zstring		
)
```
But I am sure that is the wrong way to do it.

I also added 
```
set(SolutionItems
    .clang-format
    .gitattributes
    .gitignore
    CMakeSettings.json
    CONTRIBUTING.md
    LICENSE
    README.md
    SECURITY.md
    ThirdPartyNotices.txt)
set_property(DIRECTORY APPEND PROPERTY VS_SOLUTION_ITEMS ${SolutionItems})
```
This makes all these files appears in the "Solution Items" folder in the IDE which is super useful.

cnaples79:
## Summary
- add the public GSL headers to the interface target so Visual Studio solutions list them automatically
- group the headers under a "Header Files" node and surface key repo documents as Solution Items when generating a VS solution

## Rationale
Developers opening the generated Visual Studio solution couldn\'t discover  headers or project metadata without manual include navigation. Adding the files directly to the solution keeps intellisense and browsing consistent with typical VS layout (matching the reporter\'s expectation).

## Changes
- enumerate GSL headers and append them to 'target_sources'
- define Visual Studio source groups and VS_SOLUTION_ITEMS for docs/licenses when using a VS generator
- no functional code changes; header-only interface remains unchanged

Fixes #1216

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,11 +21,9 @@ file(GLOB GSL_HEADER_FILES
     RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}/include"
     "${CMAKE_CURRENT_SOURCE_DIR}/include/gsl/*"
 )
-list(SORT GSL_HEADER_FILES)
 
 # Build/install interface lists ensure CMake is happy when exporting headers via
-# INTERFACE sources. Absolute paths would be rejected by newer CMake versions
-# (discovered while testing with 4.1).
+# INTERFACE sources. Absolute paths would be rejected starting with CMake 4.1.2.
 set(GSL_HEADER_INTERFACE_SOURCES)
 foreach(header ${GSL_HEADER_FILES})
     list(APPEND GSL_HEADER_INTERFACE_SOURCES
@@ -40,14 +38,8 @@ target_sources(GSL INTERFACE
 )
 
 if (CMAKE_GENERATOR MATCHES "Visual Studio")
-    # Visual Studio still needs concrete file paths for solution grouping.
-    set(GSL_HEADER_ABSOLUTE_PATHS)
-    foreach(header ${GSL_HEADER_FILES})
-        list(APPEND GSL_HEADER_ABSOLUTE_PATHS ${GSL_SOURCE_DIR}/include/${header})
-    endforeach()
-    source_group(TREE ${GSL_SOURCE_DIR}/include PREFIX "Header Files" FILES ${GSL_HEADER_ABSOLUTE_PATHS})
-
-    set(GSL_SOLUTION_FILES
+    file(GLOB GSL_WORKSPACE_FILES "${CMAKE_CURRENT_SOURCE_DIR}/include/gsl/*")
+    list(APPEND GSL_WORKSPACE_FILES
         ${GSL_SOURCE_DIR}/.clang-format
         ${GSL_SOURCE_DIR}/.gitattributes
         ${GSL_SOURCE_DIR}/.gitignore
@@ -59,8 +51,14 @@ if (CMAKE_GENERATOR MATCHES "Visual Studio")
         ${GSL_SOURCE_DIR}/ThirdPartyNotices.txt
     )
 
-    set_property(DIRECTORY PROPERTY VS_SOLUTION_ITEMS ${GSL_SOLUTION_FILES})
-    source_group("Solution Items" FILES ${GSL_SOLUTION_FILES})
+    target_sources(GSL PRIVATE ${GSL_WORKSPACE_FILES})
+
+    set(GSL_HEADER_ABSOLUTE_PATHS)
+    foreach(header ${GSL_HEADER_FILES})
+        list(APPEND GSL_HEADER_ABSOLUTE_PATHS ${GSL_SOURCE_DIR}/include/${header})
+    endforeach()
+    source_group(TREE ${GSL_SOURCE_DIR}/include PREFIX "Header Files" FILES ${GSL_HEADER_ABSOLUTE_PATHS})
+    source_group("Solution Items" FILES ${GSL_WORKSPACE_FILES})
 endif()
 
 if (GSL_TEST)


davidhunter22:
Hi I tried cloning you repository and doing an out of tree build using CMake 4.1.2
```
cmake ../src
```
where the `src` directory is a clone of `cnaples79:fix-1216-vs-solution-items` with no changes.
I get errors like 
```
  Target "GSL" INTERFACE_SOURCES property contains path:

    "C:/Users/.../src/include/gsl/algorithm"

  which is prefixed in the source directory.
```
I read a few things on the message and they were talking about adding header using `INTERFACE` or `PUBLIC` like
```
target_sources(GSL INTERFACE
    ${GSL_HEADER_FILES}
    $<BUILD_INTERFACE:${GSL_SOURCE_DIR}/GSL.natvis>
)
```
I tried changing the `INTERFACE` to `PRIVATE` and the CMake ran fine.
I assume I am doing something dumb, I am in now way a CMake expert!

cnaples79:
@davidhunter22 Hey! Yeah it is due to the later versions of CMake past 4.1.x. Try again and let me know if you still get any errors.

CMakeLists.txt
```diff
@@ -17,7 +17,57 @@ target_compile_features(GSL INTERFACE "cxx_std_14")
...
+        $<INSTALL_INTERFACE:include/${header}>
+    )
+endforeach()
+
+target_sources(GSL INTERFACE
+    ${GSL_HEADER_INTERFACE_SOURCES}
+    $<BUILD_INTERFACE:${GSL_SOURCE_DIR}/GSL.natvis>
+)
+
+if (CMAKE_GENERATOR MATCHES "Visual Studio")
```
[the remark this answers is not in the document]
carsonRadtke: > Consider using a more specific generator check.

The check is fine as-is.

davidhunter22:
Looking at the changes I was wondering if there was a simpler way.
I noticed that there is a `CMakeLists.txt` in the include directory.
I tried moving all the stuff to do with the `GSL` target into this `CMakeLists.txt`
This seemed to make most things work  without doing all the path manipulation, including the install step.
Note I only tested on VS so far.

I also wondered if the header file could be added to the `GSL` project as a `FILE_SET`. This also seemed to work well.

CMakeLists.txt
```diff
@@ -17,7 +17,57 @@ target_compile_features(GSL INTERFACE "cxx_std_14")
 # Setup include directory
 add_subdirectory(include)
 
-target_sources(GSL INTERFACE $<BUILD_INTERFACE:${GSL_SOURCE_DIR}/GSL.natvis>)
+set(GSL_HEADER_FILES
```
carsonRadtke, line 20: **suggestion**: Can we `file(GLOB ...)` here instead?

CMakeLists.txt
```diff
@@ -17,7 +17,57 @@ target_compile_features(GSL INTERFACE "cxx_std_14")
...
+    $<BUILD_INTERFACE:${GSL_SOURCE_DIR}/GSL.natvis>
+)
+
+if (CMAKE_GENERATOR MATCHES "Visual Studio")
+    # Visual Studio still needs concrete file paths for solution grouping.
+    set(GSL_HEADER_ABSOLUTE_PATHS)
+    foreach(header ${GSL_HEADER_FILES})
+        list(APPEND GSL_HEADER_ABSOLUTE_PATHS ${GSL_SOURCE_DIR}/include/${header})
+    endforeach()
+    source_group(TREE ${GSL_SOURCE_DIR}/include PREFIX "Header Files" FILES ${GSL_HEADER_ABSOLUTE_PATHS})
```
carsonRadtke, line 55: **issue**: Based on your PR description, it sounds like this should appear in the solution adjacent to "Solution Items"? I'm not seeing anything.

```
> cmake --version
> PS D:\GSL\build> cmake --version
cmake version 4.1.0
```

Image

carsonRadtke (commented):
Thanks for the PR! The pipeline looks good, but can you address these couple comments?

CMakeLists.txt
cnaples79: I added a source_group call, the docs should land in the solution items node now.

carsonRadtke (changes requested):
Unfortunately, this approach requires features from CMake 4.0.0. Right now, we can't force that on our customers, so we will need a different approach. I have outlined my suggestion in one of the comments but would also like to hear of any alternatives that exist.

cnaples79:
@carsonRadtke No worries! I'll do some more research to see if there's a better solution or may end up going with your suggestion. I'll update the PR.

CMakeLists.txt
```diff
@@ -17,7 +17,51 @@ target_compile_features(GSL INTERFACE "cxx_std_14")
...
 
-target_sources(GSL INTERFACE $<BUILD_INTERFACE:${GSL_SOURCE_DIR}/GSL.natvis>)
+file(GLOB GSL_HEADER_FILES
+    RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}/include"
+    "${CMAKE_CURRENT_SOURCE_DIR}/include/gsl/*"
+)
+list(SORT GSL_HEADER_FILES)
+
+# Build/install interface lists ensure CMake is happy when exporting headers via
+# INTERFACE sources. Absolute paths would be rejected by newer CMake versions
```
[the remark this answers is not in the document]
cnaples79: I believe it's 4.1.2, below that didn't throw for me.

davidhunter22:
On using the` VS_SOLUTION_ITEMS` property. This is obviously a magic property name added in  4.0.
On older version it is not magic and I believe does nothing.
Given that this only makes files visible in generated VS solutions, files which have never been visible before and that this only applies to people actually building GSL itself I am wondering what backwards compatibility there is?
I would say this is a new feature that only works if you are using 4.0 of CMaje or above. If you are using earlier than that you are "compatible" in that the behaviour has not changed.

carsonRadtke:
> On using the` VS_SOLUTION_ITEMS` property. This is obviously a magic property name added in 4.0. On older version it is not magic and I believe does nothing. Given that this only makes files visible in generated VS solutions, files which have never been visible before and that this only applies to people actually building GSL itself I am wondering what backwards compatibility there is? I would say this is a new feature that only works if you are using 4.0 of CMaje or above. If you are using earlier than that you are "compatible" in that the behaviour has not changed.

@davidhunter22 -- I see where you're coming from, but this is effectively a feature implementation for GSL and it should work for all customers (in this case, the customers are y'all: the contributors) who meet the dependency requirements. If there was no possible way for this to work pre-4.0.0, I would be more open to a `VS_SOLUTION_ITEMS` solution, but since we can do it in a way that is compatible with 3.* and 4.*, I'd like to see that.

carsonRadtke:
@cnaples79 -- Do you need any assistance with this? It is close, but I'm still waiting on a few comments to be resolved before merging!

PR #1228, opened 2026-01-28
feat: implementation of gsl::dyn_array

### LINKED ISSUE #1169: gsl::dyn_array<T, Deleter>?
Good day folks,

I'm a Microsoft employee working on one of our large internal codebases (happy to discuss more offline :)).  CppCoreGuidelines proposes that there should be a gsl::dyn_array type:

dyn_array<T> // ??? needed ??? A heap-allocated array. The number of elements is determined at construction and fixed thereafter. The elements are mutable unless T is a const type. Basically a span that allocates and owns its elements.
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#gslowner-ownership-pointers

We've been beginning to embrace ```span<T>``` and ```unique_ptr<T, OurCustomDeleter>```. However, we've noticed that there's a missing type for migrating existing code.
Namely, we have:
* ```unique_ptr<T>``` for owned single elements, 
* ```unique_ptr<T[]>``` for owned arrays where we need not track the length
* ```vector<T>``` (and some custom equivalents) for owned, resizeable arrays
* ```span<T>`` for borrowed/unowned arrays with length tracked
* And a gap (which appears that ```dyn_array<T>``` would fill if it existed) for owned, non-resizable arrays.

For us, our operator delete is global (and thus the corresponding deleter is compressible). However, our operator new is stateful / it requires a pointer to the allocator to use (because we do arena allocation and the like). 

So, the gap ```dyn_array<T, Deleter>``` could fill is instead filled with:
* Separate ```unique_ptr<T[]>``` and length arguments (e.g. inout arguments instead of returning a class using NVRO) - clunky and easy to make mistakes
* Separate ```T*``` and length - even worse than the previou soption.
* ```vector<T>``` can't fill the gap either - instead of  ```sizeof(size_t) +sizeof(T*)``` (call it 16 bytes on 64 bit arches), ```vector<T>``` imposes an additional ```sizeof(T*)``` due to the length and capacity split, and usually at least another ```sizeof(T*)``` for the custom allocator.
A ```dyn_array<T, Deleter>``` thus would be very useful to us, as it has no additional stack overhead (ignoring structure vs primitive type argument passing differences in e.g. msvc's ABI, which are unfortunate but probably livable). In other words, it would be a drop-in replacement with the same semantics as having a ```unique_ptr<T[], Deleter>```, but safer.

While our use case is somewhat niche, I suspect we're not the only codebase internally or externally who would find this useful, so I thought I should file an issue and ask if there are plans to add it.

Related issues:
https://github.com/isocpp/CppCoreGuidelines/issues/1633
#348 
#505 
#890

carsonRadtke:
resolves: https://github.com/microsoft/GSL/issues/1169

Implement gsl::dyn_array<T, Allocator> as specified by the CppCoreGuidelines here:
https://github.com/isocpp/CppCoreGuidelines/blob/master/docs/dyn_array.md

Currently the implementation is feature complete and all tests on modern compilers (with supported language versions) are passing.

What needs to be done:

- [x] Create docs
- [ ] Run dyn_array tests under ASAN (nonblocking: nice to have)
- [x] Flesh out some more tests (nonblocking: nice to have, almost out of baseball teams)

What is done (this is a running checklist; I'll delete and rewrite this message before merging):

- [x] Basic Functionality
- [x] Tests of basic functionality
- [x] Tests for constexpr functions
- [x] Tests/Support for ranges
- [x] Ensure support for C++14, C++17, C++20, and C++23
- [x] Tests for custom allocators
- [x] More constexpr tests using a constexpr allocator
- [x] Tests for const iterators
- [x] Sanity-check static_assertions to make sure gsl::dyn_array::iterator is a proper random-access iterator.
- [x] Sanity-check static_assertions to make sure gsl::dyn_array is (mostly) a proper allocator-aware container
- [x] Tests for construction of gsl::dyn_array from std::initializer_list
- [x] Tests/Support for operations on const gsl::dyn_array
- [x] Support for MSVC's unchecked iterators
- [x] Tests/Support for reverse iterators
- [x] Deduction guides for gsl::dyn_array
- [x] Ensure support for clang, gcc, MSVC

### DIFF
--- a/.github/workflows/compilers.yml
+++ b/.github/workflows/compilers.yml
@@ -88,8 +88,6 @@ jobs:
           # Regular MSVC builds use Ninja (from preset)
           - toolset: ''
             generator_override: ''
-            build_config_args: ''
-            test_config_args: ''
           # ClangCL builds require Visual Studio generator
           - toolset: 'ClangCL'
             generator_override: '-G "Visual Studio 17 2022" -T ClangCL'
--- a/.github/workflows/compilers.yml
+++ b/.github/workflows/compilers.yml
@@ -95,6 +95,7 @@ jobs:
     steps:
       - uses: actions/checkout@v6
       - uses: microsoft/setup-msbuild@v3
+      - uses: ilammy/msvc-dev-cmd@v1
 
       - name: Run CMake (configure, build, test)
         uses: ./.github/workflows/cmake
--- a/.github/workflows/compilers.yml
+++ b/.github/workflows/compilers.yml
@@ -84,6 +84,15 @@ jobs:
         build_type: [ Debug, Release ]
         toolset: [ '', 'ClangCL' ]
         cxx_version: [ 14, 17, 20, 23 ]
+        include:
+          # Regular MSVC builds use Ninja (from preset)
+          - toolset: ''
+            generator_override: ''
+            build_config_args: ''
+            test_config_args: ''
+          # ClangCL builds require Visual Studio generator
+          - toolset: 'ClangCL'
+            generator_override: '-G "Visual Studio 17 2022" -T ClangCL'
     runs-on: ${{ matrix.image }}
     steps:
       - uses: actions/checkout@v6
@@ -93,7 +102,7 @@ jobs:
         uses: ./.github/workflows/cmake
         with:
           cmake_preset: msvc-${{ matrix.cxx_version }}-${{ matrix.build_type == 'Debug' && 'debug' || 'release' }}
-          extra_cmake_configure_args: ${{ matrix.toolset != '' && format('-G "Visual Studio 17 2022" -T {0}', matrix.toolset) || '' }}
-          extra_cmake_build_args: ${{ matrix.toolset != '' && format('--config {0}', matrix.build_type) || '' }}
-          extra_ctest_args: ${{ matrix.toolset != '' && format('-C {0}', matrix.build_type) || '' }}
+          extra_cmake_configure_args: ${{ matrix.generator_override }}
+          extra_cmake_build_args: ${{ matrix.toolset == 'ClangCL' && format('--config {0}', matrix.build_type) || '' }}
+          extra_ctest_args: ${{ matrix.toolset == 'ClangCL' && format('-C {0}', matrix.build_type) || '' }}
 
--- a/.github/workflows/compilers.yml
+++ b/.github/workflows/compilers.yml
@@ -83,7 +83,7 @@ jobs:
         generator: [ 'Visual Studio 17 2022' ]
         image: [ windows-2022, windows-2025 ]
         build_type: [ Debug, Release ]
-        extra_args: [ '', '-T ClangCL' ]
+        extra_args: [ '', '-DCMAKE_CXX_COMPILER=clang-cl -DCMAKE_C_COMPILER=clang-cl' ]
         cxx_version: [ 14, 17, 20, 23 ]
     runs-on: ${{ matrix.image }}
     steps:
--- a/.github/workflows/compilers.yml
+++ b/.github/workflows/compilers.yml
@@ -77,16 +77,16 @@ jobs:
           cmake_preset: clang-${{ matrix.cxx_version }}-${{ matrix.build_type == 'Debug' && 'debug' || 'release' }}
           extra_cmake_configure_args: '-DCMAKE_CXX_FLAGS="-isysroot \"$(xcode-select --print-path)/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk\""'
 
-  VisualStudio:
+  msvc:
     strategy:
       matrix:
         image: [ windows-2022, windows-2025 ]
         build_type: [ Debug, Release ]
-        toolset: [ '', 'ClangCL' ]
+        toolset: [ 'msvc', 'ClangCL' ]
         cxx_version: [ 14, 17, 20, 23 ]
         include:
           # Regular MSVC builds use Ninja (from preset)
-          - toolset: ''
+          - toolset: 'msvc'
             generator_override: ''
           # ClangCL builds require Visual Studio generator
           - toolset: 'ClangCL'
@@ -104,4 +104,3 @@ jobs:
           extra_cmake_configure_args: ${{ matrix.generator_override }}
           extra_cmake_build_args: ${{ matrix.toolset == 'ClangCL' && format('--config {0}', matrix.build_type) || '' }}
           extra_ctest_args: ${{ matrix.toolset == 'ClangCL' && format('-C {0}', matrix.build_type) || '' }}
-
--- a/.github/workflows/compilers.yml
+++ b/.github/workflows/compilers.yml
@@ -80,10 +80,9 @@ jobs:
   VisualStudio:
     strategy:
       matrix:
-        generator: [ 'Visual Studio 17 2022' ]
         image: [ windows-2022, windows-2025 ]
         build_type: [ Debug, Release ]
-        extra_args: [ '', '-DCMAKE_CXX_COMPILER=clang-cl -DCMAKE_C_COMPILER=clang-cl' ]
+        toolset: [ '', 'ClangCL' ]
         cxx_version: [ 14, 17, 20, 23 ]
     runs-on: ${{ matrix.image }}
     steps:
@@ -94,6 +93,7 @@ jobs:
         uses: ./.github/workflows/cmake
         with:
           cmake_preset: msvc-${{ matrix.cxx_version }}-${{ matrix.build_type == 'Debug' && 'debug' || 'release' }}
-          extra_cmake_configure_args: ${{ matrix.extra_args }}
-          extra_cmake_build_args: --config ${{ matrix.build_type }}
-          extra_ctest_args: -C ${{ matrix.build_type }}
+          extra_cmake_configure_args: ${{ matrix.toolset != '' && format('-G "Visual Studio 17 2022" -T {0}', matrix.toolset) || '' }}
+          extra_cmake_build_args: ${{ matrix.toolset != '' && format('--config {0}', matrix.build_type) || '' }}
+          extra_ctest_args: ${{ matrix.toolset != '' && format('-C {0}', matrix.build_type) || '' }}
+
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -6,6 +6,7 @@
             "hidden": true,
             "binaryDir": "${sourceDir}/build/${presetName}",
             "installDir": "${sourceDir}/install/${presetName}",
+            "generator": "Ninja",
             "cacheVariables": {
                 "GSL_CXX_STANDARD": "14",
                 "GSL_TEST": "ON"
@@ -15,14 +16,11 @@
             "name": "msvc-base",
             "inherits": "base",
             "hidden": true,
-            "generator": "Visual Studio 17 2022",
             "condition": {
                 "type": "equals",
                 "lhs": "${hostSystemName}",
                 "rhs": "Windows"
             },
-            "toolset": "host=x64",
-            "architecture": "x64",
             "cacheVariables": {
                 "CMAKE_CXX_COMPILER": "cl"
             }
@@ -31,7 +29,6 @@
             "name": "gcc-base",
             "inherits": "base",
             "hidden": true,
-            "generator": "Unix Makefiles",
             "cacheVariables": {
                 "CMAKE_CXX_COMPILER": "g++",
                 "CMAKE_C_COMPILER": "gcc"
@@ -41,7 +38,6 @@
             "name": "clang-base",
             "inherits": "base",
             "hidden": true,
-            "generator": "Unix Makefiles",
             "cacheVariables": {
                 "CMAKE_CXX_COMPILER": "clang++",
                 "CMAKE_C_COMPILER": "clang"
--- a/docs/headers.md
+++ b/docs/headers.md
@@ -9,6 +9,7 @@ See [GSL: Guidelines support library](https://isocpp.github.io/CppCoreGuidelines
 - [`<algorithms>`](#user-content-H-algorithms)
 - [`<assert>`](#user-content-H-assert)
 - [`<byte>`](#user-content-H-byte)
+- [`<dyn_array>`](#user-content-H-dyn_array)
 - [`<gsl>`](#user-content-H-gsl)
 - [`<narrow>`](#user-content-H-narrow)
 - [`<pointers>`](#user-content-H-pointers)
@@ -154,6 +155,10 @@ constexpr byte to_byte() noexcept;
 ```
 
 Convert the given value `I` to a `byte`. The template requires `I` to be in the valid range 0..255 for a `gsl::byte`.
+
+## <a name="H-dyn_array" />`<dyn_array>`
+
+# TODO (@carsonradtke)
 
 ## <a name="H-gsl" />`<gsl>`
 
--- a/docs/headers.md
+++ b/docs/headers.md
@@ -158,7 +158,175 @@ Convert the given value `I` to a `byte`. The template requires `I` to be in the 
 
 ## <a name="H-dyn_array" />`<dyn_array>`
 
-# TODO (@carsonradtke)
+This header contains an owning dynamically allocated array type whose size is fixed between assignments.
+
+- [`gsl::dyn_array`](#user-content-H-dyn_array-dyn_array)
+
+### <a name="H-dyn_array-dyn_array" />`gsl::dyn_array`
+
+```cpp
+template <typename T, typename Allocator = std::allocator<T>>
+class dyn_array;
+```
+
+`gsl::dyn_array` owns a contiguous sequence of `T` objects allocated with `Allocator`.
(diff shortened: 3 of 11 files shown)


include/gsl/dyn_array
```diff
@@ -0,0 +1,352 @@
...
+{
+    template <typename T>
+    struct dyn_array_traits
+    {
+        using value_type = T;
+        using pointer = T*;
+        using reference = T&;
+        using const_reference = const T&;
+        using difference_type = std::ptrdiff_t;
+        using size_type = decltype(sizeof(0));
```
carsonRadtke:
```suggestion
        using size_type = std::size_t;
```

include/gsl/dyn_array
```diff
@@ -0,0 +1,352 @@
...
+
+        const class dyn_array_impl
+        {
+        public:
+            constexpr dyn_array_impl(pointer data, size_type count) : _data{data}, _count{count}
+            {
+                Ensures((_count == 0 && _data == nullptr) || (_count > 0 && _data != nullptr));
+            }
+
+            constexpr auto data() const { return _data; }
```
carsonRadtke:
```suggestion
            constexpr auto data() { return _data; }
            constexpr auto data() const -> const pointer { return _data; }
```

include/gsl/dyn_array
```diff
@@ -0,0 +1,352 @@
...
+
+            constexpr auto count() const { return _count; }
+
+        private:
+            pointer _data;
+            size_type _count;
+        } _impl;
+
+    public:
+        constexpr dyn_array_base(const Allocator& alloc) : Allocator{alloc}, _impl{nullptr, 0} {}
```
carsonRadtke:
```suggestion
        constexpr explicit dyn_array_base(const Allocator& alloc) : Allocator{alloc}, _impl{nullptr, 0} {}
```

tests/CMakeLists.txt
```diff
@@ -151,6 +151,7 @@ else()
           -Wno-unknown-attributes
           -Wno-used-but-marked-unused # GTest EXPECT_DEATH
           -Wno-weak-vtables
+          -Wno-poison-system-directories
```
beinhaerter: Why is this added? `poison-system-directories` looks like a reasonable warning to me, why is it disabled here?

tests/dyn_array_tests.cpp
```diff
@@ -0,0 +1,290 @@
...
+    EXPECT_EQ(*it++, 'b');
+    EXPECT_EQ(*it++, 'a');
+}
+
+#ifdef GSL_HAS_DEDUCTION_GUIDES
+TEST(dyn_array_tests, deduction_guides)
+{
+    std::vector<char> giants{10};
+#ifdef GSL_HAS_CONTAINER_RANGES
+    gsl::dyn_array mariners(std::from_range_t{}, giants);
```
beinhaerter:
```suggestion
    gsl::dyn_array mariners(std::from_range, giants);
```
https://en.cppreference.com/w/cpp/ranges/from_range.html

include/gsl/dyn_array
```diff
@@ -0,0 +1,352 @@
...
+        {
+            auto rv = *this;
+            ++(*this);
+            return rv;
+        }
+
+        constexpr auto operator--() -> dyn_array_iterator&
+        {
+            Expects(_pos > 0);
+            _pos--;
```
beinhaerter:
```suggestion
            --_pos;
```

Compare Enforcement in https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#p9-dont-waste-time-or-space

include/gsl/dyn_array
```diff
@@ -0,0 +1,352 @@
...
+
+        constexpr auto operator[](size_type pos) const -> const_reference
+        {
+            return const_cast<dyn_array_iterator&>(*this).operator[](pos);
+        }
+
+    private:
+        pointer _ptr;
+        size_type _pos;
+        size_type _end_pos;
```
beinhaerter:
```suggestion
        pointer _ptr{};
        size_type _pos{};
        size_type _end_pos{};
```
https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c48-prefer-default-member-initializers-to-member-initializers-in-constructors-for-constant-initializers

include/gsl/dyn_array
```diff
@@ -0,0 +1,352 @@
...
+    using reverse_iterator = std::reverse_iterator<iterator>;
+    using const_reverse_iterator = std::reverse_iterator<const_iterator>;
+    using difference_type = typename details::dyn_array_traits<T>::difference_type;
+    using size_type = typename details::dyn_array_traits<T>::size_type;
+
+    using allocator_type = Allocator;
+
+    explicit constexpr dyn_array(const Allocator& alloc = {}) : base{alloc} {}
+
+    explicit constexpr dyn_array(size_type count, const T& value, const Allocator& alloc = {})
```
beinhaerter:
```suggestion
    constexpr dyn_array(size_type count, const T& value, const Allocator& alloc = {})
```
This ctor takes at least two params, so `explicit` is not needed.

include/gsl/dyn_array
```diff
@@ -0,0 +1,354 @@
...
+// THE SOFTWARE.
+//
+///////////////////////////////////////////////////////////////////////////////
+
+#ifndef GSL_DYN_ARRAY_H
+#define GSL_DYN_ARRAY_H
+
+#include "./assert"
+#include "./narrow"
+#include "./util"
```
gdr-at-ms: Use `gsl/` scoping prefix

include/gsl/dyn_array
```diff
@@ -0,0 +1,354 @@
...
+#include <memory>
+
+#ifdef GSL_HAS_RANGES
+#include <ranges>
+#endif /* GSL_HAS_RANGES */
+
+namespace gsl
+{
+namespace details
+{
```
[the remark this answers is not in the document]
beinhaerter: From the README:
> The entire implementation is provided inline in the headers under the [gsl]https://github.com/microsoft/GSL/blob/main/include/gsl) directory. The implementation generally assumes a platform that implements C++14 support.

`namespace gsl::details` is C++17. Unfortunately in my company we still can't use C++17 in general, so I would prefer to still support C++14 if possible.

include/gsl/dyn_array
carsonRadtke: We changed this https://github.com/microsoft/GSL/pull/1194. The scoping prefix contradicts [SF.12](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf12-prefer-the-quoted-form-of-include-for-files-relative-to-the-including-file-and-the-angle-bracket-form-everywhere-else) in the guidelines.

include/gsl/dyn_array
```diff
@@ -0,0 +1,354 @@
...
+
+#include "./assert"
+#include "./narrow"
+#include "./util"
+
+#include <algorithm>
+#include <iterator>
+#include <memory>
+
+#ifdef GSL_HAS_RANGES
```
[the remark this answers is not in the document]
carsonRadtke: We get a compiler warning if we try to include `<ranges>` outside of C++23. I felt `GSL_HAS_RANGES` was more expressive than `#if (defined(__cpp_lib_ranges) && __cpp_lib_ranges >= 201911L)`.

include/gsl/pointers
```diff
@@ -338,15 +339,15 @@ auto make_strict_not_null(T&& t) noexcept
     return strict_not_null<std::remove_cv_t<std::remove_reference_t<T>>>{std::forward<T>(t)};
 }
 
-#if (defined(__cpp_deduction_guides) && (__cpp_deduction_guides >= 201611L))
+#ifdef GSL_HAS_DEDUCTION_GUIDES
```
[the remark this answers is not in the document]
carsonRadtke: I needed to check for deduction guides in the dyn_array implementation and creating the `GSL_HAS_XXX` macro is more expressive than the alternative.

include/gsl/dyn_array
```diff
@@ -0,0 +1,441 @@
...
+
+        constexpr dyn_array_iterator(pointer ptr, size_type pos, size_type end_pos)
+            : _ptr{ptr}, _pos{pos}, _end_pos{end_pos}
+        {
+            Ensures((_ptr != nullptr && _end_pos > 0) || (_ptr == nullptr && _end_pos == 0));
+            Ensures(_pos <= _end_pos);
+        }
+
+#if defined(_MSC_VER) && defined(GSL_HAS_RANGES)
+        // TODO (@carsonradtke): Investigate why this is necessary for MSVC.
```
carsonRadtke: TODO

include/gsl/dyn_array
```diff
@@ -0,0 +1,441 @@
...
+} // namespace details
+
+template <typename T, typename Allocator = std::allocator<T>>
+class dyn_array : private details::dyn_array_base<T, Allocator>
+{
+    using base = details::dyn_array_base<T, Allocator>;
+    using pointer = T*;
+
+    constexpr void swap(dyn_array& other) {
+        // TODO (@carsonradtke): This was not in the spec, should it have been?
```
carsonRadtke: This TODO is about making `swap` public. @gdr-at-ms, do we need to ask the guidelines folks first?

include/gsl/dyn_array
```diff
@@ -0,0 +1,441 @@
...
+    template <typename InputIt, std::enable_if_t<details::is_fwd_iterator<InputIt>::value, bool> = true>
+    constexpr dyn_array(InputIt first, InputIt last, const Allocator& alloc = {})
+        : base{gsl::narrow<size_type>(std::distance(first, last)), alloc}
+    {
+        base::copy(first, last, data());
+    }
+
+    template <typename InputIt, std::enable_if_t<!details::is_fwd_iterator<InputIt>::value && details::is_iterator<InputIt>::value, bool> = true>
+    constexpr dyn_array(InputIt first, InputIt last, const Allocator& alloc={}) : dyn_array{alloc} {
+        // TODO (@carsonradtke): Can this be done better?
```
carsonRadtke: Looking for opinions/ideas.

include/gsl/dyn_array
```diff
@@ -0,0 +1,352 @@
...
+    using size_type = typename details::dyn_array_traits<T>::size_type;
+
+    using allocator_type = Allocator;
+
+    explicit constexpr dyn_array(const Allocator& alloc = {}) : base{alloc} {}
+
+    explicit constexpr dyn_array(size_type count, const T& value, const Allocator& alloc = {})
+        : base{count, alloc}
+    {
+        std::fill(begin(), end(), value);
```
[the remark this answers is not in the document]
carsonRadtke: There is now `dyn_array_base::{fill,copy}` that starts and ends lifetimes. I don't have test coverage for this yet.

CMakePresets.json
```diff
@@ -6,6 +6,7 @@
             "hidden": true,
             "binaryDir": "${sourceDir}/build/${presetName}",
             "installDir": "${sourceDir}/install/${presetName}",
+            "generator": "Ninja",
```
carsonRadtke, line 9: This change is certainly out of scope, but it was easier than downloading VS17 locally.

include/gsl/dyn_array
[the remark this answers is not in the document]
carsonRadtke: test coverage has been added

include/gsl/dyn_array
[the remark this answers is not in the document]
carsonRadtke: Removing all GSL_HAS_*

include/gsl/pointers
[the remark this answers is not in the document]
carsonRadtke: Removing all GSL_HAS_*

include/gsl/dyn_array
```diff
@@ -0,0 +1,352 @@
...
+
+        constexpr auto operator[](size_type pos) -> reference
+        {
+            Expects(_pos + pos < _end_pos);
+            return _ptr[_pos + pos];
+        }
+
+        constexpr auto operator[](size_type pos) const -> const_reference
+        {
+            return const_cast<dyn_array_iterator&>(*this).operator[](pos);
```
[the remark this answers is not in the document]
carsonRadtke: I spent a lot of time investigating this, and I'm convinced that this case is an analyzer false negative. I'll probably end up reporting this to the MSVC folks with the internal details, but for now I'm not going to do any sort of suppression.

tests/CMakeLists.txt
```diff
@@ -76,6 +76,10 @@ include(CheckCXXCompilerFlag)
 # please try to keep entries ordered =)
 add_library(gsl_tests_config INTERFACE)
 if(MSVC) # MSVC or simulating MSVC
+    set(CMAKE_VS_GLOBALS
+        "EnableMicrosoftCodeAnalysis=true"
+        "CodeAnalysisRuleSet=AllRules.ruleset"
+    )
```
carsonRadtke: I did not mean to commit this.

.github/workflows/compilers.yml
```diff
@@ -80,20 +80,28 @@ jobs:
   VisualStudio:
```
carsonRadtke: Since we are using ninja now, might as well call this "MSVC".

include/gsl/dyn_array
```diff
@@ -0,0 +1,453 @@
...
+
+    constexpr dyn_array(std::initializer_list<T> init, const Allocator& alloc = {})
+        : dyn_array(init.begin(), init.end(), alloc)
+    {}
+
+    constexpr auto operator=(const dyn_array& other) -> dyn_array& {
+        dyn_array tmp{other, get_allocator()};
+        swap(tmp);
+        return *this;
+    }
```
carsonRadtke: copy assignment needs to be deleted. Consider:

```cpp
TEST(dyn_array_tests, iterator_invalidation)
{
    gsl::dyn_array<int> arr1 = {1, 2, 3};
    gsl::dyn_array<int> arr2 = {4, 5, 6};

    auto start = std::begin(arr1);

    arr1 = arr2;

    EXPECT_EQ(*start, /* ??? UB ??? */);
}
```

`arr1 = arr2` invalidates the `start` iterator.

include/gsl/util
```diff
@@ -95,6 +103,27 @@ namespace gsl
...
+    template <typename T>
+    struct is_iterator<T, void_t<typename std::iterator_traits<T>::value_type>>
+        : std::true_type {};
+
+    template <typename T, typename = void>
+    struct is_fwd_iterator : std::false_type
+    {};
+
+   template <typename T>
+    struct is_fwd_iterator<T, void_t<typename std::iterator_traits<T>::iterator_category>>
```
carsonRadtke, line 127: nit: formatting

docs/headers.md
```diff
@@ -155,6 +156,178 @@ constexpr byte to_byte() noexcept;
 
 Convert the given value `I` to a `byte`. The template requires `I` to be in the valid range 0..255 for a `gsl::byte`.
 
+## <a name="H-dyn_array" />`<dyn_array>`
+
+This header contains an owning dynamically allocated array type whose size is fixed between assignments.
```
carsonRadtke, line 161:
```suggestion
This header contains an owning dynamically allocated array type whose size is fixed at construction.
```

docs/headers.md
```diff
@@ -155,6 +156,178 @@ constexpr byte to_byte() noexcept;
...
+
+### <a name="H-dyn_array-dyn_array" />`gsl::dyn_array`
+
+```cpp
+template <typename T, typename Allocator = std::allocator<T>>
+class dyn_array;
+```
+
+`gsl::dyn_array` owns a contiguous sequence of `T` objects allocated with `Allocator`.
+The number of elements is established when the object is constructed and remains unchanged until the object is copy-assigned.
```
carsonRadtke, line 173:
```suggestion
The number of elements is established when the object is constructed and remains unchanged.
```

docs/headers.md
```diff
@@ -155,6 +156,178 @@ constexpr byte to_byte() noexcept;
...
+Constructs a `dyn_array` by copying the elements from another `dyn_array` or from an initializer list.
+
+```cpp
+constexpr auto operator=(const dyn_array& other) -> dyn_array&;
+
+constexpr dyn_array(dyn_array&&) = delete;
+dyn_array& operator=(dyn_array&&) = delete;
+```
+
+Copy assignment replaces the contents with copies of the elements in `other`.
```
carsonRadtke, line 245: This statement is not true. Delete this line and update the following line to reflect that copy assignment is explicitly deleted.

docs/headers.md
```diff
@@ -155,6 +156,178 @@ constexpr byte to_byte() noexcept;
...
+constexpr auto begin();
+constexpr auto begin() const;
+constexpr auto cbegin() const;
+
+constexpr auto end();
+constexpr auto end() const;
+constexpr auto cend() const;
+```
+
+Returns an iterator to the first element or to one past the last element.
```
carsonRadtke, line 289: Maybe mention that this iterator can never be invalidated?

carsonRadtke (commented):
@copilot, please address these comments.

gdr-at-ms (approved):

PR #1245, opened 2026-05-26
gsl::strict_not_null: C.66 Make move ctor noexcept

maflcko:
The `strict_not_null` move ctor actually calls the copy ctor, as explained in the comment added in commit d69e578519840b9b74eff26ac01465ac07698063. The comment also says that a custom pointer type copy ctor "could in theory throw exceptions". However, I don't think that such a smart pointer type should exist in practise.

Also, it violates C66, see  https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c66-make-move-operations-noexcept

So just use a `noexcept` for all types.

### DIFF
--- a/include/gsl/pointers
+++ b/include/gsl/pointers
@@ -298,8 +298,8 @@ public:
     {}
 
     // To avoid invalidating the "not null" invariant, the contained pointer is actually copied
-    // instead of moved. If it is a custom pointer, its constructor could in theory throw exceptions.
-    strict_not_null(strict_not_null&& other) noexcept(std::is_nothrow_copy_constructible<T>::value) = default;
+    // instead of moved. If it is a custom pointer type, its copy constructor must not throw exceptions.
+    strict_not_null(strict_not_null&& other) noexcept = default;
     strict_not_null(const strict_not_null& other) = default;
     strict_not_null& operator=(const strict_not_null& other) = default;
     strict_not_null& operator=(const not_null<T>& other)


carsonRadtke:
> However, I don't think that such a smart pointer type should exist in practise.

[boost::intrusive_ptr](https://www.boost.org/doc/libs/1_60_0/libs/smart_ptr/intrusive_ptr.html) exists and has a potentially throwing copy constructor.

> Also, it violates C66

It does. I think this is going to be a case where the GSL implementation is not going to conform to the guidelines because a specific decision was made to implement the `gsl::strict_not_null` move constructor with copy constructor semantics (which also violates [C.64: A move operation should move and leave its source in a valid state](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#rc-move-semantic)). If you think that was not a good decision, I'm happy to discuss it, but for now I think we should leave the constructor as potentially throwing.

WDYT? @gdr-at-ms

maflcko:
> > However, I don't think that such a smart pointer type should exist in practise.
> 
> [boost::intrusive_ptr](https://www.boost.org/doc/libs/1_60_0/libs/smart_ptr/intrusive_ptr.html) exists and has a potentially throwing copy constructor.

Ok, I see. I think GSL being flexible and fully supporting such a throwing copy ctor seems reason enough to close this. After all, everyone who marks their copy ctors as `noexcept` won't see any difference here anyway.

Closing for now, but happy to re-open, if GSL decides to follow C.66

Issue #862, opened 2020-04-15
delete move ctor for not_null and strict_not_null


JordanMaples:
#842 had a few unresolved comments and to-do items that came to light after the PR was merged. 
The move ctor for not_null and strict_not_null need to be explicitly marked as `delete`.

jwakely:
No no no.

Issue #1246, opened 2026-05-26
move ctor for strict_not_null should be removed


beinhaerter:
In commit 9ff6e19ea9e2ecf48488c0865be85bb1ca183279 (issue #743) in 2018 `strict_not_null` was introduced. The interface was almost identical to the one of `not_null`, but the constructor is `explicit`.

Later in commit 6b23937bafca79ddd9b544e337aaba36a06ce339 in 2020 the move ctor for `not_null` was removed, but the move constructor for `strict_not_null` was not removed.

In my opinion the move constructor for `strict_not_null` should be removed as it was done with the move constructor for `not_null`. This aligns the interfaces of both classes to be the same. And it removes code that looks like it moves but instead copies, which is at least surprising.

beinhaerter:
Thanks to @maflcko who brought my attention to it in https://github.com/microsoft/GSL/pull/1245.

maflcko:
(off-topic: Heh, I still prefer https://github.com/microsoft/GSL/pull/1238 to make it movable, but this seems possibly controversial and also off-topic here)

on-topic: I agree that removing it is the least confusing and most self-consistent fix for now.

carsonRadtke:
Are you wanting the user-defined `strict_not_null(strict_not_null&& other)` to be `= delete;`? or to just be removed? The former has already been discussed (and rejected) in #842. The latter is not really a behavior change. 

>  And it removes code that looks like it moves but instead copies, which is at least surprising.

Whether you remove it or keep it, a copy will happen. This behavior is intentional and is already the case with `gsl::not_null`. https://godbolt.org/z/svh1f8hdd

> This aligns the interfaces of both classes to be the same.

I'm happy to accept a PR that makes the interfaces match.

---

That being said, I interpret the description of #743 as "We introduced `gsl::strict_not_null` to help customers upgrade". This type is not something specified by the core guidelines (as far as I can tell). Perhaps at this point the best decision is to remove `gsl::strict_not_null` in the next major GSL release (w/ `gsl::dyn_array`)? 

WDYT? @gdr-at-ms

carsonRadtke:
> Are you wanting ...

Edit: I just saw your PR. "just be removed" is the answer.

maflcko:
[...]

PR #1247, opened 2026-05-26
remove strict_not_null move ctor, fixes issue 1246

beinhaerter:


maflcko (approved):
lgtm. This seems more self-consistent to not offer a move constructor, if only copies can be done anyway.

carsonRadtke (approved):
LGTM, for completeness this also (kind of) closes #862.

 Before merging, I'll wait for acknowledment (either here or in #1246) that `gsl::strict_not_null<int*> snn_copy = std::move(snn_original);` will still compile and will still perform a copy (same behavior as `gsl::not_null`).

maflcko:
> `gsl::strict_not_null<int*> snn_copy = std::move(snn_original);` will still compile and will still perform a copy (same behavior as `gsl::not_null`).

Yes, that should be intentional. Just like the nn-type in the std-lib (reference_wrapper):  https://godbolt.org/z/35a9c5efG and also: https://github.com/microsoft/GSL/pull/842#issuecomment-674570825

carsonRadtke (changes requested):
(no change requests, I just can't figure out how to unapprove)

Edit: apologies for the noise. I don't want to merge this until we reach a decision in 1246.

maflcko:
Needs rebase or merge with master (for the clang-format whitespace changes). Otherwise, this still looks good.

PR #1250, opened 2026-05-27
narrow_cast instead of static_cast

beinhaerter:
This removes the "warning C26472: Don't use a static_cast for arithmetic conversions. Use brace initialization, gsl::narrow_cast or gsl::narrow (type.1)." issued by VS2026.


carsonRadtke (approved):
Good find, thanks!

PR #1251, opened 2026-05-28
clang-format improvements

beinhaerter:
- Add a clang-format linter check to the PR pipeline
- Add a Windows `bat` file and a Linux `sh` script to apply formatting to all sources files, and on Linux also to all `.sh` shell scripts. Add a linter check to the pipeline for the `.sh` shell scripts (calling `shfmt` and `shellcheck`).
- Run the formatting script(s) to apply clang-format and fix the clang-format linter errors
- Remove `CommentPragmas` from `.clang-format`
- Add `WhitespaceSensitiveMacros: [GSL_SUPPRESS]` to `.clang-format`
- Remove all `// clang-format off` and `// NO-FORMAT` as they are not needed any more
- Remove a commented out `GSL_SUPPRESS`

### DIFF
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -22,5 +22,5 @@ jobs:
       - name: Check format
         uses: jidicula/clang-format-action@v4.11.0
         with:
-          clang-format-version: '19'
+          clang-format-version: '20'
           check-path: 'include tests'
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -34,6 +34,6 @@ jobs:
       - name: Check format
         run: |
           {
-            find include -type f
+            find include/gsl -type f
             find tests -type f \( -name '*.cpp' -o -name '*.h' \)
           } | xargs clang-format-${{ env.CLANG_VERSION }} --dry-run --Werror
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -18,6 +18,10 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
 
+      # Prints the version of clang-format being used
+      - name: Log clang-format version
+        run: clang-format --version
+        
       # Runs clang-format over the repository codebase
       - name: Check format
         run: |
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -1,0 +1,26 @@
+name: Code Formatting
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+permissions:
+  contents: read
+
+jobs:
+  clang-format:
+    name: Run clang-format
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      # Runs clang-format over the repository codebase
+      - name: Check format
+        uses: jidicula/clang-format-action@v4.11.0
+        with:
+          clang-format-version: '19'
+          check-path: 'include tests'
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -9,6 +9,9 @@ on:
 permissions:
   contents: read
 
+env:
+  CLANG_VERSION: "20"
+
 jobs:
   clang-format:
     name: Run clang-format
@@ -18,9 +21,14 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
 
+      # Install the exact clang-format binary
+      - name: Install clang-format
+        run: |
+          sudo apt install clang-format-${{ env.CLANG_VERSION }}
+
       # Prints the version of clang-format being used
       - name: Log clang-format version
-        run: clang-format --version
+        run: clang-format-${{ env.CLANG_VERSION }} --version
         
       # Runs clang-format over the repository codebase
       - name: Check format
@@ -28,4 +36,4 @@ jobs:
           {
             find include -type f
             find tests -type f \( -name '*.cpp' -o -name '*.h' \)
-          } | xargs clang-format --dry-run --Werror
+          } | xargs clang-format-${{ env.CLANG_VERSION }} --dry-run --Werror
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -20,7 +20,8 @@ jobs:
 
       # Runs clang-format over the repository codebase
       - name: Check format
-        uses: jidicula/clang-format-action@v4.11.0
-        with:
-          clang-format-version: '20'
-          check-path: 'include tests'
+        run: |
+          {
+            find include -type f
+            find tests -type f \( -name '*.cpp' -o -name '*.h' \)
+          } | xargs clang-format --dry-run --Werror
--- a/tests/span_tests.cpp
+++ b/tests/span_tests.cpp
@@ -70,7 +70,9 @@ struct AddressOverloaded
     [[maybe_unused]]
 #endif
     AddressOverloaded operator&() const
-    { return {}; }
+    {
+        return {};
+    }
 };
 } // namespace
 
--- a/tests/span_tests.cpp
+++ b/tests/span_tests.cpp
@@ -257,12 +257,12 @@ TEST(span_test, from_pointer_pointer_construction)
         EXPECT_TRUE(s.data() == &arr[0]);
     }
 
-#if 0
-    { // this test succeeds on all platforms, gsl::span is more relaxed than std::span where this would be UB
-        auto workaround_macro = [&]() { span<int> s{&arr[1], &arr[0]}; };
-        EXPECT_DEATH(workaround_macro(), expected);
-    }
-#endif
+    // this test succeeds on all platforms, gsl::span is more relaxed than std::span where this
+    // would be UB
+    //{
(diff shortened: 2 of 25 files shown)


beinhaerter:
I am not github pipeline expert. The clang-format.yml file is AI generated. I have no idea if it is OK or not.

tests/utils_tests.cpp
```diff
@@ -16,7 +16,7 @@
 
 #include <gtest/gtest.h>
 
-#include <algorithm>   // for move
+#include <algorithm> // for move
```
carsonRadtke, line 19: Interesting decision by clang-format. I guess there isn't really anything to do about it.

tests/span_tests.cpp
```diff
@@ -257,10 +257,12 @@ TEST(span_test, from_pointer_pointer_construction)
         EXPECT_TRUE(s.data() == &arr[0]);
     }
 
-    //{ // this test succeeds on all platforms, gsl::span is more relaxed than std::span where this would be UB
-    //    auto workaround_macro = [&]() { span<int> s{&arr[1], &arr[0]}; };
-    //    EXPECT_DEATH(workaround_macro(), expected);
-    //}
+#if 0
```
carsonRadtke: Out of curiosity, did you change the comment to `#if 0`? Or was this clang-format?

.github/workflows/clang-format.yml
```diff
@@ -0,0 +1,26 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      # Runs clang-format over the repository codebase
+      - name: Check format
+        uses: jidicula/clang-format-action@v4.11.0
+        with:
+          clang-format-version: '20'
+          check-path: 'include tests'
```
carsonRadtke: Taking a dependency on third-party actions makes me a bit uneasy. Can we change this to:
```suggestion
      - name: Check format
        run: |
          {
            find include -type f
            find tests -type f \( -name '*.cpp' -o -name '*.h' \)
          } | xargs clang-format --dry-run --Werror
```

(Also used AI for this, the bracket syntax is new to me)

<strike>You may need to install clang-format. Either from `apt` or `pip` or something.</strike>
Edit: [Looks like ubuntu-latest already has clang-format installed.](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md)

carsonRadtke (changes requested):
Thanks for making these changes! The only blocker is the third-party actions dependency - if you wouldn't mind changing that to a shell command, that'd be appreciated.

beinhaerter:
> Thanks for making these changes! The only blocker is the third-party actions dependency - if you wouldn't mind changing that to a shell command, that'd be appreciated.

OK, I'll give it a try.

tests/span_tests.cpp
beinhaerter: That was me. clang-format decided to reformat to
```
    //{ // this test succeeds on all platforms, gsl::span is more relaxed than std::span where this
    //would be UB
    //    auto workaround_macro = [&]() { span<int> s{&arr[1], &arr[0]}; };
    //    EXPECT_DEATH(workaround_macro(), expected);
    //}
```
I did not like the formatting of the comment inside the comment.
Of course we can also put the comment before or after the opening brace, that looks a bit better:
```
    // this test succeeds on all platforms, gsl::span is more relaxed than std::span where this
    // would be UB
    //{
    //    auto workaround_macro = [&]() { span<int> s{&arr[1], &arr[0]}; };
    //    EXPECT_DEATH(workaround_macro(), expected);
    //}
```

```
    // {
    //     // this test succeeds on all platforms, gsl::span is more relaxed than std::span where
    //     // this would be UB
    //     auto workaround_macro = [&]() { span<int> s{&arr[1], &arr[0]}; };
    //     EXPECT_DEATH(workaround_macro(), expected);
    // }
```
Or we can raise the `ColumnLimit` in `.clang-format`.

.github/workflows/clang-format.yml
beinhaerter: Which version of clang-format do we want to use? The one provided with the lastest Visual Studio (makes it easy for me and probably for you?)? Or the one installed on the Ubuntu that the pipeline is using? Or some other dedicated version?

If we don't write the exact version into the yml file (calling `clang-format` and not for example `clang-format-20`), it would be good to print out the clang-format version being used. Otherwise contributors only see a pipeline failure but don't know how to fix it if their formatter format the code in another way.

Side note: on my Ubuntu 24.04 I needed to install clang-format first, it was not installed. `sudo apt install clang-format` gave me version 18. I was able to install up to version 20 (`sudo apt install clang-format-20`).

carsonRadtke: > Which version of clang-format do we want to use?

This one: "Or the one installed on the Ubuntu that the pipeline is using?"
I like the idea of printing out the version too. 

> Side note: on my Ubuntu 24.04 I needed to install clang-format first, it was not installed.

If this is a response to my (striked-out) comment: I wasn't claiming that it came pre-installed on Ubuntu, I was claiming it was part of the `ubuntu-latest` GitHub Runner Image (which is `ubuntu-latest` plus a bunch of devtools).

But yeah, on a fresh ubuntu install, users need to install clang-format.

> Otherwise contributors only see a pipeline failure but don't know how to fix it if their formatter format the code in another way.

I hope this doesn't become a problem and we have to make code changes every time GitHub bumps `clang-format` when the `ubuntu-latest` image is updated...

tests/span_tests.cpp
carsonRadtke: I'm fine with whatever you pick. Was just surprised that `clang-format` would do this, but I guess it didn't.

carsonRadtke:
> RE: installed version is 18 which replaces "GSL_SUPPRESS(bounds.1)" with "GSL_SUPPRESS(bounds .1)"

Consider adding `WhitespaceSensitiveMacros: [GSL_SUPPRESS]` to .clang-format so this doesn't happen?

.github/workflows/clang-format.yml
```diff
@@ -0,0 +1,39 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      # Install the exact clang-format binary
+      - name: Install clang-format
+        run: |
+          sudo apt install clang-format-${{ env.CLANG_VERSION }}
```
carsonRadtke, line 27: I'd rather use the pre-installed version, but I'm not going to block the PR on it.

carsonRadtke (approved):
Thanks for the changes! LGTM

I'll merge it on Monday in case you want to address my non-blocking comments.

beinhaerter:
> > RE: installed version is 18 which replaces "GSL_SUPPRESS(bounds.1)" with "GSL_SUPPRESS(bounds .1)"
> 
> Consider adding `WhitespaceSensitiveMacros: [GSL_SUPPRESS]` to .clang-format so this doesn't happen?

Here is a [patch](https://github.com/user-attachments/files/28407394/whitespacesensitivemacros.txt) for .clang-format together with the changes it does to the files. I understand that it fixes the whitespace-in-suppress-attributes-problem, but I am not sure if this looks better. Which version do you prefer? Shall I apply the patch or leave it as it is?

For example:
```
 template <class Cont>
-GSL_SUPPRESS(bounds.4)
-GSL_SUPPRESS(bounds.2) constexpr auto at(Cont& cont, const index i) -> decltype(cont[cont.size()])
+GSL_SUPPRESS(bounds.4) GSL_SUPPRESS(bounds.2) constexpr auto at(Cont& cont, const index i)
+    -> decltype(cont[cont.size()])
 {
```

.github/workflows/clang-format.yml
beinhaerter: The pre-installed version needs the `WhitespaceSensitiveMacros` change, clang-format-20 does not.

Regarding the burden on the authors: I think it might be a good idea to provide an `apply-clang-format.bat` and an `apply-clang-format.sh`. Of course the bat file would only work in the developer command prompt and could only use the clang-format version shipped with Visual Studio. What do you think about it? Shall I add those files to make formatting easier?

carsonRadtke: > Shall I add those files to make formatting easier?

That sounds good to me. Can you add it to a `<repo-root>/scripts` folder?

carsonRadtke: Alternatively, we could add it as a copilot skill so [at]Copilot can make the changes automatically?

This way we don't have to worry about what `clang-format` version is installed on the user's machine when they run the script. Copilot will automatically have whatever we specify in https://github.com/microsoft/GSL/blob/main/.github/workflows/copilot-setup-steps.yml.

Either way is fine by me.

carsonRadtke:
> but I am not sure if this looks better

> Which version do you prefer? Shall I apply the patch or leave it as it is?

I don't really have a preference on what "looks better", but I do think that having `WhitespaceSensitiveMacros: [GSL_SUPPRESS]` is correct and should be kept. My vote is "apply the patch".

.github/workflows/clang-format.yml
beinhaerter: Regarding the copilot-setup-steps.yml I have no idea how that all works together, what the best place is to do changes. I either need your guidance or you do your refinements after this PR is merged. Whatever you think is better.

carsonRadtke: The Copilot stuff is a "nice to have"; that's something I may follow-up on in the future.

carsonRadtke (approved):
LGTM

Thanks for the contribution!

PR #1252, opened 2026-05-29
Add gsl::span vs std::span performance benchmark CI (fixes #1167)

### LINKED ISSUE #1167: Add performance tests to PR actions
It is important to maintain performance parity to `std::span` (at the very least when [Safe Buffers](https://clang.llvm.org/docs/SafeBuffers.html) are enabled). In order to track this, we should have performance tests between `std::span` and `gsl::span` for the latest standard library implementation across Clang, GCC, and MSVC.

@galenelias put together small benchmark comparing the two implementations on various workloads. They have been attached for convenience and provide a good starting point.

[span_bench.zip](https://github.com/user-attachments/files/17723220/span_bench.zip)

Gafoor2005:
## Summary

Adds a continuous benchmark suite that tracks performance parity between
`gsl::span` and `std::span` on every PR, across all supported compilers
and platforms.

Closes #1167.

---

## What this adds

- `benchmark/span_bench.cpp` — benchmarks from @galenelias comparing
  `gsl::span` and `std::span` across five workloads (is_sorted, ranges,
  custom iterator loop, min_element algorithm, range-for min)
- `benchmark/CMakeLists.txt` — self-contained build; fetches
  google-benchmark v1.9.0 and googletest via FetchContent; links against
  the **local repo's GSL** (not a pinned tag) so every PR tests its own
  changes
- `benchmark/check_regression.py` — parses benchmark JSON output, pairs
  StdSpan/GslSpan variants by name, computes `gsl_ns / std_ns` ratio,
  writes a Markdown table, exits 1 on regression
- `.github/workflows/span_benchmark.yml` — runs on every PR across 13
  configs (GCC 13/14, Clang 17/18, MSVC 2022, clang-cl, Apple Clang ×
  C++20/23); posts or updates a single PR comment with results from all
  configs

---

## Approach

The comparison is **in-run ratio** (`gsl_ns / std_ns`) rather than
comparing against a stored baseline. Both span variants run in the same
process on the same machine at the same moment, so GitHub-hosted runner
noise (~10–15%) cancels out. A ratio above **1.15** (15% slower than
`std::span`) fails CI.

This sidesteps the "Status: Blocked" concern about noisy shared runners —
no self-hosted runner or external service needed.

Each benchmark runs **10 repetitions**; the script uses the mean and
reports stddev so reviewers can see measurement stability.

---

## CI matrix

| OS | Compiler | Standard |
|---|---|---|
| ubuntu-latest | GCC 13 | C++20 |
| ubuntu-latest | GCC 14 | C++20, C++23 |
| ubuntu-latest | Clang 17 | C++20 |
| ubuntu-latest | Clang 18 | C++20, C++23 |
| windows-latest | MSVC 2022 | C++20, C++23 |
| windows-latest | clang-cl | C++20, C++23 |
| macos-14 | Apple Clang | C++20, C++23 |

---

## Example PR comment

The workflow posts (and updates) a single comment per PR that looks like:

> ### `GCC-14-cpp20`
> | Benchmark | std mean | gsl mean | ratio | status |
> |---|---|---|---|---|
> | IsSorted | 124 ns | 125 ns | 1.01× | ✅ 1.01× |
> | IsSortedRanges | 88 ns | 89 ns | 1.01× | ✅ 1.01× |
> | MinElementAlgorithm | 95 ns | 96 ns | 1.01× | ✅ 1.01× |

---

## Notes

- `-fbounds-safety` config is stubbed out in the matrix with a comment
  explaining it's not yet stable in mainline Clang releases — easy to
  uncomment when it lands
- FetchContent deps are cached on `CMakeLists.txt` hash so CI doesn't
  re-clone on every run
- The comment job uses `peter-evans/find-comment` +
  `create-or-update-comment` so repeated pushes to the same PR update
  the existing comment rather than flooding the thread


Gafoor2005:
@microsoft-github-policy-service agree

.github/workflows/span_benchmark.yml
```diff
@@ -0,0 +1,413 @@
+# .github/workflows/span_benchmark.yml
```
carsonRadtke: nit: probably don't need the filename here.

.github/workflows/span_benchmark.yml
```diff
@@ -0,0 +1,413 @@
...
+#
+# Three separate job groups (one per OS) because `runs-on` can't share a
+# matrix with OS-specific shell/path differences cleanly:
+#
+#   benchmark-linux   → ubuntu-latest  → GCC-13, GCC-14, Clang-17, Clang-18
+#   benchmark-windows → windows-latest → MSVC 2022, clang-cl
+#   benchmark-macos   → macos-14       → Apple Clang (Xcode latest)
+#
+# A final `comment` job waits for all three groups, collects every JSON
+# artifact, runs check_regression.py, and posts (or updates) one PR comment.
```
carsonRadtke: nit: a brief description is okay, but this seems excessive.

.github/workflows/span_benchmark.yml
```diff
@@ -0,0 +1,413 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          # ── GCC ────────────────────────────────────────────────────────
+          - label: GCC-13-cpp20
+            cxx: g++-13
+            cppstd: 20
+            extra_flags: ""
```
carsonRadtke: Looks like `extra_flags` is always empty. Can we remove it?

.github/workflows/span_benchmark.yml
```diff
@@ -0,0 +1,413 @@
...
+
+
+          - label: GCC-14-cpp23
+            cxx: g++-14
+            cppstd: 23
+            extra_flags: ""
+
+
```
carsonRadtke: Seems like a lot of empty space. Can we clean this up?

.github/workflows/span_benchmark.yml
```diff
@@ -0,0 +1,413 @@
...
+# ─────────────────────────────────────────────────────────────────────────────
+# LINUX  —  GCC 13/14  ·  Clang 17/18  (× C++20 and C++23)
+# ─────────────────────────────────────────────────────────────────────────────
+  benchmark-linux:
+    name: Linux / ${{ matrix.label }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
```
carsonRadtke: I'm not familiar with `matrix.include`, but can't we just do something like:

```
matrix:
  cxx: [g++-12, g++-13, g++-14, clang++-16, clang++-17, clang++-18]
  cppstd: [20, 23]
```

Seems easier that enumerating every variation.

.github/workflows/span_benchmark.yml
```diff
@@ -0,0 +1,413 @@
...
+
+          # ── Clang + -fbounds-safety (core motivation of issue #1167) ───
+          # Safe Buffers enforcement is what triggered this whole tracking effort.
+          # - label: Clang-18-cpp20-bounds-safety
+          #   cxx: clang++-18
+          #   cppstd: 20
+          #   extra_flags: "-fbounds-safety"
+          # ----------------------------------------------------------------------------
+          # ⚠️ Note: As of early 2026, the feature is not yet fully available or stable in mainline, official releases (like Clang 20, 21, or 22).
+          # ----------------------------------------------------------------------------
```
carsonRadtke: I don't think `-fbounds-safety` is correct here.

.github/workflows/span_benchmark.yml
```diff
@@ -0,0 +1,413 @@
...
+
+
+      # ── Compiler install ───────────────────────────────────────────────
+      # GCC-13 and Clang pre-17 are already on ubuntu-latest (24.04).
+      - name: Install GCC 14
+        if: startsWith(matrix.cxx, 'g++-14')
+        run: |
+          sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y
+          sudo apt-get update -qq
+          sudo apt-get install -y g++-14
```
carsonRadtke: We don't install gcc in the regular compilers workflow. Why do we need to do it here?

benchmark/README.md
```diff
@@ -0,0 +1,144 @@
+# `gsl::span` vs `std::span` Benchmark
```
carsonRadtke: We can delete this file, I think.

benchmark/CMakeLists.txt
```diff
@@ -0,0 +1,87 @@
+# benchmark/CMakeLists.txt
```
carsonRadtke: Some general notes:

- I'd like this to be very similar to GSL_TEST; can we have a no `-DGSL_BENCHMARK` when invoking the root CMakeLists.txt?
- I also would like to rely on the same `GSL_CXX_STANDARD`?
- We already download googletest for the test suite, is it necessary to do it again here?  They are also different versions. Why?
- Google benchmark latest tag is 1.9.5, can we update to that?
- There is a performance cost to `-fno-omit-frame-pointer`, can we not enable the omission?

carsonRadtke (changes requested):
Thanks for getting this put together! It'll be great to have the infrastructure necessary to compare GSL types to standard library types. I have briefly reviewed many of the changes, please address my comments and I will perform a more in-depth review once comments have been resolved. I will also assign Copilot as a reviewer, so you get some more feedback; please address those comments too.

This currently is a proof-of-concept of a really cool feature that will help us develop GSL, but I want to make sure the PR is clean and well-understood before running any workflows and merging your changes.

Also, using AI as a tool to write code is great. Please make sure you understand the code that you are contributing. I want to hold all contributions to a high standard and right now this PR needs some work.

Gafoor2005:
I have addressed all your changes and most of the Copilot's too, there are few unresolved Copilot changes which I don't know are that reasonable. I need your opinion on them

carsonRadtke:
I see the updates. I have been out of office the past week and will continue to be out next week. I'll review the changes when I get back the week of the 22nd. Thanks for your patience.

.github/workflows/span_benchmark.yml
```diff
@@ -0,0 +1,156 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        cxx: [g++-13, g++-14, clang++-16, clang++-17, clang++-18]
+        cppstd: [20, 23]
+        exclude:
+          - cxx: g++-13
+            cppstd: 23 
+          - cxx: clang++-17
+            cppstd: 23
```
carsonRadtke, line 25: A comment explaining why these are excluded would be helpful for future readers (and me presently).

benchmark/CMakeLists.txt
```diff
@@ -0,0 +1,57 @@
...
+# Minimum is 20 because std::span (required for this benchmark) is C++20 only.
+# If a consumer passes GSL_CXX_STANDARD < 20 we clamp and warn.
+set(GSL_CXX_STANDARD "20" CACHE STRING "Use c++ standard")
+
+if(GSL_CXX_STANDARD LESS 20)
+    message(WARNING
+        "GSL_CXX_STANDARD=${GSL_CXX_STANDARD} is too old for the span benchmark "
+        "(std::span requires C++20). Overriding to 20.")
+    set(GSL_CXX_STANDARD "20" CACHE STRING "Use c++ standard" FORCE)
+endif()
```
carsonRadtke, line 15: I think instead of a warning (and ignoring the user's standard version), I'd rather see the test just get skipped. In the future we may want to benchmark against other GSL features so the check should probably move down to `add_executable(span_bench span_bench.cpp)` business.

.github/workflows/span_benchmark.yml
```diff
@@ -0,0 +1,156 @@
...
+      matrix:
+        cxx: [g++-13, g++-14, clang++-16, clang++-17, clang++-18]
+        cppstd: [20, 23]
+        exclude:
+          - cxx: g++-13
+            cppstd: 23 
+          - cxx: clang++-17
+            cppstd: 23 
+    steps:
+      - uses: actions/checkout@v4
```
carsonRadtke, line 27: Can we make sure we are using the latest actions? I think `actions/checkout` is at v7 and `actions/cache` is at v5 and `actions/upload-artifact` is also at v7. (Please check any others I may have missed).

.github/workflows/span_benchmark.yml
```diff
@@ -0,0 +1,156 @@
...
+            -DCMAKE_CXX_STANDARD=${{ matrix.cppstd }} \
+
+      - name: Build
+        run: cmake --build build --target span_bench -j$(nproc)
+
+      - name: Run benchmark
+        run: |
+          ./build/span_bench \
+            --benchmark_format=json \
+            --benchmark_repetitions=10 \
```
carsonRadtke, line 52: How did we arrive at 10 being sufficient?

benchmark/check_regression.py
```diff
@@ -0,0 +1,283 @@
...
+    result: dict[str, dict] = {}
+    for bm in data.get("benchmarks", []):
+        name: str = bm["name"]           # e.g. "BM_IsSorted_StdSpan_mean"
+        agg  = bm.get("aggregate_name")  # "mean" | "stddev" | "median" | "cv"
+
+        if agg not in ("mean", "stddev"):
+            continue
+
+        # Strip the trailing _mean / _stddev to get the canonical name
+        base_name = name[: name.rfind(f"_{agg}")]
```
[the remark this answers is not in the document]
carsonRadtke: > Guard this by checking name.endswith(f'_{agg}')

Sounds good to me. Maybe just `assert name.endswith(...)`.

benchmark/span_bench.cpp
```diff
@@ -0,0 +1,174 @@
...
+#include <span>
+#include <vector>
+
+static std::vector<int> make_vector()
+{
+    std::vector<int> v;
+    constexpr size_t vec_size = 1000;
+    v.reserve(vec_size);
+    for (int i = 0; i < vec_size; ++i)
+        v.push_back(i);
```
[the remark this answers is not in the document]
carsonRadtke: Feel free to ignore this.

benchmark/span_bench.cpp
```diff
@@ -0,0 +1,174 @@
+#include "gsl/span"
+#include <algorithm>
+#include <benchmark/benchmark.h>
+#include <limits>
+#include <ranges>
+#include <span>
+#include <vector>
+
+static std::vector<int> make_vector()
```
carsonRadtke, line 9: `make_vector` only creates sorted vectors of `int`. It isn't really a robust benchmark suite.

Feel free to punt this, but please add a comment documenting the shortcoming.

benchmark/span_bench.cpp
```diff
@@ -0,0 +1,174 @@
...
+}
+
+static void BM_MinElementAlgorithmStdSpan(benchmark::State& state)
+{
+    auto vec = make_vector();
+
+    for (auto _ : state)
+    {
+        std::span<const int> sp = vec;
+        benchmark::DoNotOptimize(std::min_element(sp.begin(), sp.end()));
```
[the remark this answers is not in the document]
carsonRadtke: Copilot might be right here. I need to look at this more carefully.

.github/workflows/span_benchmark.yml
```diff
@@ -0,0 +1,156 @@
...
+concurrency:
+  group: span-bench-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  benchmark-linux:
+    name: Linux / ${{ format('{0}-cpp{1}', matrix.cxx, matrix.cppstd) }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
```
carsonRadtke, line 17: Why?

.github/workflows/span_benchmark.yml
```diff
@@ -0,0 +1,156 @@
...
+          - cxx: clang++-17
+            cppstd: 23 
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Cache FetchContent dependencies
+        uses: actions/cache@v4
+        with:
+          path: build/_deps
+          key: fetchcontent-linux-${{ matrix.cxx }}-${{ hashFiles('benchmark/CMakeLists.txt') }}
```
carsonRadtke, line 33: Why doesn't `cppstd` matter?

.github/workflows/span_benchmark.yml
```diff
@@ -0,0 +1,156 @@
...
+          restore-keys: |
+            fetchcontent-linux-${{ matrix.cxx }}-
+
+      - name: Configure
```
carsonRadtke, line 37: Instead of "Configure", "Build", and "Run benchmark", can we repurpose the existing `./.github/workflows/cmake` workflow?

.github/workflows/span_benchmark.yml
```diff
@@ -0,0 +1,156 @@
...
+            --benchmark_repetitions=10 \
+            --benchmark_report_aggregates_only=true \
+            --benchmark_out=results_${{ format('{0}-cpp{1}', matrix.cxx, matrix.cppstd) }}.json
+
+      - name: Upload results
+        uses: actions/upload-artifact@v4
+        with:
+          name: bench-${{ format('{0}-cpp{1}', matrix.cxx, matrix.cppstd) }}
+          path: results_${{ format('{0}-cpp{1}', matrix.cxx, matrix.cppstd) }}.json
+          retention-days: 7
```
carsonRadtke, line 61: Any particular reason for 7 days?

.github/workflows/span_benchmark.yml
```diff
@@ -0,0 +1,156 @@
...
+
+  benchmark-windows:
```
carsonRadtke, line 63: Please also apply my linux comments to windows and macos.

.github/workflows/span_benchmark_comment.yml
```diff
@@ -0,0 +1,76 @@
...
+        with:
+          comment-id: ${{ steps.find_comment.outputs.comment-id }}
+          issue-number: ${{ steps.pr.outputs.result }}
+          body-path: comment.md
+          edit-mode: replace
+
+      - name: Fail if regression detected
+        if: steps.report.outcome == 'failure'
+        run: |
+          echo "::error::Performance regression detected. See the PR comment for the full table."
```
carsonRadtke, line 75: nit: technically the cause could also have been bad arguments passed to `check_regression.py`.

carsonRadtke: Can we check the return code? I think argparse will exit with code 2 for bad arguments and `check_regression.py` will exit with code 1 for a regression.

carsonRadtke (changes requested):
Thanks for making the requested changes! It seems to have made the PR much smaller 👍. 

I went through the PR in greater detail and left another batch of comments.

PR #1257, opened 2026-07-30
Support find_package for GTest

### LINKED ISSUE #1256: Support find_package for GTest to avoid redundant downloads
### Description

In tests/CMakeLists.txt, the current logic uses pkg_search_module(GTestMain gtest_main) to detect a pre-installed GTest. If not found, it falls back to downloading GTest source via Git.

However, pkg_search_module relies on pkg-config (.pc files), which is not commonly available or generated by package managers like **vcpkg** on Windows. vcpkg provides CMake config files (e.g., GTestConfig.cmake) instead, which are meant to be used with find_package().

As a result, even when GTest is already installed via vcpkg, pkg_search_module fails to find it, and the build unnecessarily downloads and compiles GTest from source.

### Suggestion

Add find_package(GTest) as an additional detection method before falling back to the download step. For example:

```CMake
find_package(GTest QUIET)
if (NOT GTest_FOUND)
    pkg_search_module(GTestMain gtest_main)
endif()

if (NOT GTest_FOUND AND NOT GTestMain_FOUND)
    # Fall back to downloading GTest via Git
    ...
endif()
```

This would allow users with GTest installed via vcpkg (or other CMake-config-based package managers) to use their existing installation without redundant downloads.

true-brace05:
## Summary

This change attempts to locate a pre-installed GoogleTest using `find_package(GTest QUIET)` before falling back to `pkg_search_module` and downloading GoogleTest.

This avoids redundant downloads when GoogleTest is already installed through CMake package managers such as vcpkg, while preserving the existing fallback behavior.

Fixes #1256

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -22,8 +22,17 @@ if(IOS)
     add_compile_definitions(GTEST_HAS_DEATH_TEST=1 IOS_PROCESS_DELAY_WORKAROUND=1)
 endif()
 
-pkg_search_module(GTestMain gtest_main)
-if (NOT GTestMain_FOUND)
+find_package(GTest QUIET)
+
+if(GTest_FOUND)
+    set(GTestMain_LIBRARIES GTest::gtest_main)
+endif()
+
+if(NOT GTest_FOUND)
+    pkg_search_module(GTestMain gtest_main)
+endif()
+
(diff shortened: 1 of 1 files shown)


true-brace05:
Hi, I've submitted a PR for this issue: #1257. Looking forward to your review. Thanks!

PR #1268, opened 2026-08-27
Configure monthly Dependabot updates

JamieMagee:
## Summary

Add monthly Dependabot checks for GitHub Actions. Minor and patch updates are grouped, while major updates remain separate.

PR #1259 also adds a Dependabot configuration, but it uses a weekly schedule, adds a cooldown, groups major updates, and includes broader unrelated workflow changes. This PR keeps the policy and diff focused.

## Validation

- Validated `.github/dependabot.yml` against the SchemaStore schema
- Confirmed `/` covers the repository's external GitHub Actions references
- Confirmed no other supported Dependabot manifests are present
- Confirmed the diff contains only `.github/dependabot.yml`


carsonRadtke (approved):
Thanks for the PR!

PR #1271, opened 2026-09-06
Let dyn_array_iterator convert to its const_iterator

### LINKED ISSUE #1269: dyn_array_iterator cannot convert to const_iterator
`dyn_array_iterator<T>` declares `operator dyn_array_iterator<const T>()`, but the conversion builds the target through `dyn_array_iterator<const T>`'s private state constructor, and that specialization befriends only `gsl::dyn_array`. Every use of the conversion is an access error, so `iterator` never converts to `const_iterator`:

```cpp
gsl::dyn_array<int> a{1, 2, 3};
gsl::dyn_array<int>::const_iterator it = a.begin();   // error: constructor is private
bool b = a.cend() == a.end();                          // same
```

Reproduces on main (417ef68) with GCC 15 at C++14, 17 and 20. `gsl::span`'s `span_iterator` avoids this by befriending its own specializations (`template <class> friend class span_iterator;`).

lenamonj:
Fixes #1269.

`dyn_array_iterator<T>`'s conversion to `dyn_array_iterator<const T>` builds the target through that specialization's private state constructor, which befriended only `gsl::dyn_array`, so the conversion was an access error at every use. The iterator now befriends its own specializations, as `span_iterator` does.

The new test assigns `begin()` to a `const_iterator` and compares `cend()` with `end()`; it fails to compile on main.

### DIFF
--- a/include/gsl/dyn_array
+++ b/include/gsl/dyn_array
@@ -288,6 +288,8 @@ namespace details
         size_type _pos{};
         size_type _end_pos{};
 
+        template <typename>
+        friend class dyn_array_iterator;
         template <typename, typename>
         friend class ::gsl::dyn_array;
     };
--- a/tests/dyn_array_tests.cpp
+++ b/tests/dyn_array_tests.cpp
@@ -585,6 +585,15 @@ TEST(dyn_array_tests, const_operations)
     EXPECT_EQ(pirates[0], 'a');
 }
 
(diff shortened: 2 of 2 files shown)


carsonRadtke:
LGTM. Thanks for the fix and additional test!

carsonRadtke (approved):

